PR #39, opened 2016-07-29
Fix build on macOS

rgov:
I tried running the bootstrap script on macOS but hit a few snags. These changes get me through to compiling Remill, but I get a lot of output like

```
E0728 21:58:21.001000 3291550656 Translator.cpp:467] Could not find lifted block for PC 4512489476
```


dguido:
Thanks for the PR @rgov but we only support OS X.

PR #61, opened 2016-10-04
OS X build

japesinator:
Start work on building on OS X.

This PR updates install_xed.sh to use GNU install on OS X and not use ldconfig, and updates the README with instructions for building on OS X.

This still needs a Travis script, and I might have missed some of the dependencies necessary to install with brew because I already had them.


pgoodman:
@japesinator Awesome!! Thanks :-D

@yan Can you take a look at this? I don't have a mac.

yan:
So this still uses `LLVM_DIR` from the original CMakeLists.txt. Can we set `LLVM_DIR` to be relative to the LLVM tarball the instructions pull down and extract? Furthermore, can we add instructions to pull it down to the build dir, not root of remill?

i.e.

- `mkdir build && cd build/`
- `curl -O http://llvm.org/releases/3.9.0/clang+llvm-3.9.0-x86_64-apple-darwin.tar.xz`
  - (this would also drop the wget requirement)
- `tar xfz ....`

then change `LLVM_DIR` to point to it relative to `CMAKE_BINARY_DIR` and resolve all the CMAKE_C[XX]_COMPILER vars correctly

japesinator:
I've updated the README, but I'm confused about what you'd like done with CMakeLists. Do we need to detect the OS to set the LLVM binary directory, since it will be in different places in Linux and OS X?

pgoodman:
I think @yan might mean something like [Screen](https://github.com/trailofbits/screen)'s build setup in mind (for both macOS and Linux). In there, it seems that `build.sh` pulls in an LLVM tarball (for both OSes), then links that in with the usual cmake build step for screen itself.

yan:
Exactly, however I'm still thinking about how to pull in a tarball during build, but use the system LLVM during install. @pgoodman, is using the system LLVM important?

pgoodman:
@yan, no it's not all that important. One of the things I really wanted to make sure of is that I could have a downstream tool pick up on the llvm installed. So, for example, we could have a sequence of LLVM passes that we could implement elsewhere, and *know* that they can run on the remill-produced bitcode. I went the system package route so that I could "guarantee" that the right version of LLVM was available.

pgoodman:
@japesinator or @yan, can one of you figure out why the code is not building in Travis?

japesinator:
It looks to me like it's running `remill/travis/install_macos.sh` where it should be running `remill/scripts/travis/install_macos.sh` in the latest OS X build.

Also, I'm currently in the process of writing a screen style build script for remill as you mentioned earlier, and will add that to this PR once it works.

pgoodman:
@japesinator  Great! Thanks for the update :-) Please fix `.travis.yml` along the way :-D

pgoodman:
@japesinator Can you resolve these merge conflicts? Thanks!

Issue #59, opened 2016-09-30
Add __remill_hyper_call intrinsic


pgoodman:
Add a `__remill_hyper_call` intrinsic that replaces / subsumes any architecture-specific functionality that cannot be modelled in a generic way, or whose modelling would mean adding many, potentially non-generic, intrinsics.

An `enum` should be made specifying all possible hypercalls. These would be specified under `remill/Arch` and not within any arch-specific directories. Example enumeration constants would be:

``` C++
enum class HyperCall {
  kInvalid,
  kX86CPUID,
  kX86READTSC,
  kX86READTSCP,
};
```

A variable, `HYPER_CALL_ID`, should be added into `__remill_basic_block`. This variable can be used just like `PC`, `NEXT_PC`, `STATE`, and `MEMORY`. It should reference a new field in the `State` structure type, and that field should be within the architecture-neutral state structure base class (Issue #53).

pgoodman:
It is probably worth replacing `__remill_interrupt_call` and `__remill_interrupt_return`, as well as the syscall variants. That way we can have hypercalls for `kX86INT`, `kX86SYSCALL`, `kX86SYSENTER`, etc. The variants (at least w.r.t. syscall/sysenter) are a bit too diverse to handle with some generic intrinsic.

pgoodman:
I'm going to split this into a synchronous and asynchronous variants. Async variants are for control flows like `int`, `syscall`, and `sysenter`. Synchronous are for things like `cpuid`, `rdtsc`, etc.

Issue #53, opened 2016-09-15
Make State structures derived from an architecture-neutral base class


pgoodman:
This is related to the footnote in Issue #52. The idea here is that some information (e.g. the interrupt vector) cannot be passed through control-flow intrinsics (e.g. `__remill_interrupt_call`) because of the rigid  argument requirements for control-flow intrinsics (`State *state, Memory *memory, addr_t pc`). I think an appropriate solution is to put these architecture-neutral "dirty details" into a base class, and have the `State` structure derive from this base class. Then, control-flow intrinsics can be defined as accepting pointers to the base class. Implementations of the intrinsics that require access to the actual machine-specific contents can then down-cast the pointer.

Issue #45, opened 2016-09-07
X87 Instruction support


pgoodman:
Implement and test the following instructions:
 - [ ] F2XM1
 - [ ] FABS
 - [ ] FADD
 - [ ] FADDP
 - [ ] FBLD
 - [ ] FBSTP
 - [ ] FCHS
 - [ ] FCMOVB
 - [ ] FCMOVBE
 - [ ] FCMOVE
 - [ ] FCMOVNB
 - [ ] FCMOVNBE
 - [ ] FCMOVNE
 - [ ] FCMOVNU
 - [ ] FCMOVU
 - [ ] FCOM
 - [ ] FCOMI
 - [ ] FCOMIP
 - [ ] FCOMP
 - [ ] FCOMPP
 - [ ] FCOS
 - [ ] FDECSTP
 - [ ] FDISI8087_NOP
 - [ ] FDIV
 - [ ] FDIVP
 - [ ] FDIVR
 - [ ] FDIVRP
 - [ ] FENI8087_NOP
 - [ ] FFREE
 - [ ] FFREEP
 - [ ] FIADD
 - [ ] FICOM
 - [ ] FICOMP
 - [ ] FIDIV
 - [ ] FIDIVR
 - [ ] FILD
 - [ ] FIMUL
 - [ ] FINCSTP
 - [ ] FIST
 - [ ] FISTP
 - [ ] FISUB
 - [ ] FISUBR
 - [ ] FLD
 - [ ] FLD1
 - [ ] FLDCW
 - [ ] FLDENV
 - [ ] FLDL2E
 - [ ] FLDL2T
 - [ ] FLDLG2
 - [ ] FLDLN2
 - [ ] FLDPI
 - [ ] FLDZ
 - [ ] FMUL
 - [ ] FMULP
 - [ ] FNCLEX
 - [ ] FNINIT
 - [ ] FNOP
 - [ ] FNSAVE
 - [ ] FNSTCW
 - [ ] FNSTENV
 - [ ] FNSTSW
 - [ ] FPATAN
 - [ ] FPREM
 - [ ] FPREM1
 - [ ] FPTAN
 - [ ] FRNDINT
 - [ ] FRSTOR
 - [ ] FSCALE
 - [ ] FSETPM287_NOP
 - [ ] FSIN
 - [ ] FSINCOS
 - [ ] FSQRT
 - [ ] FST
 - [ ] FSTP
 - [ ] FSTPNCE
 - [ ] FSUB
[...]

PR #63, opened 2016-10-11
Issues 52 53 state struct changes

pgoodman:
This PR has a few changes:

 - It adds a new "base" state structure type, that is the first element of `State` (Issue #53). This will open the way to handling Issue #59.
 - It removes `FPU` from within the state structure. This was a comment on Issue #45. The idea is that the `FPU` struct is too low level (it covers the `fsave`, `fxsave`, and `fxsave64` variants). I started working on some of the FPU comparison instructions a while ago. These additions motivated me to pull the `FPU` struct out because I ended up using masking and flipping bits in the control and status words. I'd like these things to be like `ArithFlags`, where there is one `bool` per flag.
 - It removes the requirement of having to manually specify via an ugly macro in test cases that the test case uses MMX technology state. Instead, Run.cpp has been upgraded to recognise (heuristically) when MMX vs. ST(n) registers are used.
 - It adds a few tests for `FLD*` instructions.


tests/X86/X87/FLD.S
```diff
@@ -0,0 +1,36 @@
...
+TEST_INPUTS(0)
+    fld1
+TEST_END
+
+TEST_BEGIN(FLD2T, 1)
+TEST_INPUTS(0)
+    fldl2t
+TEST_END
+
+TEST_BEGIN(FLD2E, 1)
```
kumarak: Instruction is defined as FLDL2T & FLDL2E

pgoodman:
@akshay, can you squash and merge this?

kumarak:
Sure, I will merge this pull request.

PR #68, opened 2016-11-03
Issue 44 mmx support extension

pgoodman:


pgoodman:
The CI actually passed for Linux; I cancelled the OS X job and Travis CI is stupid and considers cancellations of an allowed-to-fail build to be a problem.

PR #74, opened 2016-11-10
OSX gtest debugging

japesinator:
Travis is being weird so I'm opening a PR to finally get it to build these tests


pgoodman:
Is the ready to merge?

japesinator:
I'd like to get those weird symbol names taken care of today first, so it can pass both travis tests

japesinator:
OK, this should be mergable

PR #88, opened 2017-01-27
Ubuntu64 terminal calls

douggard:
## README.md

* Updated to have version 2.6.1 of protobuf. 

* Removed a `cd ..` that wasn't needed. 

* Also added instructions for how to get protobuf in IDA on 64 bit Linux. Apparently IDA will not use the system's Python (even after install) on Linux. You can add the modules manually to the zip file though.

## disass.py

* Added parent dir to sys.path since those imports were failing for me. 

* Added a call `inst_terminates_bb` that uses IDA's functionality to lookup whether an instruction is the last in a basic block and the basic block has no successors. Support code is BBCache (per function) and a 5 function cache of those. Updated `is_call` to utilize the `inst_terminates_bb` function. 

* Added a `JUMP_TABLE_EAS` set used to filter out addresses within jump tables.

## x86.py

* Added loop mnemonics.

### DIFF
--- a/tools/disass/ida/disass.py
+++ b/tools/disass/ida/disass.py
@@ -27,7 +27,8 @@ except:
 
 
 POSSIBLE_CODE_REFS = set()
-
+# Stores all addresses contained in jump tables.
+JUMP_TABLE_EAS = set()
 
 # Architecture-specific functions.
 decode_instruction = None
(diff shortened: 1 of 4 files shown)


tools/disass/ida/disass.py
```diff
@@ -136,6 +139,87 @@ def get_direct_branch_target(branch_inst_ea):
         branch_inst_ea, target_ea))
     return target_ea
 
+# source: http://reverseengineering.stackexchange.com/a/1648/4197
```
pgoodman: I'd like to see the below code to be removed for the following reasons:

 * The caching functionality is already covered by using things like `program.get_basic_block`. The `program` module also keeps track of more information, including subroutines, and block successors (including jump table targets).
 * This mechanism produces blocks that will be structured differently than Remill basic blocks. IDA's basic blocks do not end at `call` instructions.
 * This abstraction doesn't handle the case where two basic blocks overlap. Mapping an `ea` to a block is a one-to-many relationship. One place where this happens is glibc, where code will conditionally jump to a `cmpxchg` after a `lock` prefix. Another place that this can sometimes happen is in the Linux kernel.
 * The licensing of this code is unclear, despite it posted on a public StackExchange forum.

tools/disass/ida/disass.py
```diff
@@ -154,7 +238,11 @@ def get_static_successors(inst):
       yield inst.next_ea  # Not recognised as a `noreturn` function.
 
   if inst.is_call():  # Indirect function call, system call.
-    yield inst.next_ea
+    if inst_terminates(inst):
```
pgoodman: This looks to be a bug, it should be `inst_terminates_bb`.

I think that a better use of this would be to make the following checks:
 * If the call is a no-return call, then don't `yield inst.next_ea`. It would be nice to have a function like `is_noreturn_call` or something.
 * If the code following the call is not code, according to `is_code`, then don't `yield inst.next_ea`.

The semantics of this instruction are to go and find the natural successors of a particular instruction, not of a block, and so recognizing whether or not a successor instruction is a block terminator (in the sense as it's another block head) does not fit into the semantics of this function. The `analyse_subroutine` function handles the identification and more fine-grained splitting of block heads, using the loop starting at line 306.

tools/disass/ida/disass.py
```diff
@@ -377,11 +464,36 @@ def get_called_subroutines(sub):
...
+def lrange(start, end):
+  num = long(start)
+  while num < end:
+    yield num
+    num += 1L
+
+# source: http://reverseengineering.stackexchange.com/a/13472/4197
+def init_jump_table_eas():
+  global JUMP_TABLE_EAS
+  JUMP_TABLE_EAS = set()
```
pgoodman: This definition of `JUMP_TABLE_EAS = set()` belongs at the top of the file.

tools/disass/ida/disass.py
```diff
@@ -377,11 +464,36 @@ def get_called_subroutines(sub):
...
   return called_sub_eas
 
+# Python 2.7 doesn't range on longs
+def lrange(start, end):
+  num = long(start)
+  while num < end:
+    yield num
+    num += 1L
+
+# source: http://reverseengineering.stackexchange.com/a/13472/4197
```
pgoodman: Can you add a comment describing what this function does, and why. The purpose of the code on the RE stack exchange is not the same as what this code is doing.

tools/disass/ida/disass.py
douggard: I'll make all those changes. 

Regarding `program.get_basic_block`, I never was able to get it to return a basic block that had data. `has_basic_block` would return False and any call to `get_basic_block`, with an arbitrary address, or the address of the first instruction of a basic block, would return a basic block with a bunch of null fields. I'll reproduce this on Monday and open a bug.

On indirect calls I could not find a way to check if the target function was `noreturn`. The flags were always -1. Is there a way to dereference the target address, check if it is in .idata, then get flags from the target in there? I'll give that a shot to see if I can get flags. The other solution would be getting the name of the target function and keeping a black list of noreturn names, that doesn't sound ideal though. I reverted to structural verification because of this, and I reverted to IDA's flow chart due to the `get_basic_block` not returning initialized blocks. 

Could each instruction store a reference to its parent block? That would make retrieval most simple. Rather than going from arbitrary address, then using IDA to get the first address in the basic block, then retrieving the basic block from `get_basic_block`.

tools/disass/ida/disass.py
douggard: Regarding program.get_basic_block, I never was able to get it to return a basic block that had data. has_basic_block would return False and any call to get_basic_block, with an arbitrary address, or the address of the first instruction of a basic block, would return a basic block with a bunch of null fields. I'll reproduce this on Monday and open a bug.

On indirect calls I could not find a way to check if the target function was noreturn. The flags were always -1. Is there a way to dereference the target address, check if it is in .idata, then get flags from the target in there? I'll give that a shot to see if I can get flags. The other solution would be getting the name of the target function and keeping a black list of noreturn names, that doesn't sound ideal though. I reverted to structural verification because of this, and I reverted to IDA's flow chart due to the get_basic_block not returning initialized blocks.

Could each instruction store a reference to its parent block? That would make retrieval most simple. Rather than going from arbitrary address, then using IDA to get the first address in the basic block, then retrieving the basic block from get_basic_block.

pgoodman: There is a subtlety to how `program.get_basic_block` is used that isn't explained but shouldn't be. It's really the `analyse_subroutine` and `analyse_block` that fills up the backing cached data structures, so before that happens, you'll get back empty things.

I think there is probably a way to dereference the data, assuming it's a pointer into something like the GOT/PLT. You may be able to inspect the address in IDA's `inst_t`. Another approach may be to look for data cross-references from the indirect call.

Each instruction could do that, and you can add it in. Why is that needed, though?

pgoodman:
@douggard, can you fix the CI failures for Linux?

pgoodman:
Also, please add your name to the acknowledgements file as part of this PR ;-)

douggard:
Passed.

PR #90, opened 2017-02-03
Optimizer.cpp: Fix crash when func.front() is non-existant

douggard:
Introduces a function `GetFuncFrontOrNull(llvm::Function)` that returns `nullptr` when the function passed to it has no basic blocks. Also ads a `nullptr` entry to map `block_transfer_functions` with the `local_live` and `local_dead (mask)` set.

Also modified is `Run03` with various speed improvements suggested by @pgoodman.

### DIFF
--- a/remill/BC/Optimizer.cpp
+++ b/remill/BC/Optimizer.cpp
@@ -850,7 +850,7 @@ static void GetBlockSuccessors(llvm::BasicBlock *block,
   }
 }
 
-// Get a functions front or a null pointer if no front available. 
+// Get a function's front or a null pointer if no front available. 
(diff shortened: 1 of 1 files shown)


remill/BC/Optimizer.cpp
```diff
@@ -781,8 +781,8 @@ static void RunO3(llvm::Module *module) {
   TLI->disableAllFunctions();
 
   llvm::PassManagerBuilder builder;
-  builder.OptLevel = 3;  // -O3.
-  builder.SizeLevel = 2;  // -Oz
+  builder.OptLevel = 0;  // -O3.
```
pgoodman: Can you update the `// -O3.` comment to `// -O0.`?

PR #103, opened 2017-02-23
Add radare2 support to disass tool

roachspray:
This has the dependency of radare2 disassembler (similar to the BN and IDA support in the disass tool). Please feel free to bash the code; I do not claim it is pretty.


tools/disass/radare2/x86.py
```diff
@@ -0,0 +1,167 @@
...
+    for fne in fn_list:
+        name = fne["name"]
+        new_block = True
+
+        # Having each function be a named block. Is enough?
+        named_block = CFG_pb2.NamedBlock()
+        named_block.name = name
+        named_block.address = fne["offset"]
+
+        # XXX: I am a bit confused on the meaning of the 1 and 0 for
```
pgoodman: Use `TODO(name)` format here. That way people know who to ask when they're trying to understand this.

tools/disass/radare2/x86.py
```diff
@@ -0,0 +1,167 @@
...
+
+    #
+    # The purpose of running these is to get the executable
+    # analyzed so the representation given from radare2 is 
+    # in a form that is useful to being remill-ingested.
+    # I am sure there are improvements to be made here.
+    #
+    # Should investigate using:
+    #   aaaa is experimental
+    log.warning("Fuck you")
```
pgoodman: Please change this.

tools/disass/radare2/x86.py
```diff
@@ -0,0 +1,167 @@
...
+import log
+
+import CFG_pb2
+try:
+    import r2pipe
+except:
+    log.critical("Install the radare2 python bindings. (https://github.com/radare/radare2-r2pipe)")
+
+#
+# I have concerns on performance in large executable scenarios.
```
pgoodman: Unnecessary comment.

tools/disass/radare2/x86.py
```diff
@@ -0,0 +1,167 @@
...
+
+            # I was using python3 before so had bytes.fromhex() API call,
+            # but I believe this is the correct juju for Python2
+            btmp = [op["bytes"][i:i+2] for i in range(0, len(op["bytes"]), 2)]
+            ii.bytes = ''.join([chr(int(x, 16)) for x in btmp])
+            ii.address = op["offset"]    
+            oc = op["opcode"]
+
+            #
+            # Unsure if this is the behavior that is desired
```
pgoodman: `ud2` is a valid block terminator, and has well-defined semantics in Remill. A block ending in a `ud2` should be included in the CFG.

tools/disass/radare2/x86.py
```diff
@@ -0,0 +1,167 @@
...
+
+        # Having each function be a named block. Is enough?
+        named_block = CFG_pb2.NamedBlock()
+        named_block.name = name
+        named_block.address = fne["offset"]
+
+        # XXX: I am a bit confused on the meaning of the 1 and 0 for
+        # visibility. I am setting the imported symbols to visibility 1.
+        if name.startswith("sym.imp.") == True:
+            named_block.visibility = 1
```
pgoodman: Use `CFG_pb2.IMPORTED` or `CFG_pb2.EXPORTED` instead of the numbers.

tools/disass/radare2/x86.py
```diff
@@ -0,0 +1,167 @@
...
+                log.debug("Disassembling function: {}".format(name))
+                log.debug("{}".format(json.dumps(dis, sort_keys=False, indent=4, separators=(",", ": "))))
+                new_block = True
+                blk = None
+                break
+
+            blk.instructions.extend([ii])
+            blk_terminators = ["call", "ret", "int", "j", "iret", "hlt", "nop", \
+              "lcall", "ljmp", "syscall" ]
+            for bt in blk_terminators:
```
pgoodman: Need to add some `mod.addressed_blocks` for things like: return address of `call`, `syscall`, `sysenter`, etc. An addressed block is one that can be targeted by an indirect control flow. For example, the return address of a `call` instruction is treated as addressable because the `ret` instruction is an indirect control flow.

tools/disass/radare2/x86.py
roachspray: Yikes. Yes it should be changed.

tools/disass/radare2/x86.py
roachspray: Ok, will do; thanks.

tools/disass/radare2/x86.py
roachspray: Ok, I will do this. Thanks

PR #110, opened 2017-05-14
Simplify runtime targets generation.

alessandrogario:
This PR adds support for the bitcode language (BC) to CMake, adding a new function that can be used to generate runtime targets in a more CMake-like way. 

Example: https://github.com/trailofbits/remill/blob/alessandro/refactor/cmake-bitcode-language/remill/Arch/X86/Runtime/CMakeLists.txt

This should not only make it easier for new architecture modules to be added to remill, but also speed up compilation time.


pgoodman:
@alessandrogario Should I delete this branch or will you continue with it?

alessandrogario:
You can delete it! Thanks! :+1:

PR #119, opened 2017-06-22
ADRP and B instructions for AArch64

artemdinaburg:
* Support for `ADRP` (untested but translates)
* Support for PC-relative `B` (looks to work -- resolves targets during translation)


artemdinaburg:
closing due to more updated PR.

PR #134, opened 2017-09-05
Issue 127 mins maxs support

pgoodman:


pgoodman:
Submitting PR because it has useful instructions.

PR #138, opened 2017-09-15
Wip aarch64 /bin/ls and libc instructions

perks:
~~I've added what I had so far. `CLZ` and the conditional instructions are failing due to lifting, i.e:~~

~~(Interesting it is complaining about the opposite condition than what the test has)~~

^ Fixed in commits below -- Bug in autogenerated extractor function needed to invert condition for alias


perks:
There is a bug in the some of the functions 64 bit lifted functions that overwrites the value of the `x3` register for seemingly no reason. Here are some differences in `native->gpr` vs `lifted->gpr` (first run of each)

- `csel_64_eq_2`: `0xdead` vs `0x8`
- `csinc_64_eq_2`: `0xdead` vs `0x8` AND`x2`: `0x0` vs `0xdead`
- `cinc_csinc_64_eq_2`: `0xfacf` vs `0x8` AND `x2: `0x0` vs `0xfacf`

So on closer inspection it looks like register #s are getting shifted down or some off by one?

remill/Arch/AArch64/Arch.cpp
```diff
@@ -1689,6 +1689,15 @@ bool TryDecodeAND_64_LOG_SHIFT(const InstData &data, Instruction &inst) {
   return TryDecodeEOR_64_LOG_SHIFT(data, inst);
 }
 
+// ANDS  <Wd>, <Wn>, #<imm>
+bool TryDecodeANDS_32S_LOG_IMM(const InstData &data, Instruction &inst) {
+  return TryDecodeEOR_32_LOG_IMM(data, inst);
```
pgoodman: This is not quite correct. `EOR` is `EOR  <Xd|SP>, <Xn>, #<imm>`, so the if `Rd` is 31, then it's treated as `SP`/`WSP`, whereas with `ANDS`, it's treated as `XZR`/`WZR`.

remill/Arch/AArch64/Arch.cpp
```diff
@@ -1774,6 +1783,148 @@ bool TryDecodeMOV_ORR_64_LOG_IMM(const InstData &data, Instruction &inst) {
   return true;
 }
 
+
+// CLZ  <Wd>, <Wn>
+bool TryDecodeCLZ_32_DP_1SRC(const InstData &data, Instruction &inst) {
+  AddRegOperand(inst, kActionWrite, kRegW, kUseAsAddress, data.Rd);
```
pgoodman: Here we want `kUseAsValue` because it's just `Wd`, not `Wd|WSP`, same for `Wn`.

remill/Arch/AArch64/Arch.cpp
```diff
@@ -1774,6 +1783,148 @@ bool TryDecodeMOV_ORR_64_LOG_IMM(const InstData &data, Instruction &inst) {
...
+  AddRegOperand(inst, kActionRead, kRegW, kUseAsAddress, data.Rn);
+  return true;
+}
+
+// CLZ  <Xd>, <Xn>
+bool TryDecodeCLZ_64_DP_1SRC(const InstData &data, Instruction &inst) {
+  AddRegOperand(inst, kActionWrite, kRegX, kUseAsAddress, data.Rd);
```
pgoodman: `kUseAsValue`.

remill/Arch/AArch64/Arch.cpp
```diff
@@ -1774,6 +1783,148 @@ bool TryDecodeMOV_ORR_64_LOG_IMM(const InstData &data, Instruction &inst) {
...
+  AddRegOperand(inst, kActionRead, kRegX, kUseAsAddress, data.Rn);
+  return true;
+}
+
+bool DecodeConditionalRegSelect(const InstData &data, Instruction &inst, RegClass r_class, bool second_source) {
+  AddRegOperand(inst, kActionWrite, r_class, kUseAsAddress, data.Rd);
```
pgoodman: `kUseAsValue`

remill/Arch/AArch64/Arch.cpp
```diff
@@ -1774,6 +1783,148 @@ bool TryDecodeMOV_ORR_64_LOG_IMM(const InstData &data, Instruction &inst) {
...
+  AddRegOperand(inst, kActionRead, r_class, kUseAsAddress, data.Rn);
+  if(second_source)
```
pgoodman: Always use `{` and `}` in `if`, `while`, `for`, etc. blocks, even if the body is empty or even just a single line.

remill/Arch/AArch64/Arch.cpp
```diff
@@ -1774,6 +1783,148 @@ bool TryDecodeMOV_ORR_64_LOG_IMM(const InstData &data, Instruction &inst) {
...
+}
+
+// CLZ  <Xd>, <Xn>
+bool TryDecodeCLZ_64_DP_1SRC(const InstData &data, Instruction &inst) {
+  AddRegOperand(inst, kActionWrite, kRegX, kUseAsAddress, data.Rd);
+  AddRegOperand(inst, kActionRead, kRegX, kUseAsAddress, data.Rn);
+  return true;
+}
+
+bool DecodeConditionalRegSelect(const InstData &data, Instruction &inst, RegClass r_class, bool second_source) {
```
pgoodman: I suggest adding a simple `enum` with two entries, one that is like "two sources", the other as "one source", as opposed to `true` and `false`.

remill/Arch/AArch64/Arch.cpp
```diff
@@ -1774,6 +1783,148 @@ bool TryDecodeMOV_ORR_64_LOG_IMM(const InstData &data, Instruction &inst) {
...
+}
+
+// CSEL  <Xd>, <Xn>, <Xm>, <cond>
+bool TryDecodeCSEL_64_CONDSEL(const InstData &data, Instruction &inst) {
+  return DecodeConditionalRegSelect(data, inst, kRegX, true);
+}
+
+// CSINC  <Wd>, <Wn>, <Wm>, <cond>
+bool TryDecodeCSINC_32_CONDSEL(const InstData &data, Instruction &inst) {
+  LOG(ERROR) << "Inside regular csinc";
```
pgoodman: Remove `LOG`.

remill/Arch/AArch64/Extract.cpp
```diff
@@ -47970,10 +47970,17 @@ static bool TryExtractCINC_CSINC_32_CONDSEL(InstData &inst, uint32_t bits) {
   inst.cond = static_cast<uint8_t>(enc.cond);
   if (!(inst.cond != 0xe)) return false;
   if (!(inst.cond != 0xf)) return false;
+  // Condition has to be inverted for the alias to work
+  // when converted to csinc instruction so we toggle last bit
+  inst.cond ^= 1;
```
pgoodman: I don't understand this. This seems like a "destructive" change, where this inversion should really be processed inside of the `TryDecode` function. Is inverting here necessary to handle aliasing?

remill/Arch/AArch64/Semantics/BITBYTE.cpp
```diff
@@ -0,0 +1,31 @@
...
+ * limitations under the License.
+ */
+
+
+
+namespace {
+
+template <typename D, typename S>
+DEF_SEM(CLZ, D dst, S src) {
+    auto val = ZExtTo<S>(Read(src));
```
pgoodman: `ZExtTo<S>` is unnecessary.

tests/AArch64/BITBYTE/CLZ_n_DP_1SRC.S
```diff
@@ -0,0 +1,59 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+
+/*
+Fails Because:
```
pgoodman: Lets add an issue to keep track of this.

tests/AArch64/COND/CSEL_n_CONDSEL.S
```diff
@@ -0,0 +1,54 @@
...
+    cmp w0, w1
+    csel w2, w3, w4, HI
+TEST_END
+    
+TEST_BEGIN(CSEL_64_CONDSEL_EQ, csel_64_eq, 2)
+TEST_INPUTS(
+    //  0x0, 0x1,
+     0xfffffffffffffffc, 0xfffffffffffffffc
+)
+    // Using x3 register will cause it fail (it gets overwritten, but other registers work...)
```
pgoodman: This should get fixed.

tests/AArch64/COND/x_CSINC_n_CONDSEL.S
```diff
@@ -0,0 +1,59 @@
...
+    cmp w0, w1
+    csinc w2, w3, w4, eq
+TEST_END
+    
+ TEST_BEGIN(CSINC_64_CONDSEL_EQ, csinc_64_eq, 2)
+ TEST_INPUTS(
+    //  0x0, 0x1,
+     0xfffffffffffffffc, 0xfffffffffffffffc
+ )
+    // Using x3 register will cause it fail (it gets overwritten, but other registers work...)
```
pgoodman: Should be fixed soon.

tests/AArch64/COND/x_CSINC_n_CONDSEL.S
```diff
@@ -0,0 +1,59 @@
...
+    cmp w0, w1
+    cinc w4, w3, eq
+TEST_END
+    
+ TEST_BEGIN(CINC_CSINC_64_CONDSEL_EQ, cinc_csinc_64_eq, 2)
+ TEST_INPUTS(
+    //  0x0, 0x1,
+     0xfffffffffffffffc, 0xfffffffffffffffc
+ )
+    // Using x3 register will cause it fail (it gets overwritten, but other registers work...)
```
pgoodman: This should be fixed.

tests/AArch64/BINARY/ANDS_n_LOG_IMM.S
```diff
@@ -0,0 +1,57 @@
...
+
+TEST_BEGIN(ANDS_64S_LOG_IMM, ands_x9_x0_0xfff80007ffffffff, 1)
+TEST_INPUTS(
+    0,
+    1,
+    0xFFFFFFFF,
+    0xfafbfbfd,
+    0xf1f2f3f4,
+    0x0a0b0c0d,
+    0x01020304)
```
pgoodman: I suggest adding in a number here like `0xf....` or `0x7..` so that we can have at least one test that covers `N=1` (result is negative).

PR #154, opened 2017-10-31
Issue 127 missing x86 instructions

mike-myers-tob:
This can be my first proper pull request asking for a review.

If you can look it over and let me know if I made any mistakes in style or data types or anything, that would be great.

Also, if you think this is too big of a pull request and they should be more frequent?


tests/X86/SSE/SQRTSS.S
```diff
@@ -0,0 +1,80 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+TEST_BEGIN(SQRTSSv128v128, 1)
+TEST_INPUTS(
+    0x41100000,   // 9.0, a "perfect square," of which sqrt() is an integer
+    0x00000000,   // sqrt(0) should be 0
```
pgoodman: I suggest adding in negative zero here too.

pgoodman: Easier would be to define a macro with all these SQRT inputs, like:
```c++
#defne SQRT_INPUTS_32 \
  TEST_INPUTS( \
    ...)
```
Then just put `SQRT_INPUTS_32` in each square root test taking 32-bit arguments :-)

tests/X86/SSE/UNPCKHPD.S
```diff
@@ -0,0 +1,67 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+ 
+TEST_BEGIN_64(UNPCKHPDv128v128, 2)
+TEST_INPUTS(
+    0x0000000000000000, 0x0000000000000000,  // sanity check
```
pgoodman: One way to get at "arbitrarily" chosen values is to use what's already pre-loaded in the `xmm`/`ymm` registers. Here is the data: https://github.com/trailofbits/remill/blob/master/tests/X86/Tests.S#L183  And `xmm`/`ymm` zero through seven are initialized with these data: https://github.com/trailofbits/remill/blob/master/tests/X86/Tests.S#L240

Otherwise this is fine, nothing to change.

PR #156, opened 2017-11-05
fix a bug in semantics of POP_GPRv

sdasgup3:
Hello
I tried to test some simple programs and got the following error on lifting to BC stage
```
E1104 14:21:00.103818 10394 Lifter.cpp:119] Cannot lift instruction
at 4005aa, POP_GPRv_58_64 doesn't exist: (AMD64 4005aa 1 (BYTES 5b)
POP_GPRv_58_64 (WRITE_OP (REG_64 RBX)))
```
After consulting with @pgoodman, it seems that the `POP_GPRv` instruction might have some problem. 
After digging a bit, it seems that there is a small bug in the specification of ISEL for pop. That is: the file  `xed/include/xed/xed-iform-enum.h` has the an entry for `XED_IFORM_POP_GPRv_58', but not any for  `XED_IFORM_POP_GPRv_51'

If this feels a valid fix, please accept the pull request.


pgoodman:
I believe you are right about. It turns out that the _51 was a typo in XED: https://github.com/intelxed/xed/commit/8922c2d21ec7110f0ccf33bbdad146d93222918b

pgoodman:
I am not going to merge this yet because we'll need to update our cxx-common packages on s3 to the latest version of XED to pick up the corresponding changes.

Roping in @alessandrogario because we'll need to get those things built and uploaded and he's been doing a lot of work on cxx-common recently.

PR #158, opened 2017-11-08
Adds in explicit tracking of the FPU control word into the State stru…

pgoodman:
…cture. This change helps make sure that VMill can initialize the FPU rounding modes correctly during its emulation. The AArch64 State struct already includes the fpcr.

### DIFF
--- a/remill/Arch/X86/Runtime/State.h
+++ b/remill/Arch/X86/Runtime/State.h
@@ -372,8 +372,6 @@ union XCR0 {
 
 static_assert(8 == sizeof(XCR0), "Invalid packing of `XCR0`.");
 
-
-
 #if COMPILING_WITH_GCC
 using RequestPrivilegeLevel = uint16_t;
 using TableIndicator = uint16_t;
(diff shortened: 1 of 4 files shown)


remill/Arch/X86/Runtime/State.h
```diff
@@ -566,9 +564,11 @@ struct alignas(16) State final : public ArchState {
   MMX mmx;  // 128 bytes.
   FPUStatusFlags sw;  // 8 bytes
   XCR0 xcr0;  // 8 bytes.
+  FPUControlWord fpu_control;  // 2 bytes;
+  uint8_t _0[14];  // 14 bytes.
```
artemdinaburg: Everything looks great; my only side comment is about the padding -- is there any way to make it a macro (e.g. `PAD_16`) or something to show that its there to pad things out to a 16 byte boundary?

pgoodman: would rewriting it as
```c++
uint8_t _padding[14];  // Pad to a 16-byte boundary.
```
be acceptable?

artemdinaburg: Absolutely

PR #163, opened 2017-11-14
Implements more accurate tracking of x87 exception flags, last instru…

pgoodman:
…ction pointer, last data pointer, and floating point instruction opcode.


tests/X86/X87/MISC.S
```diff
@@ -149,11 +149,23 @@ TEST_INPUTS(0)
...
+    0x00000000,   /* sqrt(0) should be 0 */
+    0x412E3D71,   /* 10.89, should return 3.33 */
+    0x411CCCCD,   /* 9.8, whose square root result needs to be rounded */
+    0x7fbfffff,   /* should return the SNaN converted to a QNaN */
+    0xffbfffff,   /* should return the -SNaN converted to a -QNaN */
+    0x7f800000,   /* inf */
+    0xBF800000,   /* -1.0, should return a constant (the QNaN floating-point indefinite) */
+    0xff800000,   /* -inf */
+    0x80000000)    /* -0.0 */
+
```
mike-myers-tob: Are these the SQRT_INPUTS_32 from SQRTSS.S? I wonder if we ought to have sets of edge cases in header files if we will be re-using them across source files?

remill/Arch/X86/Semantics/X87.cpp
```diff
@@ -418,67 +630,130 @@ DEF_ISEL(FIDIVR_ST0_MEMmem16int) = FIDIVR<M16>;
...
+  return FSTP(memory, state, dst, src, pc, fop);
+}
+
+//printf("input=%lx rounded=%lx back=%lx conv_cast=%lx conv_adj=%lx ie=%d pe=%d\n",
+//         reinterpret_cast<uint64_t &>(input),
+//         reinterpret_cast<uint64_t &>(rounded),
+//         reinterpret_cast<uint64_t &>(back),
+//         static_cast<int64_t>(conv_cast), static_cast<int64_t>(conv_adj),
+//         state.sw.ie, state.sw.pe);
+
```
mike-myers-tob: Was this `printf` statement left here commented out for later use?

remill/Arch/X86/Semantics/SSE.cpp
```diff
@@ -1497,25 +1448,23 @@ DEF_HELPER(SquareRoot32, float32_t src_float) -> float32_t {
...
+  } else {  // a number, that is, not a NaN
     // A negative operand (except -0.0) results in the QNaN indefinite value.
-    if (std::signbit(src_float) && src_float != -0.0) {
-      uint32_t indef_qnan = 0xFFC00000;
-      square_root = *reinterpret_cast<float32_t*>(&indef_qnan);
-    }
-    else {
+    if (IsNegative(src_float) && src_float != -0.0) {
+      uint32_t indef_qnan = 0xFFC00000U;
+      square_root = reinterpret_cast<float32_t &>(indef_qnan);
```
mike-myers-tob: Ah, yea, thanks for simplifying my messy cast.

tests/X86/X87/MISC.S
pgoodman: Yes but I think I left out qnan and snan semi on-purpose.

remill/Arch/X86/Semantics/X87.cpp
pgoodman: Good catch.

PR #167, opened 2017-11-29
Issue 150 lift bin dir

pgoodman:


remill/Arch/Runtime/Operators.h
```diff
@@ -1185,6 +1185,14 @@ T _ZeroVec(void) {
       __asm__ __volatile__ ("" ::: "memory"); \
     } while (false)
 
+// A 'compiler' barrier that forced a particular value to be used in a specific
+// spot.
+#define BarrierUsedHere(x) \
```
mike-myers-tob: I'm not sure I understand these compiler barriers. `BarrierReorder` prevents the compiler from reordering of instructions before and after it, but does its effect extend to a certain scope around it? If `BarrierReorder` prevents a reordering, shouldn't `BarrierUsedHere` prevent a value from being used (it seems to do the opposite: force it to be used)?

I think what is needed is a one-or-two line comment for each Barrier operator that explains its purpose: when is it needed? Is it to trick the compiler or the direct the CPU?

Nitpick: esthetically -> aesthetically

tests/AArch64/Run.cpp
```diff
@@ -300,6 +300,7 @@ static void RunWithFlags(const test::TestInfo *info,
   // native program state recorded before executing the native testcase,
   // but after swapping execution to operate on `gStack`.
   if (!sigsetjmp(gJmpBuf, true)) {
+    std::fesetenv(FE_DFL_ENV);
```
mike-myers-tob: Does this require `#pragma STDC FENV_ACCESS ON`?

Do we want to warn if `std::fesetenv()` returns a non-zero value, as in it failed to control the FP environment?

remill/Arch/Runtime/Operators.h
pgoodman: I will document these in my next PR.

tests/AArch64/Run.cpp
pgoodman: Clang doesn't support the pragma, and libc correctly implements the function. It's more that LLVM doesn't yet understand that these functions mutate the floating point environment, and therefore affect optimizations of nearby floating point code.

PR #169, opened 2017-12-02
Issue 150 lift bin dir

pgoodman:


tests/AArch64/Tests.S
```diff
@@ -254,9 +254,12 @@ SYMBOL(__aarch64_test_table_begin):
     0x7ff0000000000000 /* inf */, 0x7ff8000000000000, /* nan */              \
     0x7ff8000000000000 /* nan */, 0x7ff0000000000000 /* inf */
 
+#include "tests/AArch64/DATAXFER/LD2_ASISDLSE_R2.S"
+#include "tests/AArch64/DATAXFER/LD2_ASISDLSEP_R2_R.S"
+#include "tests/AArch64/DATAXFER/LD2_ASISDLSEP_I2_I.S"
 
+#if 0
```
mike-myers-tob: Doesn't this `#if 0` cause all of the below tests to be skipped?

remill/Arch/AArch64/Arch.cpp
```diff
@@ -4045,6 +4106,174 @@ bool TryDecodeSCVTF_D64_FLOAT2INT(const InstData &data, Instruction &inst) {
...
+  }
+  inst.function += "_4S";
+  AddRegOperand(inst, kActionWrite, kRegV, kUseAsValue, data.Rd);
+  AddRegOperand(inst, kActionRead, kRegV, kUseAsValue, data.Rn);
+  return true;
+}
+
+// FMINNMV  <V><d>, <Vn>.<T>
+bool TryDecodeFMINNMV_ASIMDALL_ONLY_H(const InstData &data, Instruction &inst) {
+  return TryDecodeFMAXV_ASIMDALL_ONLY_H(data, inst);
```
mike-myers-tob: These decoders have non-generic names but then they're used generically. What I mean is, e.g., `TryDecodeFMAX` being used to decode `TryDecodeFMIN`. It looked like an error at first but I realize this is just code re-use.

Would it be less error-prone if you always had generically named decoders like `TryDecodeFxxxASIMDALL_ONLY_H`? Not worth it?

remill/Arch/AArch64/Semantics/SIMD.cpp
```diff
@@ -464,3 +513,324 @@ DEF_ISEL(SMAXP_ASIMDSAME_ONLY_4H) = SMAXP_16<V64, int16v4_t>;
...
+//  // (e.g. in the case of +0 and -0).
+//  } else {
+//    auto a = reinterpret_cast<I &>(lhs);
+//    auto b = reinterpret_cast<I &>(rhs);
+//    auto res = SMax(a, b);
+//    return reinterpret_cast<T &>(res);
+//  }
+}
+
+// NOTE(pag): These aren't quite right w.r.t. NaN propagation.
```
mike-myers-tob: Worth filing an issue in GitHub?

tests/AArch64/Tests.S
pgoodman: Yes. I should re-enable it!

remill/Arch/AArch64/Semantics/SIMD.cpp
pgoodman: Added as Issue #170.

remill/Arch/AArch64/Arch.cpp
pgoodman: It's probably worth is as a rule going forward, but it's more busywork for me now. Filed as an enhancement Issue #171.

Issue #168, opened 2017-11-30
More accurate floating-point rounding in X86 FMA extension instructions


mike-myers-tob:
The FMA extension instructions, like `VFMADD213SD`, perform multiplications and additions or subtractions of floating-point values, given several floating point values as operands. When performing, for example, `value2 * value 3 + value 1`, the FMA instruction _supposedly_ performs the multiplication as an "infinite precision intermediate result" before adding the last value and _then_ performing rounding to store the result. We have been unsuccessful in trying to model this.

It does not seem to be an issue of precision and the C++ `double` type. Using boost::multiprecision and its arbitrary-precision binary floating point data type with 100 digits, the result of an operation like this can be calculated with more accuracy. But once converted back down to store in a `double`, the result is exactly the same as doing `double foo = (value2 * value 3 + value 1)` directly. 

It seems like an issue of rounding. The native result and lifted result of the instruction differ by only the least significant bit. The C++ rounding style for floating point is called "to nearest". In testing the behavior of `VFMADD213SD`, it seems to sometimes generate results equivalent to C++ rounding style `FE_DOWNWARD` and other times its results match that of the C++ rounding style `FE_UPWARD`. In other words, controlling `fenv` seems to only help us match its behavior 50% of the time.

ranweiler:
Hi! This bug made me curious, but I'm having trouble reproducing it locally.

Based on [this comment](https://github.com/trailofbits/remill/blob/master/tests/X86/FMA/VFMADDSD.S#L43-L44), my expectation is that adding the inputs `1.2, 2.3, 3.4` to the constant `VFMADDSD_INPUTS_64` should elicit a failure. Is that right? If so, I'm unable to locally repro on an Intel Core i7-6700K on Linux 4.14.5. I also can't repro on CI (example branch build on my account [here](https://travis-ci.org/ranweiler/remill/builds/318779651)).

mike-myers-tob:
Well that's good to hear, and I can revisit this.

I think Peter added something to the test runner that ignores small floating point rounding errors though, could this be covering up the issue? https://github.com/trailofbits/remill/blob/master/tests/X86/Run.cpp#L538

pgoodman:
That's for x87 precision differences, not sse/avx ones.

mike-myers-tob:
For not completely understood reasons, this bug went away on its own (perhaps our migration to a newer LLVM, perhaps our improvements to state modeling). Thanks for noticing @ranweiler !

[...]

PR #197, opened 2017-12-20
Restore previously failing FMA tests

mike-myers-tob:
Relating to issue #168

### DIFF
--- a/tests/X86/FMA/VFMADDSD.S
+++ b/tests/X86/FMA/VFMADDSD.S
@@ -16,19 +16,13 @@
 
 #if HAS_FEATURE_AVX
 
-// We have not yet been successful in modeling the so-called "infinite
-// precision" floating point that the FMA instructions supposedly use to
-// calculate the intermediary results of the fused-multiply operations. This 
-// means that tests with operand values that cannot be precisely stored using
-// binary floating point (e.g., 0.1, as opposed to 0.5 which can be 
-// stored precisely) will fail to match the output of the native FMA
-// instruction -- by exactly 1 bit.
 #define VFMADDSD_INPUTS_64                                             \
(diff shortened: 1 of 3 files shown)


tests/X86/FMA/VFMADDSD.S
```diff
@@ -16,19 +16,13 @@
...
-// means that tests with operand values that cannot be precisely stored using
-// binary floating point (e.g., 0.1, as opposed to 0.5 which can be 
-// stored precisely) will fail to match the output of the native FMA
-// instruction -- by exactly 1 bit.
 #define VFMADDSD_INPUTS_64                                             \
     0.0, 0.0, 0.0,    /* will evaluate to 0.0 */                       \
     1.0, 1.0, 1.0,    /* will evaluate to 2.0 */                       \
     2.0, 3.0, 4.0,    /* will evaluate to 10.0, 11.0, or 14.0 */       \
     2.5, 2.5, 2.5,    /* will evaluate to exactly 8.75 */              \
     1.75, 1.5, -2.25, /* test with negative values */                  \
```
pgoodman: You'll want to test this type of thing in IDA, but these are equivalent to the following in assembly:
```asm
.quad 1.2
```
It's not clear to me that `1.2` will expand into the qword representation that actually represents the `1.2` value of a `double`.

mike-myers-tob: In practice they do (I've tested it at runtime) but if there's a risk they they won't _always_ compile this way, should they be specified as the hex?

pgoodman: TIL :-) I don't think there's a risk because we control the compiler via `cxx-common`.

PR #204, opened 2017-12-21
Adds possibly missing check to mark an instruction as atomic

pgoodman:


remill/Arch/Runtime/Intrinsics.h
```diff
@@ -135,6 +135,31 @@ extern Memory *__remill_atomic_begin(Memory *);
...
+[[gnu::used, gnu::const]]
+extern uint32_t __remill_compare_exchange_memory_16(Memory *, addr_t addr, uint16_t *expected, uint16_t desired);
+
+[[gnu::used, gnu::const]]
+extern uint32_t __remill_compare_exchange_memory_32(Memory *, addr_t addr, uint32_t *expected, uint32_t desired);
+
+[[gnu::used, gnu::const]]
+extern uint32_t __remill_compare_exchange_memory_64(Memory *, addr_t addr, uint64_t *expected, uint64_t desired);
+
+[[gnu::used, gnu::const]]
```
pgoodman: I don't think the non-memory ones should be intrinsics in their own right, instead they should be fully implemented in Operators.h

remill/Arch/Runtime/Intrinsics.cpp
```diff
@@ -63,6 +63,15 @@ extern "C" void __remill_intrinsics(void) {
   USED(__remill_atomic_begin);
   USED(__remill_atomic_end);
 
+  USED(__remill_compare_exchange_memory_8);
+  USED(__remill_compare_exchange_memory_16);
+  USED(__remill_compare_exchange_memory_32);
+  USED(__remill_compare_exchange_memory_64);
+  USED(__remill_compare_exchange_8);
```
pgoodman: Per below, I don't think the non-memory ones should be intrinsics.

remill/Arch/Runtime/Operators.h
```diff
@@ -465,6 +465,27 @@ MAKE_WRITE_REF(float64_t)
 
 #undef MAKE_WRITE_REF
 
+#define MAKE_CMPXCHG(size, type_prefix, access_suffix) \
+    ALWAYS_INLINE static bool _CmpXchg( \
+        Memory *&memory, RnW<type_prefix ## size ## _t> op, \
+        type_prefix ## size ## _t expected, type_prefix ## size ## _t desired) { \
+      return __remill_compare_exchange_ ## access_suffix (memory, op.val_ref, &expected, desired);\
```
pgoodman: Here I think you should just implement the semantics of the non-memory compare exchange.

remill/Arch/Runtime/Operators.h
```diff
@@ -465,6 +465,27 @@ MAKE_WRITE_REF(float64_t)
...
+      return __remill_compare_exchange_memory_ ## access_suffix (memory, op.addr, &expected, desired); \
+    }
+
+MAKE_CMPXCHG(8, uint, 8)
+MAKE_CMPXCHG(16, uint, 16)
+MAKE_CMPXCHG(32, uint, 32)
+MAKE_CMPXCHG(64, uint, 64)
+
+#undef MAKE_CMPXCHG
+#define CmpXchg(op, oldval, newval) _CmpXchg(memory, op, oldval, newval)
```
pgoodman: This should be `UCmpXchg`.

remill/Arch/Runtime/Intrinsics.h
```diff
@@ -135,6 +135,31 @@ extern Memory *__remill_atomic_begin(Memory *);
 [[gnu::used, gnu::const]]
 extern Memory *__remill_atomic_end(Memory *);
 
+[[gnu::used, gnu::const]]
+extern uint32_t __remill_compare_exchange_memory_8(Memory *, addr_t addr, uint8_t *expected, uint8_t desired);
```
pgoodman: These should return a `Memory *`.

remill/Arch/Runtime/Operators.h
```diff
@@ -465,6 +465,27 @@ MAKE_WRITE_REF(float64_t)
...
+    ALWAYS_INLINE static bool _CmpXchg( \
+        Memory *&memory, RnW<type_prefix ## size ## _t> op, \
+        type_prefix ## size ## _t expected, type_prefix ## size ## _t desired) { \
+      return __remill_compare_exchange_ ## access_suffix (memory, op.val_ref, &expected, desired);\
+    } \
+    \
+    ALWAYS_INLINE static bool _CmpXchg( \
+        Memory *&memory, MnW<type_prefix ## size ## _t> op, \
+        type_prefix ## size ## _t expected, type_prefix ## size ## _t desired) { \
+      return __remill_compare_exchange_memory_ ## access_suffix (memory, op.addr, &expected, desired); \
```
pgoodman: I think you want this as
```c++
ALWAYS_INLINE static bool _CmpXchg( \
    Memory *&memory, MnW<type_prefix ## size ## _t> op, \
    type_prefix ## size ## _t *expected, type_prefix ## size ## _t desired) { \
  auto prev_val = *expected;
  memory = __remill_compare_exchange_memory_ ## access_suffix (memory, op.addr, expected, desired);
  return prev_val == *expected;
}
```

pgoodman: So the above will return `true` if it appears that the value was swapped.

remill/Arch/X86/Semantics/SEMAPHORE.cpp
```diff
@@ -22,10 +22,9 @@ namespace {
       auto curr_val = Read(src1); \
       auto desired_val = Read(src2); \
       auto check_val = Read(REG_ ## xax); \
-      auto cmp_res = USub(check_val, curr_val); \
-      auto replace = UCmpEq(cmp_res, 0); \
-      WriteFlagsAddSub<tag_sub>(state, check_val, curr_val, cmp_res); \
-      WriteZExt(dst, Select(replace, desired_val, curr_val)); \
+      auto swap_flag = CmpXchg(dst, check_val, desired_val); \
+      auto replace = UCmpEq(UInt32(swap_flag), UInt32(0)); \
```
pgoodman: `UCmpXchg` will return a `bool`, so you can probably just cast it to an integer.

remill/Arch/X86/Semantics/SEMAPHORE.cpp
```diff
@@ -22,10 +22,9 @@ namespace {
...
+      Write(FLAG_ZF, replace); \
       WriteZExt(REG_ ## xax, Select(replace, check_val, curr_val)); \
```
pgoodman: There is a data race here, because the `Read(src1)` won't be atomic. The `UCmpXchg` that I've proposed above takes `expected` by pointer, and updates it. That is was gets written back to rax.

remill/Arch/Runtime/Operators.h
```diff
@@ -465,6 +465,73 @@ MAKE_WRITE_REF(float64_t)
 
 #undef MAKE_WRITE_REF
 
+#define MAKE_CMPXCHG(size, type_prefix, access_suffix) \
+    template <typename T> \
+    ALWAYS_INLINE static bool _CmpXchg( \
+        Memory *&memory, RnW<T> op, type_prefix ## size ## _t &expected, \
+        type_prefix ## size ## _t desired) { \
+        auto prev_val = expected; \
+        if (*op.val_ref == expected) {\
```
pgoodman: I think what you want is:

```c++
 +        if (*op.val_ref == expected) {\
 +          *op.val_ref = desired; \
 +        } else { \
 +          expected = *reinterpret_cast<type_prefix ## size ## _t*>(op.val_ref); \
 +        }
```

remill/Arch/Runtime/Intrinsics.cpp
```diff
@@ -63,6 +63,11 @@ extern "C" void __remill_intrinsics(void) {
   USED(__remill_atomic_begin);
   USED(__remill_atomic_end);
 
+  USED(__remill_compare_exchange_memory_8);
+  USED(__remill_compare_exchange_memory_16);
+  USED(__remill_compare_exchange_memory_32);
+  USED(__remill_compare_exchange_memory_64);
+
```
pgoodman: Add all the fetch variants as `USED` here.

remill/Arch/Runtime/Operators.h
```diff
@@ -465,6 +465,73 @@ MAKE_WRITE_REF(float64_t)
...
+    make_atomic_intrinsic(name, 64, uint, op) \
+
+MAKE_ATOMIC(fetch_and_add, +, MAKE_ATOMIC_INTRINSIC)
+MAKE_ATOMIC(fetch_and_sub, -, MAKE_ATOMIC_INTRINSIC)
+MAKE_ATOMIC(fetch_and_or, |, MAKE_ATOMIC_INTRINSIC)
+MAKE_ATOMIC(fetch_and_and, &, MAKE_ATOMIC_INTRINSIC)
+MAKE_ATOMIC(fetch_and_xor, ^, MAKE_ATOMIC_INTRINSIC)
+//MAKE_ATOMIC(fetch_and_nand, MAKE_ATOMIC_INTRINSIC)
+
+#define fetch_and_add(op1, op2) _fetch_and_add(memory, op1, op2)
```
pgoodman: Can you make these names follow the naming convention? E.g. `UFetchAdd16`, for example.

remill/Arch/Runtime/Operators.h
```diff
@@ -465,6 +465,111 @@ MAKE_WRITE_REF(float64_t)
...
+#define MAKE_CMPXCHG(size, type_prefix, access_suffix) \
+    template <typename T> \
+    ALWAYS_INLINE static bool _CmpXchg( \
+        Memory *&memory, RnW<T> op, type_prefix ## size ## _t &expected, \
+      type_prefix ## size ## _t desired) { \
+      if (*op.val_ref == expected) {\
+        *op.val_ref = desired; \
+        return true; \
+      } else { \
+        expected = *reinterpret_cast<type_prefix ## size ## _t*>(op.val_ref); \
```
pgoodman: This `reinterpret_cast` is redundant.

remill/Arch/Runtime/Operators.h
```diff
@@ -465,6 +465,111 @@ MAKE_WRITE_REF(float64_t)
...
+MAKE_CMPXCHG(16, uint, 16)
+MAKE_CMPXCHG(32, uint, 32)
+MAKE_CMPXCHG(64, uint, 64)
+MAKE_CMPXCHG(128, uint, 128)
+
+#undef MAKE_CMPXCHG
+#define UCmpXchg(op, oldval, newval) _CmpXchg(memory, op, oldval, newval)
+
+#define MAKE_ATOMIC_ACCESSOR(type_prefix, size) \
+    ALWAYS_INLINE static Memory *UFetchAdd ## size( \
```
pgoodman: Should be `_`-prefixed, kind of a way to tell people not to use these :-P

remill/Arch/Runtime/Operators.h
```diff
@@ -498,16 +498,49 @@ MAKE_CMPXCHG(128, uint, 128)
 #undef MAKE_CMPXCHG
 #define UCmpXchg(op, oldval, newval) _CmpXchg(memory, op, oldval, newval)
 
+#define MAKE_ATOMIC_ACCESSOR(type_prefix, size) \
```
pgoodman: I think you scan simplify a bunch of these various atomic operation-making macros into:
```c++
#define MAKE_ATOMIC_INTRINSIC(name, intrinsic_name, size, type_prefix, op)\
  template<typename T> \
  ALWAYS_INLINE type_prefix ## size ## _t _U ## name( \
      Memory *&memory, MnW<T> addr, type_prefix ## size ## _t &value) { \
    memory = intrinsic_name ## size(memory, addr.addr, value); \
    return value; \
  } \
  \
  template<typename T> \
  ALWAYS_INLINE type_prefix ## size ## _t _U ## name ( \
      Memory *&memory, RnW<T> addr, type_prefix ## size ## _t &value) { \
    auto prev_value = *reinterpret_cast<type_prefix ## size ## _t*>(addr.val_ref); \
    *addr.val_ref = prev_value op value; \
    value = prev_value; \
    return value; \
  }

#define MAKE_ATOMIC(name, op)  \
    MAKE_ATOMIC_INTRINSIC(name, 8, uint, op) \
    MAKE_ATOMIC_INTRINSIC(name, 16, uint, op) \
    MAKE_ATOMIC_INTRINSIC(name, 32, uint, op) \
    MAKE_ATOMIC_INTRINSIC(name, 64, uint, op) \

MAKE_ATOMIC(FetchAdd, +)
MAKE_ATOMIC(FetchSub, -)
MAKE_ATOMIC(FetchOr, |)
MAKE_ATOMIC(FetchAnd, &)
MAKE_ATOMIC(FetchXor, ^)

#undef MAKE_ATOMIC
#undef MAKE_ATOMIC_INTRINSIC
```

pgoodman: updated with some fixes

remill/Arch/Runtime/Operators.h
kumarak: It is to suppress warning about casting unsigned long to uint32_t.

pgoodman:
@kumarak ready to merge? The change set looks good to me :-)

kumarak:
I see some of the test cases failure after merging. The arguments are different than the value in the intrinsic functions for some test-cases.

pgoodman:
Try removing `gnu::pure` entirely.

tests/AArch64/Run.cpp
```diff
@@ -152,6 +152,71 @@ NEVER_INLINE Memory *__remill_write_memory_f80(Memory *, addr_t, float64_t) {
...
+Memory *__remill_compare_exchange_memory_64(
+    Memory *memory, addr_t addr, uint64_t &expected, uint64_t desired) {
+  expected = __sync_val_compare_and_swap(
+      reinterpret_cast<uint64_t *>(addr), expected, desired);
+  return memory;
+}
+
+Memory *__remill_compare_exchange_memory_128(
+    Memory *memory, addr_t addr, uint128_t &expected, uint128_t &desired) {
+#ifdef _GXX_EXPERIMENTAL_CXX0X__
```
pgoodman: Use `#if __has_builtin(__sync_val_compare_and_swap`

pgoodman: Also `_GXX_EXPERIMENTAL_CXX0X__` is missing a additional leading underscore, it should be `__GXX_EXPERIMENTAL_CXX0X__`. In this situation, failure of this macro to be set should be handled by an `#else`, perhaps with an `#error`. But this is a pretty GNU libstdc++-specific macro that I don't think we want to depend upon.

tests/X86/Run.cpp
```diff
@@ -213,9 +213,29 @@ Memory *__remill_compare_exchange_memory_64(
 
 Memory *__remill_compare_exchange_memory_128(
     Memory *memory, addr_t addr, uint128_t &expected, uint128_t &desired) {
-#ifdef _GXX_EXPERIMENTAL_CXX0X__
+#if !(defined(__x86_64__) || defined(__i386__) || defined(_M_X86))
   expected = __sync_val_compare_and_swap(
       reinterpret_cast<uint128_t *>(addr), expected, desired);
+#else
+  bool result;
+  struct uint128 {
```
pgoodman: You can use `struct alignas(16) uint128` in C++11.

tests/X86/Run.cpp
```diff
@@ -213,9 +213,29 @@ Memory *__remill_compare_exchange_memory_64(
...
+    uint64_t lo;
+    uint64_t hi;
+  } __attribute__ ((__aligned__( 16 )));
+
+  uint128 *oldval = reinterpret_cast<uint128*>(&expected);
```
pgoodman: `uint128 *` (space before the `*`).

PR #206, opened 2017-12-21
AArch64 EXT insn implementation, with test

GarretReece:
AArch64 EXT instruction implementation

### DIFF
--- a/remill/Arch/AArch64/Semantics/SIMD.cpp
+++ b/remill/Arch/AArch64/Semantics/SIMD.cpp
@@ -863,9 +863,11 @@ DEF_SEM(EXT, V128W dst, T src1, T src2, I32 src3) {
   auto vn = UReadV8(src1);
   auto vm = UReadV8(src2);
   uint8v16_t result = {};
+  _Pragma("unroll")
   for (size_t i = 0, max_i = count; i+lsb < max_i; ++i) {
     result.elems[count-1-i] = UExtractV8(vm, i+lsb);
   }
+  _Pragma("unroll")
   for (size_t i = lsb; i < count; ++i) {
     result.elems[count-1-i] = UExtractV8(vn, i-lsb);
   }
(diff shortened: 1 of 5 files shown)


remill/Arch/AArch64/Semantics/SIMD.cpp
```diff
@@ -855,6 +855,34 @@ DEF_SEM(NOT_8, V128W dst, S src) {
 DEF_ISEL(NOT_ASIMDMISC_R_8B) = NOT_8<V64>; 
 DEF_ISEL(NOT_ASIMDMISC_R_16B) = NOT_8<V128>; 
 
+namespace {
+
+#define MAKE_EXT(size, count) \
```
pgoodman: Here I'd recommend using a template rather than a macro, and to get the `count`, where you could have a source type `V64` in one case, and `V128` in another, and you can also pass in a `size_t count` in the template parameter list, or infer `count` using `NumVectorElements(vn)`.

remill/Arch/AArch64/Semantics/SIMD.cpp
```diff
@@ -855,6 +855,29 @@ DEF_SEM(NOT_8, V128W dst, S src) {
...
 
+namespace {
+
+template <typename T, size_t count>
+DEF_SEM(EXT, V128W dst, T src1, T src2, I32 src3) {
+  auto lsb = Read(src3);
+  auto vn = UReadV8(src1);
+  auto vm = UReadV8(src2);
+  uint8v16_t result = {};
+  for (size_t i = 0, max_i = count; i+lsb < max_i; ++i) {
```
pgoodman: `_Pragma("unroll")`

remill/Arch/AArch64/Semantics/SIMD.cpp
```diff
@@ -855,6 +855,29 @@ DEF_SEM(NOT_8, V128W dst, S src) {
...
+    result.elems[count-1-i] = UExtractV8(vm, i+lsb);
+  }
+  for (size_t i = lsb; i < count; ++i) {
```
pgoodman: `_Pragma("unroll")`

PR #209, opened 2017-12-22
Use GCC/Clang intrinsics for CLZ and CTZ

ESultanik:
Clang supports `__builtin_clzll`. We should use it!


remill/Arch/AArch64/Arch.cpp
```diff
@@ -729,6 +729,14 @@ static void AddPostIndexMemOp(Instruction &inst, Action action,
 }
 
 static bool MostSignificantSetBit(uint64_t val, uint64_t *highest_out) {
+  #if defined(__GCC__) || defined(__clang__)
```
pgoodman: Can you bring the `#if` to have no indentation, and use `__has_builtin(__builtin_clzll)`. For example: https://godbolt.org/g/txSAup

remill/Arch/AArch64/Arch.cpp
```diff
@@ -737,16 +745,26 @@ static bool MostSignificantSetBit(uint64_t val, uint64_t *highest_out) {
     }
   }
   return found;
+  #endif
 }
 
 static bool LeastSignificantSetBit(uint64_t val, uint64_t *highest_out) {
+  #if defined(__GCC__) || defined(__clang__)
```
pgoodman: Outdent the `#if`.

ESultanik:
@pgoodman I've outdented the `#if` blocks and switched to the `__has_builtin` extension.

Issue #187, opened 2017-12-12
Differences in CPU behavior cause failing x86 tests on Travis CI


mike-myers-tob:
Many of the Remill instruction tests fail on Travis, and not even deterministically. They do not fail locally on developer machines, so we have been ignoring the issue and letting it annoy us for a while now.

Sometimes all of the tests pass. Re-running the build job myself tonight I have seen 100% pass, one instruction test fail, or 3 fail, or so many fail that it stopped running tests.

### Usually failing
`FTST`

### Occasionally failing
 `TZCNT` and `LZCNT`

### Seen failing at least once
The floating-point conditional move instructions (all of these tests failed): `FCMOVNU`, `FCMOVNB`, `FCMOVNE`, `FCMOVBE`, `FCMOVNBE`, `FCMOVU`, etc.

## Tips for Troubleshooting on Travis CI
We have enabled "debug builds" on our Travis CI repo. By using the API, you can trigger a debug build session. Use your own authorization token in place of the `**********`. Use your own Job ID in place of the `314970938`.

```shell
curl -s -X POST \ 
  -H "Content-Type: application/json" \ 
  -H "Accept: application/json" \ 
  -H "Travis-API-Version: 3" \ 
  -H "Authorization: token *************" \
   -d "{\"quiet\": true}" \   https://api.travis-ci.org/job/314970938/debug
```

Then wait multiple minutes for a response from the API (sometimes it does not respond, just do it again). Check the status on https://travis-ci.org/trailofbits/remill/builds/ as well, it will eventually trigger and say "pending". Using the web UI, observe the build log for the line with SSH instructions.

SSH in, and run the following commands to perform the Travis build steps and debug the tests:
```shell
travis_run_before_install
travis_run_install
travis_run_before_script
travis_run_script
sudo apt-get install gdb
gdb ./tests/X86/run-amd64_avx-tests
```

As usual, you definitely want to comment out some of the tests rather than build them all.

Once in GDB, you can run steps similar to this:
```gdb
handle SIGFPE nostop noprint
set $native = (State *)&gNativeState
set $lifted = (State *)&gLiftedState
b /home/travis/build/trailofbits/remill/tests/X86/Run.cpp:564
run
p/x $native->rflag
p/x $lifted->rflag
```

To scroll up/down in the terminal session you have to use the awful `ctrl-b [` and then use the arrow keys until you want to stop with `q`. Suggest you just view the log in the Travis CI web UI.

[...]

PR #210, opened 2017-12-23
Remove illegal instruction from tests

### LINKED ISSUE #168: More accurate floating-point rounding in X86 FMA extension instructions

### LINKED ISSUE #205: False positives in tests
## Repro

Pick a semantic def for an instruction (whose tests pass!) and make it equivalent to a nop. Example: https://github.com/ranweiler/remill/commit/779b4ec3e239d84282571674804d1fa8c66b1358

## Expected

In the above case, I would (at a minimum) expect the tests in `tests/X86/DATAXFER/MOVD.S` to fail. I would also expect a failure from any other test which implicitly tests the correctness of the `movd` instruction.

## Observed

No tests fail at all, either locally (64-bit Linux 4.14) or in CI, though they seem to be run. See the Travis job log here: https://travis-ci.org/ranweiler/remill/builds/319817422

## Note

I initially observed this when trying to add an initial failing test case for some x87 instructions, and at first thought the scope of this issue might be limited to the test harness's comparisons of native/lifted FPU state (my first repro was after making `fild` a nop). However, it seems to be a broader issue.

ranweiler:
- Closes #205
- Reverts the commit that closed #168 (see [here](https://github.com/trailofbits/remill/issues/168#issuecomment-353503497))
- Includes segment registers in saved test state, as suggested [here](https://github.com/trailofbits/remill/issues/205#issuecomment-353658141)

Tests pass locally for me, but CI fails right now, likely due to #187 being hidden by #205.


pgoodman:
I'm seeing lots of failures for `fcmov`-related instructions on Travis, e.g.
```
E1223 02:31:53.966231 10449 Run.cpp:566] States did not match for FCMOVNU_1 with ARG1=0x0 and CF=0 PF=0 AF=0 ZF=0 SF=0 DF=0 OF=0
/home/travis/build/trailofbits/remill/tests/X86/Run.cpp:568: Failure
Value of: !"Lifted and native states did not match."
  Actual: false
Expected: true
```

pgoodman:
I think what is happening is that the `DEF_FPU_SEM` macro is updating the fpu `cs`, and that maybe is not quite right anymore?

ranweiler:
@pgoodman I just did one last double-check on my local box right now with fully-rebuilt Docker images, and actually am getting some of the CI errors— I think I had some stale state when locally testing the last commit. I'm going to try it (a) without checking the CS register, (b) without saving _any_ segment registers.

pgoodman:
This could be an artefact of the test runner being a 64-bit program, even when running the 32-bit tests. Specifically, we'll need to figure out if the running `fxsave` in a 32-bit program saves the cs, where the equivalent `fxsave` in a 64-bit program doesn't.

pgoodman:
What I'm getting at is that, the `fxsave64` format doesn't record the `cs`, so maybe behind the scenes, 64-bit mode doesn't record the `cs` either, and so it just zeroes it out when using `fxsave`.

ranweiler:
@pgoodman that's consistent with what I see locally on a totally fresh image: I'm saving all the segment registers except CS, and all tests pass. I could amend the last commit to save all the seg registers except CS, and we could open a separate issue for the 32/64-bit(?) CS situation.

ranweiler:
Okay, not including the CS register gets rid of almost all of the CI test failures (on _that_ run, anyway). Locally, I observe no test failures if I exclude the CS register.

ranweiler:
I've amended the last commit to exclude CS, which ensures we have only CI failures for now.

pgoodman:
I figure this is the source of the issue:
image

I think the "right" thing to do is re-add your saving of `cs` in the PrintSaveState.cpp, and then perhaps in the Run.cpp, what we can do is a `cpuid` check on whether or not we should bother checking `cs` in the `fxsave`, and if not, zero out the lifted version (which does save it).

tests/X86/Run.cpp
```diff
@@ -326,6 +326,38 @@ static void InitFlags(void) {
...
 
+// Check if we are in a mode such that FCS and FDS are deprecated, and
+// are thus zeroed out in FXSAVE, XSAVE, and XSAVEOPT.
+//
+// Per the Intel SDM Vol. 1, Section 8.1.8, this happens when:
+//
+//    CPUID.(EAX=07H,ECX=0H):EBX[bit 13] = 1
+//
+// Where "bit 13" is a 0-based index.
+static bool AreFCSAndFDSDeprecated() {
```
pgoodman: `(void)` as param list.

tests/X86/Run.cpp
```diff
@@ -458,6 +490,14 @@ static void RunWithFlags(const test::TestInfo *info,
   memset(lifted_state->x87.fxsave.st, 0, kill_size);
   memset(native_state->x87.fxsave.st, 0, kill_size);
 
+#if 32 == ADDRESS_SIZE_BITS
+  // If FCS and FDS are deprecated, don't compare them.
+  if (AreFCSAndFDSDeprecated()) {
```
pgoodman: `if (32 == ADDRESS_SIZE_BITS && ...)`, that way we don't get a compiler warning about `AreFCSAndFDSDeprecated` being unused.

tests/X86/Run.cpp
```diff
@@ -458,6 +490,14 @@ static void RunWithFlags(const test::TestInfo *info,
   memset(lifted_state->x87.fxsave.st, 0, kill_size);
   memset(native_state->x87.fxsave.st, 0, kill_size);
 
+#if 32 == ADDRESS_SIZE_BITS
```
pgoodman: I meant to remove the `#if` and just depend on the actual `if` :-P

ranweiler: Oh, that makes more sense (-:

The reason I used `#if` is because the referenced fields don't exist in 64-bit mode, so you get a compiler error.

pgoodman: Oh duh! I didn't make that connection.. lol. So then the additional check in the `if` is unnecessary.

ranweiler: Apparently not? One moment...

ranweiler: Nope, yes, the `#if` is necessary, I just assumed you were talking about some other warning I missed. Reverting the redundant expression.

PR #215, opened 2017-12-27
Fix #if condition for the `ctz` builtin

ESultanik:
Minor nitpick, but I just realized that I was checking for `__builtin_clzll` where I should have been checking for `__builtin_ctzll`. Hence a 1 character pull request. We can delete this branch after it's merged in :-) Relevant commit is 8e39450.


remill/Arch/AArch64/Arch.cpp
```diff
@@ -737,16 +745,26 @@ static bool MostSignificantSetBit(uint64_t val, uint64_t *highest_out) {
     }
   }
   return found;
+#endif
 }
 
 static bool LeastSignificantSetBit(uint64_t val, uint64_t *highest_out) {
+#if __has_builtin(__builtin_ctzll)
+  if (val) {
+    *highest_out = __builtin_ctzll(val);
```
pgoodman: Given that above you have a `sizeof(unsigned long long)`, which is there to kind of handle the case that it's possibly a different size than `uint64_t`, then I think you should add in a

```c++
static_assert(sizeof(unsigned long long) >= sizeof(uint64_t),
              "Invalid use of `__builtin_clzll` on a `uint64_t`.");
```

remill/Arch/AArch64/Arch.cpp
```diff
@@ -737,16 +745,26 @@ static bool MostSignificantSetBit(uint64_t val, uint64_t *highest_out) {
...
+  }
+#else
   for (uint64_t i = 0; i < 64; ++i) {
     if ((val >> i) & 1) {
       *highest_out = i;
       return true;
     }
   }
   return false;
+#endif
```
pgoodman: Add trailing  comment, `#endif  // __has_builtin(__builtin_ctzll)`

remill/Arch/AArch64/Arch.cpp
```diff
@@ -737,16 +745,26 @@ static bool MostSignificantSetBit(uint64_t val, uint64_t *highest_out) {
     }
   }
   return found;
+#endif
```
pgoodman: Add trailing comment, `#endif  // __has_builtin(__builtin_ctzll)`

PR #217, opened 2018-01-05
Issue 150: Implement USHR_ASISDSHF_R

ehennenfent:
Also includes prototype decoder for USHR_ASIMDSHF_R, but no semantic implementation
(fails Travis, but for the same reason as master)


remill/Arch/AArch64/Arch.cpp
```diff
@@ -4656,6 +4656,39 @@ bool TryDecodeMVNI_ASIMDIMM_M_SM(const InstData &data, Instruction &inst) {
...
+  AddRegOperand(inst, kActionWrite, kRegV, kUseAsValue, data.Rd);
+  AddRegOperand(inst, kActionRead, kRegV, kUseAsValue, data.Rn);
+  AddImmOperand(inst, shift);
+  return true;
+}
+
+// USHR  <Vd>.<T>, <Vn>.<T>, #<shift>
+bool TryDecodeUSHR_ASIMDSHF_R(const InstData &data, Instruction &inst) {
+  return false; // TODO remove this after adding semantics for vector version
+  if (((data.immh.uimm & 8) == 0) && !data.Q) {
```
pgoodman: Shouldn't that be `!= 0`?

remill/Arch/AArch64/Arch.cpp
```diff
@@ -4656,6 +4656,39 @@ bool TryDecodeMVNI_ASIMDIMM_M_SM(const InstData &data, Instruction &inst) {
   return true;
 }
 
+// USHR  <V><d>, <V><n>, #<shift>
+bool TryDecodeUSHR_ASISDSHF_R(const InstData &data, Instruction &inst) {
+  if ((data.immh.uimm & 8) == 0) {
```
pgoodman: Shouldn't this be `if (0 != (data.immh.uimm & 8) && !data.Q)`?

ehennenfent: The logic here is correct, but I copied the wrong comment from the documentation. Fixed in c6035e3.

Issue #223, opened 2018-01-16
Fully model X87 handling of SNaN/QNaN, other exceptional inputs


ranweiler:
While adding X87 instructions, I've found that we don't model (or sometimes test) various input classes for X87 semantics. This surfaced when I started adding more exhaustive test inputs for the BCD instructions.

Much of this relates to inaccurate modeling of the pushing SNaNs onto the X87 stack. The X87 stack does not appear to typically store SNaNs; at least, they are quietized by `fld` before being pushed. Our `fld` semantic pushes an explicitly _signaling_ NaN onto the X87 stack.

Even after fixing this, in our semantics tests, there are cases where we push an SNaN onto the X87 stack (after being quietized), which sets the IE bit. Then we perform an op with ST(0) whose semantic has a line like `state.sw.ie = IsSignalingNaN(val);`, which always _clears_ the IE flag, since `val` was quietized when pushed. What we should do is `|=`, to avoid clearing IE if it was set. See `fsqrt`, `frndint`, &c.

Right now, I think we have a situation where the two modeling errors cancel each other out in the existing semantics tests that _are_ sensitive to this. I'm also wondering if this relates to the Travis-only errors we're seeing with `ftst`, but as I write this issue, Travis is having service issues, so I haven't been able to test that at all.

ranweiler:
I'd like to wrap up the FBLD/FBSTP impls in #219, but this is a bit of cross-cutting issue, and it might make sense to address first. Otherwise I'll have to do somethign like _not_ test SNaNs, or cut out a partial fix in that PR to get tests all passing.

ranweiler:
This is mostly done in #224, but it needs one more pass to add NaN test cases to _all_ the instructions that consume the X87 stack. Some of the existing tests (e.g. for `fyl2x`) just use `fldpi` and `fldlg2` to set up happy path test inputs.

PR #224, opened 2018-01-17
Improve modeling of X87 exception and NaN handling

ranweiler:
For #223.

- [x] Add NaN test inputs to surface inaccurate modeling
- [x] Ensure X87 exceptions flags are sticky
- [x] Ensure SNaNs are quietized before pushing onto X87 stack

### DIFF
--- a/remill/Arch/X86/Semantics/FLAGS.cpp
+++ b/remill/Arch/X86/Semantics/FLAGS.cpp
@@ -202,12 +202,13 @@ struct Carry<tag_sub> {
     } while (false)
 
 
+// X87 status flags are sticky, so we must not unset flags if set.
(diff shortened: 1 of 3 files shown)


ranweiler:
Okay, this is ready for review. There is some repetition of code right now in e.g. FSIN, FCOS, but first attempts at DRYing things up didn't seem to really reduce the actual complexity. I'd rather take another pass at that later when e.g. improving the implementation of `IsImprecise()`.

remill/Arch/X86/Semantics/FLAGS.cpp
```diff
@@ -202,12 +202,13 @@ struct Carry<tag_sub> {
     } while (false)
 
 
+// X87 status flags are sticky, so we must not unset flags if set.
```
pgoodman: TIL, I did not realize they were cumulative.

PR #255, opened 2018-05-18
CR8 read/write support

tathanhdinh:
Hello all,

The `cr8` register contains `Task Priority Register` (`TPR`) which is used by kernel code to set the interrupt priority, for example the piece of code:

```
44 0f 20 c0             mov rax, cr8
ba 02 00 00 00          mov edx, 0x2
44 0f 22 c2             mov cr8, rdx
```
used by Windows kernel to set `TPR`.

I have used the same principle (applied by `remill` for other control registers `CR0..4`) to support the read/write on `cr8`.


pgoodman:
You'll need to add a definition of `gCR8` into the Run.cpp of the x86 test suite, and can you submit a similar PR to the McSema repo with a definition of `gCR8` in the x86 Runtime.cpp file? Besides that, this looks good.

Also, do CR5, CR6, and CR7 exist? Should they be defined, or should I expect them from you in future PRs?

pgoodman:
WHoops I must have missed that! And yes, just add definitions to McSema's runtime, really just to make sure the symbols are defined and linkable.

PR #257, opened 2018-05-25
Dead store elimination (draft)

alberdingk-thijm:
Draft PR for dead-store elimination code.

### DIFF
--- a/remill/BC/DeadStoreEliminator.cpp
+++ b/remill/BC/DeadStoreEliminator.cpp
@@ -105,6 +105,33 @@ enum class VisitResult {
   Error
 };
 
+enum class OpType {
+  Plus,
+  Minus,
+};
+
+// Get the unsigned offset of two int64_t numbers with bounds checking
+bool GetUnsignedOffset(int64_t v1, int64_t v2, OpType op, int64_t max, uint64_t *result) {
+  auto signed_result = v1;
+  //LOG(INFO) << "v1: " << v1 << (op == OpType::Plus ? " + " : " - ") << "v2: " << v2;
+  switch (op) {
+    case OpType::Plus:
+      signed_result += v2;
+      break;
+    case OpType::Minus:
+      signed_result -= v2;
+      break;
+    default:
+      break;
+  }
+  if (signed_result >= 0 && signed_result < max) {
+    *result = static_cast<uint64_t>(signed_result);
+    return true;
+  } else {
+    return false;
+  }
+}
+
 ForwardAliasVisitor::ForwardAliasVisitor(const std::vector<StateSlot> &state_slots_,
     llvm::DataLayout *dl_,
     llvm::Value *sp_)
@@ -231,7 +258,7 @@ VisitResult ForwardAliasVisitor::visitGetElementPtrInst(llvm::GetElementPtrInst 
       return VisitResult::NoProgress;
     } else {
       // get the constant offset
-      llvm::APInt const_offset;
+      llvm::APInt const_offset(64, 0);
       if (I.accumulateConstantOffset(*dl, const_offset)) {
         // the final offset (adding the ptr->second value to the const_offset)
         uint64_t offset = 0;
@@ -279,27 +306,6 @@ VisitResult ForwardAliasVisitor::visitAdd(llvm::BinaryOperator &I) {
 // Visit a sub instruction and update the offset map.
 VisitResult ForwardAliasVisitor::visitSub(llvm::BinaryOperator &I) {
   return ForwardAliasVisitor::visitBinaryOp_(I, OpType::Minus);
-}
-
-// Get the unsigned offset of two int64_t numbers with bounds checking
-bool GetUnsignedOffset(int64_t v1, int64_t v2, OpType op, int64_t max, uint64_t *result) {
-  auto signed_result = v1;
-  switch (op) {
-    case OpType::Plus:
-      signed_result += v2;
-      break;
-    case OpType::Minus:
-      signed_result -= v2;
-      break;
-    default:
-      break;
-  }
-  if (signed_result >= 0 && signed_result < max) {
-    *result = static_cast<uint64_t>(signed_result);
-    return true;
-  } else {
-    return false;
-  }
 }
 
 // Visit an add or sub instruction.
@@ -407,8 +413,8 @@ VisitResult ForwardAliasVisitor::visitPHINode(llvm::PHINode &I) {
 
 // For each instruction in the alias map, add an AAMDNodes struct 
 // which specifies the aliasing stores and loads to the instruction's byte offset.
-void AddAAMDNodes(const ValueToOffset &inst_to_offset, const std::vector<llvm::AAMDNodes> &offset_to_aamd) {
-  for (const auto &map_pair : inst_to_offset) {
+void AddAAMDNodes(const ValueToOffset &val_to_offset, const std::vector<llvm::AAMDNodes> &offset_to_aamd) {
+  for (const auto &map_pair : val_to_offset) {
     auto val = map_pair.first;
     auto offset = map_pair.second;
     if (auto load_inst = llvm::dyn_cast<llvm::LoadInst>(val)) {
--- a/remill/BC/DeadStoreEliminator.cpp
+++ b/remill/BC/DeadStoreEliminator.cpp
@@ -216,10 +216,9 @@ static llvm::MDNode *GetScopeFromInst(llvm::Instruction &inst) {
   return inst.getMetadata(llvm::LLVMContext::MD_alias_scope);
 }
 
-static LiveSet GetLiveSetFromArgs(
-    llvm::iterator_range<llvm::Use *> args,
-    ValueToOffset val_to_offset,
-    const std::vector<StateSlot> state_slots) {
+static LiveSet GetLiveSetFromArgs(llvm::iterator_range<llvm::Use *> args,
+                                  ValueToOffset val_to_offset,
+                                  const std::vector<StateSlot> state_slots) {
   LiveSet live;
   for (auto &arg_it : args) {
     auto arg = arg_it->stripPointerCasts();
@@ -252,7 +251,8 @@ struct ForwardAliasVisitor
   virtual ~ForwardAliasVisitor(void) = default;
 
   ForwardAliasVisitor(const llvm::DataLayout &dl_,
-                      const std::vector<StateSlot> &state_slots_);
+                      const std::vector<StateSlot> &state_slots_,
+                      std::unordered_map<llvm::Instruction *, LiveSet> &live_args_);
 
   bool Analyze(llvm::Function *func);
 
@@ -269,6 +269,8 @@ struct ForwardAliasVisitor
   virtual VisitResult visitSub(llvm::BinaryOperator &I);
   virtual VisitResult visitSelect(llvm::SelectInst &inst);
   virtual VisitResult visitPHINode(llvm::PHINode &I);
+  virtual VisitResult visitCallInst(llvm::CallInst &inst);
+  virtual VisitResult visitInvokeInst(llvm::InvokeInst &inst);
 
  private:
   void AddInstruction(llvm::Instruction *inst);
@@ -279,17 +281,21 @@ struct ForwardAliasVisitor
   const std::vector<StateSlot> &state_slots;
   ValueToOffset state_offset;
   InstToOffset state_access_offset;
+  std::unordered_map<llvm::Instruction *, LiveSet> live_args;
   std::unordered_set<llvm::Value *> exclude;
   std::vector<llvm::Instruction *> curr_wl;
   llvm::Value *state_ptr;
 };
 
 ForwardAliasVisitor::ForwardAliasVisitor(
-    const llvm::DataLayout &dl_, const std::vector<StateSlot> &state_slots_)
+    const llvm::DataLayout &dl_,
+    const std::vector<StateSlot> &state_slots_,
+    std::unordered_map<llvm::Instruction *, LiveSet> &live_args_)
     : dl(dl_),
       state_slots(state_slots_),
       state_offset(),
       state_access_offset(),
+      live_args(live_args_),
       exclude(),
       curr_wl(),
       state_ptr(nullptr) {}
@@ -680,6 +686,26 @@ VisitResult ForwardAliasVisitor::visitPHINode(llvm::PHINode &I) {
   }
 }
 
+VisitResult ForwardAliasVisitor::visitCallInst(llvm::CallInst &inst) {
+  // If we have not seen this instruction before, add it.
+  if (!live_args.count(&inst)) {
+    auto args = inst.arg_operands();
+    auto live = GetLiveSetFromArgs(args, state_offset, state_slots);
+    live_args.emplace(&inst, live);
+  }
+  return VisitResult::Ignored;
+}
+
+VisitResult ForwardAliasVisitor::visitInvokeInst(llvm::InvokeInst &inst) {
+  // If we have not seen this instruction before, add it.
+  if (!live_args.count(&inst)) {
+    auto args = inst.arg_operands();
+    auto live = GetLiveSetFromArgs(args, state_offset, state_slots);
+    live_args.emplace(&inst, live);
+  }
+  return VisitResult::Ignored;
+}
+
 // Back-and-forth mapping between LLVM meta-data node that we create per slot,
 // and `StateSlot`s.
 class AAMDInfo {
@@ -870,8 +896,11 @@ bool LiveSetBlockVisitor::VisitBlock(llvm::BasicBlock *block) {
       } else {
         args = llvm::dyn_cast<llvm::InvokeInst>(inst)->arg_operands();
       }
-      
-      live |= GetLiveSetFromArgs(args, val_to_offset, state_slots);
+
+      auto arg_live = GetLiveSetFromArgs(args, val_to_offset, state_slots);
+      live |= arg_live;
+      // Since `GetLiveSetFromArgs` only sets bits, this is safe to do.
+      func_used |= arg_live;
 
     } else if (auto store_inst = llvm::dyn_cast<llvm::StoreInst>(inst)) {
       auto scope = GetScopeFromInst(*inst);
@@ -1390,6 +1419,8 @@ void RemoveDeadStores(llvm::Module *module,
   const AAMDInfo aamd_info(slots, module->getContext());
   const llvm::DataLayout dl = module->getDataLayout();
 
+  std::unordered_map<llvm::Instruction *, LiveSet> live_args = {};
+
   for (auto &func : *module) {
     if (&func == bb_func ||
         func.isDeclaration() ||
@@ -1397,7 +1428,7 @@ void RemoveDeadStores(llvm::Module *module,
       continue;
     }
 
-    ForwardAliasVisitor fav(dl, slots);
+    ForwardAliasVisitor fav(dl, slots, live_args);
 
     // If the analysis succeeds for this function, add the AAMDNodes.
     if (fav.Analyze(&func)) {
--- a/remill/BC/DeadStoreEliminator.cpp
+++ b/remill/BC/DeadStoreEliminator.cpp
@@ -218,8 +218,8 @@ static llvm::MDNode *GetScopeFromInst(llvm::Instruction &inst) {
 }
 
 static LiveSet GetLiveSetFromArgs(llvm::iterator_range<llvm::Use *> args,
-                                  ValueToOffset val_to_offset,
-                                  const std::vector<StateSlot> state_slots) {
+                                  const ValueToOffset &val_to_offset,
+                                  const std::vector<StateSlot> &state_slots) {
   LiveSet live;
   for (auto &arg_it : args) {
     auto arg = arg_it->stripPointerCasts();
@@ -285,6 +285,7 @@ struct ForwardAliasVisitor
   InstToLiveSet live_args;
   std::unordered_set<llvm::Value *> exclude;
   std::vector<llvm::Instruction *> curr_wl;
+  std::vector<llvm::Instruction *> calls;
   llvm::Value *state_ptr;
 };
 
@@ -305,13 +306,20 @@ void ForwardAliasVisitor::AddInstruction(llvm::Instruction *inst) {
   if (auto store_inst = llvm::dyn_cast<llvm::StoreInst>(inst)) {
     llvm::AAMDNodes aamd;
     store_inst->setAAMetadata(aamd);
+    curr_wl.push_back(inst);
 
   } else if (auto load_inst = llvm::dyn_cast<llvm::LoadInst>(inst)) {
     llvm::AAMDNodes aamd;
     load_inst->setAAMetadata(aamd);
-  }
+    curr_wl.push_back(inst);
 
-  curr_wl.push_back(inst);
+  } else if (llvm::isa<llvm::CallInst>(inst) ||
+             llvm::isa<llvm::InvokeInst>(inst)) {
+
+    calls.push_back(inst);
+  } else {
+    curr_wl.push_back(inst);
+  }
 }
 
 // Iterate through the current worklist, updating the `state_offset` and
@@ -323,6 +331,7 @@ void ForwardAliasVisitor::AddInstruction(llvm::Instruction *inst) {
 bool ForwardAliasVisitor::Analyze(llvm::Function *func) {
   curr_wl.clear();
   exclude.clear();
+  calls.clear();
   state_access_offset.clear();
   state_offset.clear();
 
@@ -378,6 +387,10 @@ bool ForwardAliasVisitor::Analyze(llvm::Function *func) {
     } else if (!pending_wl.empty()) {
       bump = true;
     }
+  }
+
+  for (auto inst : calls) {
+    visit(inst);
   }
 
   // TODO(tim): This condition is triggered a lot.
@@ -688,6 +701,7 @@ VisitResult ForwardAliasVisitor::visitPHINode(llvm::PHINode &I) {
 }
 
 VisitResult ForwardAliasVisitor::visitCallInst(llvm::CallInst &inst) {
+
   // If we have not seen this instruction before, add it.
   if (!live_args.count(&inst)) {
     auto args = inst.arg_operands();
@@ -698,12 +712,14 @@ VisitResult ForwardAliasVisitor::visitCallInst(llvm::CallInst &inst) {
 }
 
 VisitResult ForwardAliasVisitor::visitInvokeInst(llvm::InvokeInst &inst) {
+
   // If we have not seen this instruction before, add it.
   if (!live_args.count(&inst)) {
     auto args = inst.arg_operands();
     auto live = GetLiveSetFromArgs(args, state_offset, state_slots);
     live_args.emplace(&inst, live);
   }
+
   return VisitResult::Ignored;
 }
 
@@ -900,10 +916,18 @@ bool LiveSetBlockVisitor::VisitBlock(llvm::BasicBlock *block) {
         args = llvm::dyn_cast<llvm::InvokeInst>(inst)->arg_operands();
       }
 
-      auto arg_live = live_args.at(inst);
-      live |= arg_live;
-      // Since `GetLiveSetFromArgs` only sets bits, this is safe to do.
-      func_used |= arg_live;
+      auto arg_live_it = live_args.find(inst);
+      if (arg_live_it == live_args.end()) {
+        live.set();
+      } else {
+        live |= arg_live_it->second;
+
+        // Masking in all the bits is heavier than the branch or just flipping
+        // one bit.
+        if (!FLAGS_dot_output_dir.empty()) {
+          func_used |= arg_live_it->second;
+        }
+      }
 
     } else if (auto store_inst = llvm::dyn_cast<llvm::StoreInst>(inst)) {
       auto scope = GetScopeFromInst(*inst);
@@ -1141,10 +1165,6 @@ class ForwardingBlockVisitor {
     const ScopeToOffset &scope_to_offset;
     const std::vector<StateSlot> &state_slots;
     const llvm::FunctionType *lifted_func_ty;
-    int truncated_loads;
-    int truncated_stores;
-    int replaced_loads;
-    int replaced_stores;
 
     ForwardingBlockVisitor(
       llvm::Function &func_,
@@ -1176,17 +1196,10 @@ ForwardingBlockVisitor::ForwardingBlockVisitor(
 
 void ForwardingBlockVisitor::Visit(const ValueToOffset &val_to_offset,
                                    KillCounter &stats) {
-  // if any visit makes progress, continue the loop
-  truncated_loads = 0;
-  truncated_stores = 0;
-  replaced_loads = 0;
-  replaced_stores = 0;
+  // If any visit makes progress, continue the loop.
   for (auto &block : func) {
     VisitBlock(&block, val_to_offset, stats);
   }
-  stats.fwd_loads = replaced_loads;
-  stats.fwd_stores = replaced_stores;
-  stats.fwd_truncated = truncated_loads + truncated_stores;
 }
 
 void ForwardingBlockVisitor::VisitBlock(llvm::BasicBlock *block,
@@ -1424,7 +1437,7 @@ void RemoveDeadStores(llvm::Module *module,
   const AAMDInfo aamd_info(slots, module->getContext());
   const llvm::DataLayout dl = module->getDataLayout();
 
-  InstToLiveSet live_args = {};
+  InstToLiveSet live_args;
 
   for (auto &func : *module) {
     if (&func == bb_func ||
@@ -1437,44 +1450,41 @@ void RemoveDeadStores(llvm::Module *module,
 
     // If the analysis succeeds for this function, add the AAMDNodes.
     if (fav.Analyze(&func)) {
-      DLOG(INFO) << "Offsets: " << fav.state_offset.size();
-      DLOG(INFO) << "Aliases: " << fav.state_access_offset.size();
-      DLOG(INFO) << "Excluded: " << fav.exclude.size();
       AddAAMDNodes(fav.state_access_offset, aamd_info.slot_aamds);
     }
 
     // Perform load and store forwarding.
     if (!FLAGS_disable_register_forwarding) {
       ForwardingBlockVisitor fbv(func, fav.state_access_offset,
-        aamd_info.slot_scopes, slots, &dl);
+                                 aamd_info.slot_scopes, slots, &dl);
       fbv.Visit(fav.state_offset, stats);
     }
   }
 
   // Perform live set analysis
   LiveSetBlockVisitor visitor(*module, live_args, aamd_info.slot_scopes,
-    slots, bb_func->getFunctionType(), &dl);
+                              slots, bb_func->getFunctionType(), &dl);
 
   visitor.FindLiveInsts();
   visitor.CollectDeadInsts();
 
-  for (auto &func : *module) {
-    std::stringstream fname;
+  if (!FLAGS_dot_output_dir.empty()) {
+    for (auto &func : *module) {
+      std::stringstream fname;
 
-    // Log useful DOT digraphs to a directory for every successfully
-    // analyzed function.
-    if (!FLAGS_dot_output_dir.empty()) {
+      // Log useful DOT digraphs to a directory for every successfully
+      // analyzed function.
       fname << FLAGS_dot_output_dir << PathSeparator();
-    }
 
-    if (func.getName().empty()) {
-      fname << "func_" << std::hex << reinterpret_cast<uintptr_t>(&func);
-    } else {
-      fname << func.getName().str();
-    }
+      if (func.getName().empty()) {
+        fname << "func_" << std::hex << reinterpret_cast<uintptr_t>(&func);
+      } else {
+        fname << func.getName().str();
+      }
 
-    std::ofstream dot(fname.str());
-    visitor.CreateDOTDigraph(dot);
+      std::ofstream dot(fname.str());
+      visitor.CreateDOTDigraph(dot);
+    }
   }
 
   visitor.DeleteDeadInsts(stats);
--- a/remill/BC/DeadStoreEliminator.cpp
+++ b/remill/BC/DeadStoreEliminator.cpp
@@ -60,7 +60,11 @@ struct KillCounter {
   uint64_t removed_insts;
   uint64_t fwd_loads;
   uint64_t fwd_stores;
+  uint64_t fwd_perfect;
   uint64_t fwd_truncated;
+  uint64_t fwd_casted;
+  uint64_t fwd_reordered;
+  uint64_t fwd_failed;
 };
 
 // Recursive visitor of the `State` structure that assigns slots of ranges of
@@ -1107,7 +1111,8 @@ class ForwardingBlockVisitor {
       const llvm::DataLayout *dl_);
 
     void Visit(const ValueToOffset &val_to_offset, KillCounter &stats);
-    void VisitBlock(llvm::BasicBlock *block, const ValueToOffset &val_to_offset);
+    void VisitBlock(llvm::BasicBlock *block, const ValueToOffset &val_to_offset,
+                    KillCounter &stats);
 
   private:
     const llvm::DataLayout *dl;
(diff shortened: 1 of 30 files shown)


remill/BC/DeadStoreEliminator.cpp
```diff
@@ -0,0 +1,532 @@
...
+    auto ptr = offset_map.find(val);
+    if (ptr == offset_map.end()) {
+      // no pointer found
+      return VisitResult::NoProgress;
+    } else {
+      // get the offset
+      llvm::APInt offset;
+      if (I.accumulateConstantOffset(*dl, offset)) {
+        // use offset (getRawData extracts the uint64_t)
+        offset_map[&I] = *(ptr->second + offset.getRawData());
```
pgoodman: This doesn't make sense. `getRawData` returns a pointer to the data (memory backing an arbitrary sized integer), then you are displacing and dereferencing it, which is equivalent to indexing into an array. `ptr->second` could be pretty big, and who knows where in memory the APInt's internal storage points, so this is really a buffer overflow that reads arbitrary memory. Instead, what you want is this:

```c++
offset_map[&I] = ptr->second + offset.getZExtValue();
```

remill/BC/DeadStoreEliminator.cpp
```diff
@@ -0,0 +1,532 @@
...
+// Visit an add or sub instruction.
+VisitResult ForwardAliasVisitor::visitBinaryOp_(llvm::BinaryOperator &I, bool plus) {
+  auto val1 = I.getOperand(0);
+  auto val2 = I.getOperand(1);
+  if (exclude.count(val1) || exclude.count(val2)) {
+    exclude.insert(&I);
+    LOG(INFO) << "excluding: " << LLVMThingToString(&I);
+    return VisitResult::Progress;
+  } else {
+    if (auto cint = llvm::dyn_cast<llvm::ConstantInt>(val1)) {
```
pgoodman: Rename to `cint1`

remill/BC/DeadStoreEliminator.cpp
```diff
@@ -0,0 +1,532 @@
...
+      auto ptr = offset_map.find(val2);
+      if (ptr == offset_map.end()) {
+        return VisitResult::NoProgress;
+      } else {
+        auto offset = (plus ? ptr->second + cint->getZExtValue() : ptr->second - cint->getZExtValue());
+        offset_map.insert({&I, offset});
+        LOG(INFO) << "offsetting: " << LLVMThingToString(&I);
+        return VisitResult::Progress;
+      }
+    } else if (auto cint = llvm::dyn_cast<llvm::ConstantInt>(val2)) {
```
pgoodman: Rename to `cint2`.

remill/BC/DeadStoreEliminator.cpp
```diff
@@ -0,0 +1,532 @@
...
+      auto ptr = offset_map.find(val1);
+      if (ptr == offset_map.end()) {
+        return VisitResult::NoProgress;
+      } else {
+        auto offset = (plus ? ptr->second + cint->getZExtValue() : ptr->second - cint->getZExtValue());
+        offset_map.insert({&I, offset});
```
pgoodman: Add a `CHECK` that `offset` doesn't exceed the maximum possible offset.

remill/BC/DeadStoreEliminator.cpp
```diff
@@ -0,0 +1,532 @@
...
+    exclude.insert(&I);
+    LOG(INFO) << "excluding: " << LLVMThingToString(&I);
+    return VisitResult::Progress;
+  } else {
+    if (auto cint = llvm::dyn_cast<llvm::ConstantInt>(val1)) {
+      auto ptr = offset_map.find(val2);
+      if (ptr == offset_map.end()) {
+        return VisitResult::NoProgress;
+      } else {
+        auto offset = (plus ? ptr->second + cint->getZExtValue() : ptr->second - cint->getZExtValue());
```
pgoodman: Add a `CHECK` that `offset` doesn't exceed the maximum possible offset.

remill/BC/DeadStoreEliminator.cpp
```diff
@@ -0,0 +1,532 @@
...
+      }
+    } else {
+      // check if both are in the offset_map
+      auto ptr1 = offset_map.find(val1);
+      auto ptr2 = offset_map.find(val2);
+      if (ptr1 == offset_map.end() || ptr2 == offset_map.end()) {
+        return VisitResult::NoProgress;
+      } else {
+        auto offset = (plus ? ptr1->second + ptr2->second : ptr1->second - ptr2->second);
+        offset_map.insert({&I, offset});
```
pgoodman: Add a `CHECK` that `offset` doesn't exceed the maximum possible offset.

remill/BC/DeadStoreEliminator.cpp
```diff
@@ -0,0 +1,532 @@
...
+VisitResult ForwardAliasVisitor::visitPHINode(llvm::PHINode &I) {
+  LOG(INFO) << "Entered PHI node";
+  // iterate over each operand
+  auto in_offset_map = false;
+  uint64_t offset;
+  for (auto &operand : I.operands()) {
+    if (exclude.count(operand)) {
+      // fail if some operands are excluded and others are state offsets
+      if (in_offset_map) {
+        return VisitResult::Error;
```
pgoodman: This check isn't sufficient to check the property of interest. I think what you want is to have a `bool` that is something like `bool in_exclude_set = false;` then you set to `true` if you find something in the exclude set, but keep going! (just like with `in_offset_map`). At the end of the function, `if (in_offset_map && in_exclude_set)` then you want to error out.

The key is that every time we visit the phi, we want to inspect all operands, so we can never return early inside this loop.

remill/BC/DeadStoreEliminator.cpp
```diff
@@ -0,0 +1,532 @@
...
+
+// Return the scope of the given instruction.
+llvm::MDNode *GetScopeFromInst(llvm::Instruction &I) {
+  llvm::AAMDNodes N;
+  I.getAAMetadata(N);
+  return N.Scope;
+}
+
+// Remove all dead stores.
+void removeDeadStores(std::unordered_set<llvm::Instruction *> dead_stores) {
```
pgoodman: Uppercase the `R`. Pass `dead_stores` by const reference.

remill/BC/DeadStoreEliminator.cpp
```diff
@@ -0,0 +1,532 @@
...
+    // The machines rose from the ashes of the nuclear fire....
+    if (block.getTerminator()) {
+      curr_wl.push_back(&block);
+      // add block to map with initial entry and exit sets
+      block_map.insert({&block, std::make_pair(LiveSet(), LiveSet())});
+    }
+  }
+}
+
+void LiveSetBlockVisitor::visit() {
```
pgoodman: `void` inside of empty parens. If you have a function that takes no arguments, always put in `void`.

remill/BC/DeadStoreEliminator.cpp
```diff
@@ -0,0 +1,532 @@
...
+    // update curr_wl to the next set of insts
+    curr_wl.swap(next_wl);
+    // reset upcoming insts
+    next_wl.clear();
+  }
+  LOG(INFO) << "Found " << to_remove.size() << " dead stores to remove";
+  removeDeadStores(to_remove);
+}
+
+VisitResult LiveSetBlockVisitor::visitBlock(llvm::BasicBlock *B) {
```
pgoodman: Rename to `VisitBlock`

remill/BC/DeadStoreEliminator.cpp
```diff
@@ -0,0 +1,532 @@
...
+void removeDeadStores(std::unordered_set<llvm::Instruction *> dead_stores) {
+  /* for (auto *store : dead_stores) { */
+  /*   store->eraseFromParent(); */
+  /* } */
+}
+
+LiveSetBlockVisitor::LiveSetBlockVisitor(std::unordered_map<llvm::MDNode *, uint64_t> scope_to_offset_)
+  : scope_to_offset(scope_to_offset_), curr_wl(), next_wl(), block_map(), to_remove(), live() { }
+
+void LiveSetBlockVisitor::addFunction(llvm::Function &func) {
```
pgoodman: Rename to `AddFunction`

remill/BC/DeadStoreEliminator.cpp
```diff
@@ -0,0 +1,532 @@
...
+void ForwardAliasVisitor::addInstructions(std::vector<llvm::Instruction *> &insts) {
+  curr_wl.insert(insts.begin(), insts.end());
+}
+
+// Iterate through the current worklist, updating the offset_map and alias_map
+// according to the instructions in the list. Any instruction that is not
+// currently interpretable (some of its pointers are not yet in the offset_map)
+// is withheld to the next analysis round in the next worklist.
+// Analysis repeats until the current worklist is empty.
+void ForwardAliasVisitor::analyze(void) {
```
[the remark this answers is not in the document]
pgoodman: Rename to `Analyze`.

remill/BC/DeadStoreEliminator.cpp
```diff
@@ -0,0 +1,532 @@
...
+      inst->setAAMetadata(aamd);
+    }
+  }
+}
+
+/* Return a map of MDNode scopes and a vector of AAMDNodes based on the given vector of StateSlots,
+ * where each byte offset (i.e. index) in the slots vector is mapped to a
+ * corresponding AAMDNodes struct.
+ */
+std::pair<std::unordered_map<llvm::MDNode *, uint64_t>, std::vector<llvm::AAMDNodes>> generateAAMDInfo(
```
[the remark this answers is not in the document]
pgoodman: Rename to `GenerateAAMDInfo`.

remill/BC/DeadStoreEliminator.cpp
```diff
@@ -0,0 +1,532 @@
...
+    // update StateVisitor fields
+    idx++;
+    offset += len;
+  }
+}
+
+/* For each instruction in the alias map, add an AAMDNodes struct 
+ * which specifies the aliasing stores and loads to the instruction's byte offset.
+ */
+void addAAMDNodes(AliasMap alias_map, std::vector<llvm::AAMDNodes> aamds) {
```
pgoodman: Rename to `AddAAMDNodes`

remill/BC/DeadStoreEliminator.cpp
```diff
@@ -0,0 +1,532 @@
...
+
+StateVisitor::StateVisitor(llvm::DataLayout *dl_)
+  : slots(std::vector<StateSlot>()), idx(0), offset(0), dl(dl_) { }
+
+/* Update the StateVisitor's slots field to hold
+ * a StateSlot for every byte offset into the state.
+ * The StateSlot element is the same across each byte offset
+ * that is within the element's begin offset and end offset.
+ */
+void StateVisitor::visit(llvm::Type *ty) {
```
pgoodman: Rename to `Visit`

remill/BC/DeadStoreEliminator.cpp
```diff
@@ -0,0 +1,532 @@
...
+  return vis.slots;
+}
+
+StateSlot::StateSlot(uint64_t i_, uint64_t offset_, uint64_t size_)
+  : i(i_), offset(offset_), size(size_) { }
+
+StateVisitor::StateVisitor(llvm::DataLayout *dl_)
+  : slots(std::vector<StateSlot>()), idx(0), offset(0), dl(dl_) { }
+
+/* Update the StateVisitor's slots field to hold
```
pgoodman: Use `//` for comments, even block comments.

remill/BC/DeadStoreEliminator.cpp
```diff
@@ -0,0 +1,532 @@
...
+      auto aamd = aamds[alias.second];
+      inst->setAAMetadata(aamd);
+    } else if (auto inst = llvm::dyn_cast<llvm::StoreInst>(alias.first)) {
+      auto aamd = aamds[alias.second];
+      inst->setAAMetadata(aamd);
+    }
+  }
+}
+
+/* Return a map of MDNode scopes and a vector of AAMDNodes based on the given vector of StateSlots,
```
pgoodman: Use `//` for comments.

remill/BC/DeadStoreEliminator.cpp
```diff
@@ -0,0 +1,532 @@
...
+
+// Remove all dead stores.
+void removeDeadStores(std::unordered_set<llvm::Instruction *> dead_stores) {
+  /* for (auto *store : dead_stores) { */
+  /*   store->eraseFromParent(); */
+  /* } */
+}
+
+LiveSetBlockVisitor::LiveSetBlockVisitor(std::unordered_map<llvm::MDNode *, uint64_t> scope_to_offset_)
+  : scope_to_offset(scope_to_offset_), curr_wl(), next_wl(), block_map(), to_remove(), live() { }
```
pgoodman: Formatting. Take `scope_to_offset_` by const reference.

remill/BC/DeadStoreEliminator.cpp
pgoodman: Rename to `Visit`

remill/BC/DeadStoreEliminator.cpp
```diff
@@ -0,0 +1,532 @@
...
+    }
+  }
+  return VisitResult::Progress;
+}
+
+void GenerateLiveSet(llvm::Module *module, std::unordered_map<llvm::MDNode *, uint64_t> &scopes) {
+  auto bb_func = BasicBlockFunction(module);
+  //size_t slots = scopes.size();
+  for (auto &func : *module) {
+    if (&func != bb_func
```
pgoodman: Restructure `if` into format previously described.

remill/BC/DeadStoreEliminator.h
```diff
@@ -0,0 +1,137 @@
...
+
+typedef std::bitset<4096> LiveSet;
+
+llvm::MDNode *GetScopeFromInst(llvm::Instruction &I);
+
+void removeDeadStores(std::unordered_set<llvm::Instruction *> dead_stores);
+
+class LiveSetBlockVisitor {
+  public:
+    std::unordered_map<llvm::MDNode *, uint64_t> scope_to_offset;
```
pgoodman: make this into a const reference.

remill/BC/DeadStoreEliminator.h
```diff
@@ -0,0 +1,137 @@
...
+    virtual VisitResult visitAdd(llvm::BinaryOperator &I);
+    virtual VisitResult visitSub(llvm::BinaryOperator &I);
+    virtual VisitResult visitPHINode(llvm::PHINode &I);
+
+  private:
+    const llvm::DataLayout *dl;
+    virtual VisitResult visitBinaryOp_(llvm::BinaryOperator &I, bool plus);
+};
+
+typedef std::bitset<4096> LiveSet;
```
pgoodman: Add a `CHECK` somewhere that the number of slots does not exceed the number of bits in a `LiveSet`.

scripts/build.sh
```diff
@@ -118,7 +119,8 @@ function GetOSVersion
 
     *)
       printf "[x] ${distribution_name} is not yet a supported distribution.\n"
-      return 1
+      # FIXME: remove workaround
```
pgoodman: Maybe just add in a case for your operating system that makes `OS_VERSION` into `ubuntu1604` and print out a warning? Then this will be an officially supported workaround.

remill/BC/DeadStoreEliminator.cpp
```diff
@@ -0,0 +1,532 @@
...
+VisitResult LiveSetBlockVisitor::visitBlock(llvm::BasicBlock *B) {
+  for (llvm::BasicBlock *succ : successors(B)) {
+    // all successors must be default have been visited first
+    // b.exit = U { s.entry | for s in successors(b)}
+    block_map[B].second |= block_map[succ].first;
+  }
+  for (auto inst = B->rbegin(); inst != B->rend(); ++inst) {
+    if (inst != B->rbegin() && llvm::isa<llvm::TerminatorInst>(&*inst)) {
+      // mark as all live
+      block_map[B].first = LiveSet(4095);
```
pgoodman: I am pretty sure that this doesn't do what you think it does. Also, maps/unordered maps default initialize their things. So `.first` is already a "valid" bitset, with all bits cleared. If you want to set all bits to `1`, then what you want is `block_map[B].first.set();`.

remill/BC/DeadStoreEliminator.cpp
```diff
@@ -0,0 +1,532 @@
...
+}
+
+VisitResult LiveSetBlockVisitor::visitBlock(llvm::BasicBlock *B) {
+  for (llvm::BasicBlock *succ : successors(B)) {
+    // all successors must be default have been visited first
+    // b.exit = U { s.entry | for s in successors(b)}
+    block_map[B].second |= block_map[succ].first;
+  }
+  for (auto inst = B->rbegin(); inst != B->rend(); ++inst) {
+    if (inst != B->rbegin() && llvm::isa<llvm::TerminatorInst>(&*inst)) {
```
pgoodman: A `BranchInst` is a `TerminatorInst`.

scripts/build.sh
alberdingk-thijm: I'm running Antergos Linux; weirdly enough, `cat /etc/issue` returns "Antergos Linux \r (\l)", which does weird things to any kind of string manipulation: `"${distribution_name}"` becomes "(\l)gos Linux". Maybe I can see about some other kind of check we can add.

alberdingk-thijm: Perhaps `/etc/os-release/` would be a better file to check?

remill/BC/DeadStoreEliminator.cpp
```diff
@@ -0,0 +1,549 @@
...
+void LiveSetBlockVisitor::Visit(void) {
+  //bool progress = true;
+  auto currlive = new LiveSet;
+  // if any visit makes progress, continue the loop
+  while (!curr_wl.empty()) {
+    LOG(INFO) << "LSBV: " << curr_wl.size() << " blocks in worklist";
+    //progress = false;
+    for (auto block : curr_wl) {
+      auto newlive = VisitBlock(block, currlive);
+      if (*newlive == *currlive) {
```
pgoodman: This is premature. What you are saying here is: "if any block didn't make progress, then the whole system has converged", yet that is too simplistic. It's more that, if the new and current live sets *don't* match, then we want to add the predecessors of a block to the worklist.

remill/BC/DeadStoreEliminator.cpp
```diff
@@ -0,0 +1,549 @@
...
+    curr_wl.swap(next_wl);
+    // reset upcoming insts
+    next_wl.clear();
+  }
+end:
+  LOG(INFO) << "Found " << to_remove.size() << " dead stores to remove";
+  RemoveDeadStores(to_remove);
+}
+
+LiveSet *LiveSetBlockVisitor::VisitBlock(llvm::BasicBlock *B, LiveSet *init) {
```
pgoodman: This function can just return a `bool` about whether or not progress was made.

remill/BC/DeadStoreEliminator.cpp
```diff
@@ -0,0 +1,549 @@
...
+  // Set the exit value based on the initial liveset unioned with the successors
```
pgoodman: I suggest you start with something like this:
```c++
LiveSet live;
for (auto inst_it = B->rbegin(); inst_it != B->rend(); ++inst_it) {
  auto inst = &*inst_it;
  if (llvm::isa<llvm::ReturnInst>(inst) || llvm::isa<llvm::UnreachableInst>(inst)) {
    // mark all live
  } else if (auto br_inst = llvm::dyn_cast<llvm::BrInst>(inst)) {
    // update live set from successors
  } else if (auto call_inst = llvm::dyn_cast<llvm::CallInst>(inst)) {
    // mark all live if the called function type is the same as a lifted function type
  } else if (auto invoke_inst = llvm::dyn_cast<llvm::InvokeInst>(inst)) {
    // treated like a call
  } else if (llvm::isa<llvm::LoadInst>(inst)) {
    ...
  } else if (llvm::isa<llvm::StoreInst>(inst)) {
    ...
  }
}
```

The loop updates `live`. At the end of the function, you have something like:
```c++
auto &old_live_on_entry = block_map[B];
if (old_live_on_entry != live) {
  old_live_on_entry = live;
  return true;
} else {
  return false;
}
```

remill/BC/DeadStoreEliminator.cpp
```diff
@@ -0,0 +1,562 @@
...
+    llvm::MDNode *noalias = llvm::MDNode::get(context, llvm::MDTuple::get(context, noalias_vec));
+    aamds.emplace_back(nullptr, node_sizes[i].first, noalias);
+  }
+  std::unordered_map<llvm::MDNode *, uint64_t> scopes(node_sizes.begin(), node_sizes.end());
+  slot_scopes = scopes;
+  slot_aamds = aamds;
+}
+
+ScopeMap AnalyzeAliases(llvm::Module *module, std::vector<StateSlot> slots) {
+  auto aamd_info = new AAMDInfo(slots, module->getContext());
```
pgoodman: You can just stack-allocate this, `AAMDInfo aamd_info(...);`.

remill/BC/DeadStoreEliminator.cpp
```diff
@@ -0,0 +1,562 @@
...
+  }
+}
+
+// Visit the basic blocks in the worklist and update the block_map.
+void LiveSetBlockVisitor::Visit(void) {
+  bool progress = true;
+  // if any visit makes progress, continue the loop
+  while (!curr_wl.empty() && progress) {
+    LOG(INFO) << "LSBV: " << curr_wl.size() << " blocks in worklist";
+    progress = false;
```
pgoodman: I don't think you need a `progress` tracker or two separate work lists. You should be able to just have a single vector as a worklist, and alway add on to the end of it.

remill/BC/DeadStoreEliminator.cpp
```diff
@@ -0,0 +1,562 @@
...
+
+bool LiveSetBlockVisitor::VisitBlock(llvm::BasicBlock *B) {
+  //block_map[B] = *init;
+  LiveSet live;
+  for (auto inst_it = B->rbegin(); inst_it != B->rend(); ++inst_it) {
+    auto inst = &*inst_it;
+    if (llvm::isa<llvm::ReturnInst>(inst) || llvm::isa<llvm::UnreachableInst>(inst)) {
+      live.set();
+    } else if (llvm::isa<llvm::BranchInst>(inst)) {
+      // update live from successors
```
pgoodman: Oh we also want to handle `llvm::SwitchInst` and other types of terminators: https://llvm.org/docs/LangRef.html#terminator-instructions

remill/BC/DeadStoreEliminator.cpp
```diff
@@ -0,0 +1,562 @@
...
+        // not the callee might not use any of our slots
+        live.set();
+      }
+    } else if (llvm::isa<llvm::StoreInst>(inst)) {
+      if (auto scope = GetScopeFromInst(*inst)) {
+        if (!live.test(scope_to_offset.at(scope))) {
+          // Case 1: slot is marked dead by a previous store and we hit a store.
+          // (i.e. testing the slot at this offset for liveness returns false)
+          // This is a dead store, so we can add it to the chopping block.
+          to_remove.insert(inst);
```
pgoodman: This is too soon. We can't mark anything for removal until the algorithm converges. Perhaps you can have a boolean parameter that, when `true`, adds stuff to the `to_remove` set. That way, we can run the data flow algorithm with `false` until it converges, then re-run on each block with `true` to collect the final set of instructions to remove.

pgoodman: Also, you need to get at the size of the value being stored, and compare it to the size of the slot.

remill/BC/DeadStoreEliminator.cpp
alberdingk-thijm: Couldn't that cause iterator invalidation to add to a vector worklist while iterating it, [in the case that the capacity changes](http://kera.name/articles/2011/06/iterator-invalidation-rules-c0x/)?
The `progress` tracker was added because the program would loop infinitely without it before, so I'll remove it.

remill/BC/DeadStoreEliminator.cpp
```diff
@@ -0,0 +1,768 @@
...
+      LOG(INFO) << "Analyzing aliases of func " << func.getName().str();
+      for (auto &block : func) {
+        for (auto &inst : block) {
+          fav.AddInstruction(&inst);
+        }
+      }
+      // if the analysis succeeds for this function, add the AAMDNodes
+      if (fav.Analyze()) {
+        LOG(INFO) << "Offsets: " << fav.offset_map.size();
+        LOG(INFO) << "Aliases: " << fav.alias_map.size();
```
pgoodman: Let's rename `offset_map` and `alias_map` to something like `state_offset` and `state_access_offset`, respectively.

remill/BC/DeadStoreEliminator.cpp
```diff
@@ -0,0 +1,768 @@
...
+  return N.Scope;
+}
+
+LiveSetBlockVisitor::LiveSetBlockVisitor(llvm::Function &func_,
+    const std::vector<StateSlot> &state_slots_,
+    const ScopeMap &scope_to_offset_,
+    const ValueToOffset &val_to_offset_,
+    const llvm::FunctionType *lifted_func_ty_,
+    const llvm::DataLayout *dl_)
+  : func(func_),
```
pgoodman: Indent as above.

remill/BC/DeadStoreEliminator.cpp
```diff
@@ -0,0 +1,768 @@
...
+    state_slots(state_slots_),
+    curr_wl(),
+    next_wl(),
+    block_map(),
+    to_remove(),
+    lifted_func_ty(lifted_func_ty_),
+    func_used(),
+    on_remove_pass(false),
+    dl(dl_) {
+      for (auto &block_it : func) {
```
pgoodman: `auto &block : func`

remill/BC/DeadStoreEliminator.cpp
```diff
@@ -0,0 +1,768 @@
...
+        auto block = &block_it;
+        auto succ_block_it = successors(block);
```
pgoodman: Just pass in `&block`

remill/BC/DeadStoreEliminator.cpp
```diff
@@ -0,0 +1,768 @@
...
+        if (succ_block_it.begin() == succ_block_it.end()) {
+          curr_wl.push_back(block);
```
pgoodman: `&block`

remill/BC/DeadStoreEliminator.cpp
```diff
@@ -0,0 +1,768 @@
...
+        // Add predecessors to next rotation
+        for (llvm::BasicBlock *pred : predecessors(block)) {
+          next_wl.push_back(pred);
+        }
+      }
+    }
+    // update curr_wl to the next set of insts
+    curr_wl.swap(next_wl);
+    // reset upcoming insts
+    next_wl.clear();
```
pgoodman: Make `next_wl` a local variable of this function, rather than a field of the class. It's not needed anywhere else.

remill/BC/DeadStoreEliminator.cpp
```diff
@@ -0,0 +1,768 @@
...
+  }
+}
+
+// Return true if the call accesses the state struct.
+// If a single slot is accessed, set the slotn to that slot.
+bool LiveSetBlockVisitor::CallAccessesState(llvm::CallInst *call, uint64_t *slotn) {
+  slotn = nullptr;
```
pgoodman: Remove this, it will cause a `SIGSEGV` if you ever get to `*slotn = ...`

remill/BC/DeadStoreEliminator.cpp
```diff
@@ -0,0 +1,768 @@
...
+  for (auto &arg_it : call->arg_operands()) {
+    auto arg = arg_it->stripPointerCasts();
+    if (val_to_offset.count(arg)) {
+      auto offset = val_to_offset.at(arg);
+      if (offset != 0) {
+        // if we access a single non-zero offset, mark just that offset
+        *slotn = state_slots[offset].i;
+      }
+      return true;
```
pgoodman: This is premature; what if the first argument was for slot 1, and the second for slot 0? 

What we really want is so pass in a reference to the live set. Revive any slots that are seen, but if we see slot offset 0, revive everything. Then the function doesn't need to return `true` or `false` because it will have done the work.

remill/BC/DeadStoreEliminator.cpp
```diff
@@ -0,0 +1,768 @@
...
+    } else if (auto call_inst = llvm::dyn_cast<llvm::CallInst>(inst)) {
+      uint64_t slotn = 0;
+      // Alternative 1:
+      // mark as all live if called function has the same type as lifted function
+      // we're not able to see inside other paths, so we can't predict whether or
+      // not the callee might not use any of our slots
+      // Alternative 2:
+      // mark live based on how much of the state struct could be accessed
+      if ((call_inst->getFunctionType() == lifted_func_ty) || 
+          CallAccessesState(call_inst, &slotn)) {
```
pgoodman: Lets rename this function to `MarkLiveArgs` or something like that, and now the type of the called function doesn't matter, we *always* do this, and the function becomes sufficient.

remill/BC/DeadStoreEliminator.cpp
```diff
@@ -0,0 +1,768 @@
...
+      live.set();
+    } else if (llvm::isa<llvm::BranchInst>(inst) || llvm::isa<llvm::SwitchInst>(inst)) {
+      // update live from successors
+      for (llvm::BasicBlock *succ : successors(B)) {
+        // all successors must be default have been visited first
+        // b = U { s | for s in successors(b)}
+        live |= block_map[succ];
+      }
+    } else if (auto call_inst = llvm::dyn_cast<llvm::CallInst>(inst)) {
+      uint64_t slotn = 0;
```
pgoodman: No longer needed.

remill/BC/DeadStoreEliminator.cpp
```diff
@@ -0,0 +1,768 @@
...
+      // Alternative 1:
+      // mark as all live if called function has the same type as lifted function
+      // we're not able to see inside other paths, so we can't predict whether or
+      // not the callee might not use any of our slots
+      // Alternative 2:
+      // mark live based on how much of the state struct could be accessed
+      if ((call_inst->getFunctionType() == lifted_func_ty) || 
+          CallAccessesState(call_inst, &slotn)) {
+        if (!slotn) {
```
pgoodman: No longer needed.

remill/BC/DeadStoreEliminator.cpp
```diff
@@ -0,0 +1,768 @@
...
+          live.set();
+        } else {
+          live.set(slotn);
+        }
+      }
+    } else if (llvm::isa<llvm::InvokeInst>(inst)) {
```
pgoodman: Fold this into handling of `CallInst`; I think we can do both the same way :-)

remill/BC/DeadStoreEliminator.cpp
```diff
@@ -0,0 +1,768 @@
...
+          // Full stores kill the slot.
+          live.reset(state_slot.i);
+        }
+        // Mark that we have accessed this slot
+        func_used.set(state_slot.i);
+      }
+    } else if (llvm::isa<llvm::LoadInst>(inst)) {
+      if (auto scope = GetScopeFromInst(*inst)) {
+        // Loads revive the slot.
+        live.set(state_slots[scope_to_offset.at(scope)].i);
```
pgoodman: Pull out the `state_slots[scope_to_offset.at(scope)].i` into a local var so that you don't need to do it twice.

remill/BC/DeadStoreEliminator.cpp
```diff
@@ -0,0 +1,768 @@
...
+        // Mark that we have accessed this slot
+        func_used.set(state_slots[scope_to_offset.at(scope)].i);
+      }
+    }
+  }
+  if (block_map.count(B)) {
```
pgoodman: We don't need this `if` or the `else` branch, just everything inside the `if`.

remill/BC/DeadStoreEliminator.cpp
```diff
@@ -0,0 +1,768 @@
...
+        //LOG(INFO) << LLVMThingToString(inst);
+      //}
+    //}
+    store->eraseFromParent();
+  }
+  on_remove_pass = false;
+}
+
+// Generate a DOT digraph file representing the dataflow of the LSBV.
+void LiveSetBlockVisitor::CreateDOTDigraph() {
```
pgoodman: Make all functions that take zero arguments take `(void)`

remill/BC/DeadStoreEliminator.cpp
```diff
@@ -0,0 +1,768 @@
...
+          dot << "<td align=\"left\" bgcolor=\"red\">" << LLVMThingToString(&inst);
+        } else {
+          dot << "<td align=\"left\">" << LLVMThingToString(&inst);
+        }
+        inst.setAAMetadata(original);
+      }
+      dot << "</td></tr>\n";
+    }
+    // last row: exit liveset
+    LiveSet exit_live = blive;
```
pgoodman: No `= blive`, `blive` represents the live set of the block on entry, so we can't just assign it to the exit live set.

remill/BC/DeadStoreEliminator.h
```diff
@@ -0,0 +1,164 @@
...
+//template<typename SubClass> class InstVisitor;
+}  // namespace llvm
+
+namespace remill {
+
+class StateSlot {
+public:
+  StateSlot(uint64_t i, uint64_t offset, uint64_t size);
+  // slot index
+  uint64_t i;
```
pgoodman: Rename to `index`.

remill/BC/DeadStoreEliminator.h
```diff
@@ -0,0 +1,164 @@
...
+  // Inclusive beginning byte offset
+  uint64_t offset;
+  // Size of the slot
```
pgoodman: `// Size of the slot in bytes`

remill/BC/DeadStoreEliminator.h
```diff
@@ -0,0 +1,164 @@
...
+class GetElementPtrInst;
+class GetPtrToIntInst;
+class GetIntToPtrInst;
+class BitCastInst;
+//template<typename SubClass> class InstVisitor;
+}  // namespace llvm
+
+namespace remill {
+
+class StateSlot {
```
pgoodman: Add a comment describing what a "slot" actually is.

remill/BC/DeadStoreEliminator.h
```diff
@@ -0,0 +1,164 @@
...
+  uint64_t offset;
+  // Size of the slot
+  uint64_t size;
+};
+
+class StateVisitor {
+  public:
+    std::vector<StateSlot> slots;
+    // the current index in the state structure
+    uint64_t idx;
```
pgoodman: Rename to `index`.

remill/BC/DeadStoreEliminator.h
```diff
@@ -0,0 +1,164 @@
...
+  StateSlot(uint64_t i, uint64_t offset, uint64_t size);
+  // slot index
+  uint64_t i;
+  // Inclusive beginning byte offset
+  uint64_t offset;
+  // Size of the slot
+  uint64_t size;
+};
+
+class StateVisitor {
```
pgoodman: Add comment describing purpose of this class.

remill/BC/DeadStoreEliminator.h
```diff
@@ -0,0 +1,164 @@
...
+  public:
+    std::vector<StateSlot> slots;
+    // the current index in the state structure
+    uint64_t idx;
+    // the current offset in the state structure
+    uint64_t offset;
+
+    StateVisitor(llvm::DataLayout *dl_);
```
pgoodman: Add `explicit` keyword before `StateVisitor(...)`. Always mark single argument constructors as such.

remill/BC/DeadStoreEliminator.h
```diff
@@ -0,0 +1,164 @@
...
+
+typedef std::unordered_map<llvm::Value *, uint64_t> ValueToOffset;
+typedef std::unordered_map<llvm::Instruction *, uint64_t> InstToOffset;
+typedef std::unordered_map<llvm::MDNode *, uint64_t> ScopeMap;
+
+enum class VisitResult;
+
+enum class OpType;
+
+bool GetUnsignedOffset(int64_t v1, int64_t v2, OpType op, int64_t max, uint64_t *result);
```
pgoodman: We don't need this or `OpType` in the header file because they won't be used outside of your file, so we can make this function `static`.

remill/BC/DeadStoreEliminator.cpp
alberdingk-thijm: This is my favourite kind of refactoring.

remill/BC/DeadStoreEliminator.cpp
```diff
@@ -0,0 +1,804 @@
...
+  }
+}
+
+// Visit a select instruction and update the offset map.
+VisitResult ForwardAliasVisitor::visitSelect(llvm::SelectInst &I) {
+  auto in_state_offset = false;
+  auto in_exclude_set = false;
+  uint64_t offset = 0;
+  // check both the true branch and the false branch
+  auto trueval = I.getTrueValue();
```
pgoodman: `true_val`, `false_val`

remill/BC/DeadStoreEliminator.cpp
```diff
@@ -0,0 +1,804 @@
...
+  auto falseval = I.getFalseValue();
+  if (exclude.count(trueval)) {
+    in_exclude_set = true;
+  } else {
+    auto ptr = state_offset.find(trueval);
+    if (ptr == state_offset.end()) {
+      return VisitResult::NoProgress;
+    } else {
+      if (!in_state_offset) {
```
pgoodman: This is never true. I think you should update this function to do:
```
auto true_ptr = state_offset.find(true_val);
auto false_ptr = state_offset.find(false_val);
auto in_exclude_set = exclude.count(true_val) || exclude.count(false_val);
auto in_state_offset = true_ptr != state_offset.end() || false_ptr != state_offset.end();
```

Then handle the cases accordingly.

remill/BC/DeadStoreEliminator.cpp
```diff
@@ -0,0 +1,804 @@
...
+  }
+  on_remove_pass = false;
+}
+
+// Remove all dead stores.
+bool LiveSetBlockVisitor::DeleteDeadInsts(void) {
+  bool changed = false;
+  while (!to_remove.empty()) {
+    auto inst = to_remove.back();
+
```
pgoodman: Move the `to_remove.pop_back()` to here

remill/BC/DeadStoreEliminator.cpp
```diff
@@ -0,0 +1,804 @@
...
+      //}
+    //}
+
+    inst->replaceAllUsesWith(llvm::UndefValue::get(inst->getType()));
+
+    for (auto &operand : inst->operands()) {
+      if (auto op_inst = llvm::dyn_cast<llvm::Instruction>(operand)) {
+        operand = nullptr;
+        if (llvm::isInstructionTriviallyDead(op_inst)) {
+          to_remove.insert(to_remove.begin(), op_inst);
```
pgoodman: Make this insert at the end of the vector rather than at the beginning. Otherwise the whole thing needs to be continually shifted.

remill/BC/DeadStoreEliminator.cpp
```diff
@@ -0,0 +1,804 @@
...
+        }
+      }
+    }
+    inst->eraseFromParent();
+    to_remove.pop_back();
```
pgoodman: Delete from here and move this up.

remill/BC/DeadStoreEliminator.cpp
```diff
@@ -0,0 +1,804 @@
...
+  }
+}
+
+// Visit a store instruction and update the alias map.
+VisitResult ForwardAliasVisitor::visitStoreInst(llvm::StoreInst &I) {
+  // clear any aametadata
+  llvm::AAMDNodes aamd(nullptr, nullptr, nullptr);
+  I.setAAMetadata(aamd);
+  // if the store is using a state pointer, fail immediately
+  if (state_offset.find(I.getOperand(0)) != state_offset.end()) {
```
pgoodman: `state_offset.count(I.getOperand(0))` is slightly shorter.

remill/BC/DeadStoreEliminator.cpp
```diff
@@ -0,0 +1,804 @@
...
+  } else if (exclude.count(val)) {
+    exclude.insert(&I);
+    return VisitResult::Progress;
+  } else {
+    auto ptr = state_offset.find(val);
+    if (ptr == state_offset.end()) {
+      return VisitResult::NoProgress;
+    } else {
+      // loads mean we now have an alias to the pointer
+      exclude.insert(&I);
```
pgoodman: Actually, let's not do this. My bad, it could pessimize our `visitBinaryOp_` analysis!

remill/BC/DeadStoreEliminator.cpp
```diff
@@ -0,0 +1,804 @@
...
+  llvm::AAMDNodes aamd(nullptr, nullptr, nullptr);
+  I.setAAMetadata(aamd);
+  // get the initial ptr
+  auto val = I.getPointerOperand();
+  // special case: loading the state ptr
+  if (I.getType() == state_ptr->getType()) {
+    state_offset.emplace(&I, 0);
+    return VisitResult::Progress;
+  } else if (exclude.count(val)) {
+    exclude.insert(&I);
```
pgoodman: Actually, let's not do this. My bad, it could pessimize our `visitBinaryOp_` analysis!

alberdingk-thijm: Duplicate? Or do you never want to update the `exclude` set from `visitLoadInst(&I)`?

pgoodman: Oh hrmm, actually now that I think of it again, we *do* want to add to exclude set. At first I thought this would screw up the `visitBinaryOp_`, but then I realized there's no way to convert a loaded value, whose value isn't known, into a constantint ;-)

remill/BC/DeadStoreEliminator.cpp
pgoodman: Let's do this!!! Ignore my earlier comment!!!

PR #261, opened 2018-06-12
Fix NoAlias Attributes for older LLVM versions

artemdinaburg:
Properly add noAlias for older LLVMs (tested on 3.8).

Requesting someone do a sanity check on 4.0 and 5.0+


kumarak:
@pgoodman, The changes look good to me. Approving this PR.

PR #262, opened 2018-06-19
Pag dead store

pgoodman:
Some code cleanups / stylistic changes.


remill/BC/DeadStoreEliminator.cpp
```diff
@@ -33,153 +34,314 @@
...
     next_wl.clear();
+
+    if (progress || bump) {
+      bump = false;
+    } else if (!pending_wl.empty()) {
+      bump = true;
+    }
+  }
+
+  // TODO(tim): This condition is triggered a lot.
```
alberdingk-thijm: Interesting. You'll have to run me through this a bit more; I'm not sure I see what kind of code would allow this to occur.

PR #269, opened 2018-08-03
Compile Options for BC Language Targets

artemdinaburg:
Sets target compile options for BC language targets. 

* Enables usage of `-m32` to generate 32-bit code on amd64 machines.

Necessary to close mcsema issue #442 (https://github.com/trailofbits/mcsema/issues/442)


pgoodman:
These test failures are unusual

pgoodman:
@artemdinaburg, any idea about these test failures?

PR #275, opened 2018-08-23
add ISEL for XORPS_XMMxud_XMMxud

tathanhdinh:
I'am trying to add support for some `xorps` instructions on `xmm` registers, e.g.

```
0f 57 c1    xorps xmm0, xmm1
```
Following [the document](https://github.com/trailofbits/remill/blob/master/docs/LIFE_OF_AN_INSTRUCTION.md), I got its string representation as:

```
(AMD64 0 3 (BYTES 0f 57 c1) 
    XORPS_XMMxud_XMMxud 
        (WRITE_OP (REG_128 XMM0)) 
        (READ_OP (REG_128 XMM0)) 
        (READ_OP (REG_128 XMM1)))
```

and its instruction selection has name `ISEL_XORPS_XMMxud_XMMxud`, that leads to this [tiny change](https://github.com/trailofbits/remill/pull/275/commits/34be1a125d1548fe9373db03c6b55bd1e4d80ca2).

What I have still do not understand is the semantics of `xorps` and its operands (seeing at representation given by `xed`): while `xorps` has its operands of _packed single-precision floating point_ type, `XMMxud` means that the type is _packed unsigned dword interger_. This difference may not lead to any wrong lifting result, but why does `xed` do this way?


pgoodman:
May as well also as `VXORPS`.

tathanhdinh:
I've updated the commit to add `andps` instruction

pgoodman:
As to why XED does things the way it does, I suppose you could file an issue on https://github.com/intelxed/xed :-P

PR #280, opened 2018-08-29
[WIP] Add more diagnosis info to Run.cpp. Looks to be the x87 control, stat…

pgoodman:
…us, and tag word that are off after the signal handler catches the division by zero.


pgoodman:
I merged this so we can get better info about differences from travis logs, though the origin issue of div/idiv on macOS.

Issue #277, opened 2018-08-28
x86: bswap <16bitReg> is missing


pgarba:
Hi,

Support for the opcode bswap for 16bit regs is missing lile:

66 0f c8 bswap ax

tathanhdinh:
IMHO, adding this instruction would be not difficult, but should we? Since the semantics of `bswap ax` is [undefined](https://www.felixcloutier.com/x86/BSWAP.html), and currently remill lifts it into [this function](https://github.com/trailofbits/remill/blob/master/remill/Arch/X86/Runtime/Instructions.cpp#L135).

pgarba:
In reality it just clears the register.  

ax = 0

This instruction will be used in some obfuscated code

tathanhdinh:
Thanks a lot @pgarba, I can make a PR if this is the behaviour of this instruction (@pgoodman what do you think?)

pgoodman:
Sure this is interesting. Definitely add test cases for it :-) I am also curious if XED supports it.

tathanhdinh:
@pgoodman  XED can disassemble the instruction, I got its string representation:

```
(AMD64 0 3 (BYTES 66 0f c8) 
    BSWAP_GPRv_16 
        (WRITE_OP (REG_16 AX)) 
        (READ_OP (REG_16 AX)))
```

I will try to add it if you don't mind.

pgoodman:
Absolutely, go for it! :-D

Here's where to put the test case: https://github.com/trailofbits/remill/blob/master/tests/X86/BITBYTE/BSWAP.S

I don't know if GAS can assemble `bswap ax`, so you may need to do:

```
TEST_BEGIN(BSWAPr16, 1)
    TEST_INPUTS(...)
    mov rax, ARG1_64
    .byte 0x66, 0x0f, 0xc8
TEST_END
```

pgarba:
Thanks for adding.

Here is the link to the obfusacted file:
https://drive.google.com/open?id=1v74LzJAvz597yBZBIkI8dggnVeHmBJp8

PR #284, opened 2018-08-31
add bswap_r16

tathanhdinh:
This PR tries to support #277.


remill/Arch/X86/Semantics/BITBYTE.cpp
```diff
@@ -253,6 +253,11 @@ DEF_ISEL_MnW_Mn_Rn(BTC_MEMv_GPRv, BTCmem);
 DEF_ISEL_RnW_Rn_Rn(BTC_GPRv_GPRv, BTCreg);
 
 namespace {
+DEF_SEM(BSWAP_16, R16W dst, R16 src) {
+  Write(dst, (int16_t)0);
```
pgoodman: Don't use C-style casts. Perhaps instead even use `__remill_undefined_16()` in place of zero.

tathanhdinh:
Hmmm, tests are failed. I will come back Monday.

tathanhdinh:
I cannot reproduce the failed tests in my machine (maybe it's machine dependent?). Moreover, [unicorn](https://github.com/unicorn-engine/unicorn) seems to really swap the register!!?.

tathanhdinh:
Yes, I saw that also but I still do not understand neither. I've tried to debug but I cannot reproduce this behavior on my machine.

tathanhdinh:
I'am able to reproduce this bug in my machine, all failed tests come from `run-x86-tests`. This should be digged deeper, but the behavior is that: if the test case uses `rax` then the function `BSWAPr16_1_lifted` in `run-x86-tests` is very difference from (much sophisticated than) this one in `run-amd64-tests`:

```
push rbp			      push rbp
mov rbp, rsp			      mov rbp, rsp
mov r9d, [rdi+0x8a8]		      mov rax, [rdi+0x8f8]
mov r8d, [rdi+0x8f8]		      mov [rdi+0x8a8], rax
lea r10d, [r9-0x1]		      add rsi, 0x6
movzx eax, r10b			      mov [rdi+0x9a8], rsi
mov ecx, eax			      mov word ptr [rdi+0x8a8], 0x0
shr ecx, 0x1			      pop rbp
and ecx, 0x55			      jmp __remill_missing_block
sub eax, ecx			      
mov ecx, eax			      
and ecx, 0x33333333		      
shr eax, 0x2			      
and eax, 0x33333333		      
add eax, ecx			      
mov ecx, eax			      
shr ecx, 0x4			      
add ecx, eax			      
and ecx, 0x10f0f0f		      
imul eax, ecx, 0x1010101	      
shr eax, 0x18			      
not al				      
and al, 0x1			      
mov [rdi+0x813], al		      
mov eax, r10d			      
xor eax, r9d			      
mov ecx, eax			      
shr cl, 0x4			      
and cl, 0x1			      
mov [rdi+0x815], cl		      
test r10d, r10d			      
setz byte ptr [rdi+0x817]	      
shr r10d, 0x1f			      
mov [rdi+0x819], r10b		      
shr r9d, 0x1f			      
shr eax, 0x1f			      
add eax, r9d			      
cmp eax, 0x2			      
setz byte ptr [rdi+0x81d]	      
mov [rdi+0x8a8], r8d		      
add esi, 0x6			      
mov [rdi+0x9a8], esi		      
mov word ptr [rdi+0x8a8], 0x0	      
pop rbp				      
jmp __remill_missing_block
```
But if `eax` is used in the test case, then everything should be fine: the test passes (in my machine). I still do not understand why there is such difference, maybe you can help?

pgoodman:
Ah I think I know what's happening. So the `0x66` prefix on the bswap is changing it's operand size in 64-bit from the default, but that prefix's effect in 32-bit has a different effect. Just make the test 64-bit only, i.e. `TEST_BEGIN_64` and `TEST_END_64`.

tathanhdinh:
It passes now (on my machine), hope the same goes for Travis-Ci.

PR #290, opened 2018-09-05
add shufps_xmmps_xmmps

tathanhdinh:
Hello all,

This PR tries to support lifting `shufps` instruction, e.g.

```
0f c6 e1 ab    shufps xmm4, xmm1, 0xab
```

### DIFF
--- a/tests/X86/SSE/SHUFPS.S
+++ b/tests/X86/SSE/SHUFPS.S
@@ -1,0 +1,79 @@
+/*
+ * Copyright (c) 2017 Trail of Bits, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#define SHUFPS_INPUTS_64 \
+    0xF82DAC26, 0x091D41A7, \
+    0x5FD4D3D9, 0x39A5f1D8, \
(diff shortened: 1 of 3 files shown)


tests/X86/SSE/SHUFPS.S
```diff
@@ -0,0 +1,47 @@
...
+TEST_BEGIN_64(SHUFPSv128v128imm8_b1, 2)
+TEST_INPUTS(SHUFPS_INPUTS_64)
+    push ARG1_64
+    push ARG1_64
+    movq xmm0, [rsp]
+    push ARG2_64
+    push ARG2_64
+    movq xmm1, [rsp]
+    shufps xmm0, xmm1, 0xb1
+TEST_END_64
```
pgoodman: Add an extra new line at the end.

pgoodman: Can you also make another test that takes only `TEST_INPUTS(0)` and performs many `shufps` on many different xmm regs, with a bunch of different selectors? The xmm registers are pre-filled to have semi-interesting data, so that will really flesh out more things.

pgoodman: So I actually mean something like

```
TEST_BEGIN_64(SHUFPSv128v128imm8_many, 1)
TEST_INPUTS(0)
    shufps xmm7, xmm6, 0x2f
    shufps xmm6, xmm5, 0x57
    shufps xmm5, xmm4, 0xa6
    shufps xmm4, xmm3, 0xc3
    ...
TEST_END_64
```

There are already a bunch of data in xmm0-xmm7:
https://github.com/trailofbits/remill/blob/master/tests/X86/Tests.S#L168-L179
https://github.com/trailofbits/remill/blob/master/tests/X86/Tests.S#L225-L234

Otherwise these additional tests are good 👍

tathanhdinh: Ah, ok. I didn't know that xmm0-7 are initialized already :)

pgoodman:
@tathanhdinh Can you look into these test failures?

tathanhdinh:
Fortunately (or not?), tests are failed on my machine also then I can debug. All tests are passed now (on my machine).

PR #296, opened 2018-09-10
Pass FINDPACKAGE_LLVM_HINTS to parent scope in FindAndSelectClangCompiler

dureuill:
In `CMakeLists.txt:50`, the `FINDPACKAGE_LLVM_HINTS` variable is used, but this variable is only assigned in the `FindAndSelectClangCompiler` function, with a local scope. 

As a result, in `CMakeLists.txt:50`, an empty value is used for `FINDPACKAGE_LLVM_HINTS`.

This PR assigns  `FINDPACKAGE_LLVM_HINTS` with[ `PARENT_SCOPE`](https://cmake.org/cmake/help/v3.2/command/set.html) visibility in `FindAndSelectClangCompiler()` CMake function,  so that it is visible to the caller.


dureuill:
@pgoodman Ah. I have now signed the CLA :-).

PR #298, opened 2018-09-12
Prefer using LLVM_PREFIX if available to find clang and llvm-link programs

dureuill:
Before this PR, when specifying a `LLVM_PREFIX`, it is not used to determine the `CLANG_PATH` and the `LLVMLINK_PATH`, if these binaries are available system-wide.

After this PR, if `LLVM_PREFIX` has been specified, only this variable is used to determine `CLANG_PATH` and `LLVMLINK_PATH`. If it hasn't been specified, the lookup code is unchanged.

Depending on the precise semantics attached to `LLVM_PREFIX`, this may or may be the wanted behavior. Alternatively, we could first check only looking at the `LLVM_PREFIX` path, and then the other paths if not found.
The method I chose has the property that the user cannot accidentally have a `CLANG_PATH` or `LLVMLINK_PATH` not matching their `LLVM_PREFIX`. 
I chose this method because I believe that when the `LLVM_PREFIX` is specified, it precisely conveys the intent to use it, and so it should completely replace the system paths. 

I can change this PR if this is not the wanted behavior.

### DIFF
--- a/cmake/BCCompiler.cmake
+++ b/cmake/BCCompiler.cmake
@@ -28,20 +28,39 @@ if(DEFINED ENV{LLVM_INSTALL_PREFIX})
   set(LLVM_INSTALL_PREFIX $ENV{LLVM_INSTALL_PREFIX})
 endif()
 
-if("${CMAKE_CXX_COMPILER}" STREQUAL "${CLANG_CXX_EXECUTABLE_NAME}")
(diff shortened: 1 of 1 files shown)


dureuill:
Hello,

Any news on this PR? Is there maybe a next step needed?

Thank you for answering!

dureuill:
Wow that was a quick response! :+1:  Thank you

PR #300, opened 2018-09-13
add sti

tathanhdinh:
Hello all,

A nano PR: you might forget `sti` :)


pgoodman:
I approved it, but we'll probably want some hypercalls for enabling/disabling interrupts, that we can then call here, in `cli`, and also in `popfq`.

https://github.com/trailofbits/remill/blob/8e0a28ebae9460bc71302ab9dc4b8fe4fbbf32a0/remill/Arch/X86/Semantics/POP.cpp#L81

pgoodman:
Maybe like, `kX86EnableInterruptsIfPrivileged` and `kX86DisableInterruptsIfPrivileged`

PR #302, opened 2018-09-18
Trace lifter api

pgoodman:


tathanhdinh:
Hi @pgoodman 
I do not fully understand what you are doing yet, but it seems a part of a DBI tool?

pgoodman:
@tathanhdinh I am migrating code over from a DBI tool into the public repo, and refactoring it into a more workable API for others. Mostly though, the motivation for all this is to present remill's APIs on a poster.

tathanhdinh:
Great!!! I'm just noticed by this work several minutes ago, looking into the committed code, it suggests me that this should something like a DBI.

PR #307, opened 2018-09-21
add float128 type

tathanhdinh:
Hello all,

With the goal of simplifying [movaps](https://github.com/trailofbits/remill/compare/master...tathanhdinh:fp128?expand=1#diff-999092029e9e7e4d0e50a5cffa8a497dL50) instruction, this PR tries to add the type `__float128` into remill.

### DIFF
--- a/remill/Arch/Runtime/Intrinsics.cpp
+++ b/remill/Arch/Runtime/Intrinsics.cpp
@@ -49,10 +49,12 @@ extern "C" void __remill_intrinsics(void) {
   USED(__remill_read_memory_f32);
   USED(__remill_read_memory_f64);
   USED(__remill_read_memory_f80);
+  USED(__remill_read_memory_f128);
 
   USED(__remill_write_memory_f32);
   USED(__remill_write_memory_f64);
   USED(__remill_write_memory_f80);
+  USED(__remill_write_memory_f128);
 
   USED(__remill_barrier_load_load);
   USED(__remill_barrier_load_store);
--- a/remill/Arch/Runtime/Intrinsics.h
+++ b/remill/Arch/Runtime/Intrinsics.h
@@ -60,11 +60,17 @@ extern float64_t __remill_read_memory_f64(Memory *, addr_t);
 [[gnu::used]]
 extern float64_t __remill_read_memory_f80(Memory *, addr_t);
 
+[[gnu::used]]
+extern float128_t __remill_read_memory_f128(Memory *, addr_t);
+
 [[gnu::used, gnu::const]]
 extern Memory *__remill_write_memory_f32(Memory *, addr_t, float32_t);
 
 [[gnu::used, gnu::const]]
 extern Memory *__remill_write_memory_f64(Memory *, addr_t, float64_t);
+
+[[gnu::used]]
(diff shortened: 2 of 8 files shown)


tathanhdinh:
Well, it doesn't compile, that comes from

```
remill/remill/Arch/Runtime/Types.h:64:9: error: unknown type name '__float128'
```

pgoodman:
I understand the intention of this PR, but don't agree with its implementation. `MOVAPS` moves a vector of packed aligned, single-precision floats. In general I would like to maintain that semantic as much as possible. There are definitely cases where we re-use some semantics for "convenience" (e.g. semantic is for 32-bit floats, but we use one for 32-bit ints), and that should change.

I think a better option might be to create several derived intrinsics, implementing them in remill/Arch/Runtime/Intrinsics.cpp in terms of other intrinsics. These derived intrinsics would basically be multi-element vector move operations, like `__remill_read_memory_f32v4` for example. They would be marked as `noinline`, and after you have lifted to bitcode and optimized the bitcode, you can replace all calls with these derived intrinsics with whatever you want, or get the status quo behaviour by inlining them.

In your particular use case, it sounds like what you're really getting at is that you want better vectorization of the bitcode. It's important to understand that this is an anti-goal of Remill's current design, though it is not out of the question. Specifically, Remill tries to produce the simplest possible bitcode. It could choose to represent things like `uint8v18_t` with the C++ `vector_size` attribute, but uses arrays. Similarly, optimizations are done with block and SLP vectorization disabled. You probably want these enabled (this is possible if you look at the new `remill::OptimizeModule` function from the recent trace lifting merge). In your case, it may be worth investigating tweaking some of these choices, e.g. replacing some uses of arrays with vectors, and seeing what happens.

tathanhdinh:
Hello @pgoodman  and thank you for the detailed response. 

The `movaps` is a move of vector of float, so this implementation is indeed a "shortcut". I didn't understand yet the ideas lying behind the design of remill, but what you have said:

> ...Remill tries to produce the simplest possible bitcode.

makes sense to me, I'm agree with that: it would be better to let optimizer to do the work.

Frankly, I've tried some available passes of LLVM but still not be able to make it vectorize the current implementation `movaps` (and similar instructions). So I will investigate to the new `remill::OptimizeModule` to understand how it works.

Again, thanks a lot for your response. I really appreciate your help.

pgoodman:
A key thing is that the semantics bitcode module is created with vectorization disabled, and so there are some LLVM module debug entries that may contain this attribute. They should get removed [1] but it's worth looking at the disassembled IR file to see if any compiler options remain.

The relevant vectorization options are [2]. A key thing is that promotion of loads/stores to vector loads/stores is unlikely to happen until *after* you inline/replace all memory access intrinsics with actual loads/stores, then re-optimize to do vectorization. Even then, it seems likely that LLVM will miss stuff. In that case, I think these derived intrinsics may be ideal, because then you can replace all uses of a called derived vector memory access intrinsic with a vector load/store.

[1] https://github.com/trailofbits/remill/blob/master/remill/Arch/Arch.cpp#L538-L564
[2] https://github.com/trailofbits/remill/blob/master/remill/BC/Optimizer.h#L33-L34

PR #325, opened 2019-01-25
add iretq instruction lifting

tathanhdinh:
This PR adds lifting for `iretq`instruction.

Many thanks for any comment.

### DIFF
--- a/remill/Arch/X86/Semantics/CALL_RET.cpp
+++ b/remill/Arch/X86/Semantics/CALL_RET.cpp
@@ -70,8 +70,8 @@ DEF_ISEL_32or64(RET_NEAR, RET);
 728 IRETD IRETD RET BASE I386 ATTRIBUTES: FIXED_BASE0 NOTSX SCALABLE STACKPOP0
 */
 
-#if ADDRESS_SIZE_BITS == 32
 namespace {
+#if ADDRESS_SIZE_BITS == 32
 DEF_SEM(IRETD) {
   auto new_eip = PopFromStack<uint32_t>(memory, state);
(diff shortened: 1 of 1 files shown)


remill/Arch/X86/Semantics/CALL_RET.cpp
```diff
@@ -91,7 +91,40 @@ DEF_SEM(IRETD) {
...
+
+  // update hidden part (segment descriptor) of cs
+
+  auto new_rsp = PopFromStack<uint64_t>(memory, state);
+  auto new_ss = static_cast<uint16_t>(PopFromStack<uint64_t>(memory, state));
+  Write(REG_RSP, new_rsp);
+  Write(REG_SS.flat, new_ss);
+
+  // update hidden part (segment descriptor) of ss
+
```
pgoodman: Can you file an Issue for this? Also, put this as someting like:

```c++
// TODO(tathanhdinh): Update hidden part (segment descriptor) of ss,
//                    see Issue #NNN.
```

My guess is that you'll want to do the segment update by a `__remill_sync_hyper_call`?

tathanhdinh: @pgoodman thanks you for the comment.

> My guess is that you'll want to do the segment update by a `__remill_sync_hyper_call`?
I think we can make the update be explicit (see #334).

tathanhdinh: > Can you file an Issue for this? Also, put this as someting like:

Sure, PR updated.

Issue #288, opened 2018-09-04
Lifting a long jump


tathanhdinh:
Hello all,

When lifting a jmp far instruction, e.g.

```
ea 5f b9 15 77 33 00        jmp far 0x33:0x7715b95f
```

remill will crash at [instruction decoding](https://github.com/trailofbits/remill/blob/master/remill/Arch/X86/Arch.cpp#L718) because it rejects an operand of of "name" `XED_OPERAND_PTR`.

Such an instruction will modify `CS` and `RIP` so to support this instruction, I think that an approach is to parse the operand to get these values. 

Do you know how to parse these values, i.e. segment register and immediate value, using `xed`? Or is the another approach which is more "remill"?

Many thanks for any response.

pgoodman:
Looks like we need to add support for `XED_OPERAND_PTR` to Remill. Once we add support for decoding that operand kind, then I think the right way to handle a far jump/call/return in semantics is via a `__remill_async_hyper_call`. So the semantics would probably take in a 16-bit segment value, which we would place into `cs`, it would take in a 64-bit target PC, which would get placed into `rip`, and then the hyper call name would be set, with different ones for each of jmp/call/ret. This means that a decoded `XED_OPERAND_PTR` will expand into two Remill instruction `Operand`s.

PR #332, opened 2019-02-26
Supporting Intel's far jumps

tathanhdinh:
Hello all,

The PR tries to handle the issue #288. Now remill can lift far jumps instructions, for example

```Assembly
ea 5f b9 15 77 33 00    jmp far 0x33:0x7715b95f
```

will has string presentation

```Lisp
(X86 0 7 (BYTES ea 5f b9 15 77 33 00) 
   JMP_FAR_PTRp_IMMw_32 
     (READ_OP (PTR_32 7715b95f)) 
     (READ_OP (IMM_16 33)))
```

and is lifted into

```
remill-lift-7.0 --arch x86 --bytes  ea5fb915773300 --ir-out /dev/stdout
...
define dso_local %struct.Memory* @sub_0(%struct.State* noalias dereferenceable(3376), i32, %struct.Memory* noalias) {
  %4 = getelementptr inbounds %struct.State, %struct.State* %0, i32 0, i32 6, i32 33, i32 0, i32 0
  %5 = getelementptr inbounds %struct.State, %struct.State* %0, i32 0, i32 0, i32 0
  store i32 11, i32* %5, align 16
  store i32 1997912415, i32* %4, align 4
  %6 = getelementptr inbounds %struct.State, %struct.State* %0, i32 0, i32 4, i32 11, i32 0
  store i16 51, i16* %6, align 2
  %7 = tail call %struct.Memory* @__remill_async_hyper_call(%struct.State* nonnull dereferenceable(3376) %0, i32 1997912415, %struct.Memory* %2) #3
  %8 = load i32, i32* %4, align 4
  %9 = tail call %struct.Memory* @__remill_jump(%struct.State* nonnull %0, i32 %8, %struct.Memory* %7)
  ret %struct.Memory* %9
}
...
```

Many thanks for any comment.

### DIFF
--- a/remill/Arch/Runtime/HyperCall.h
+++ b/remill/Arch/Runtime/HyperCall.h
@@ -87,6 +87,9 @@ class AsyncHyperCall {
     kX86SysEnter,
     kX86SysExit,
 
+    // Far jumps: CS should be updated.
+    kX86JmpFar,
+
     kAArch64SupervisorCall,
 
     // Invalid instruction.
--- a/remill/Arch/X86/Arch.cpp
+++ b/remill/Arch/X86/Arch.cpp
@@ -97,13 +97,28 @@ static bool IsDirectFunctionCall(const xed_decoded_inst_t *xedd) {
   return XED_ICLASS_CALL_NEAR == iclass && XED_OPERAND_RELBR == op_name;
 }
 
+static bool IsDirectFunctionCallFar(const xed_decoded_inst_t *xedd) {
+  auto xedi = xed_decoded_inst_inst(xedd);
+  auto xedo = xed_inst_operand(xedi, 0);
+  auto op_name = xed_operand_name(xedo);
+  auto iclass = xed_decoded_inst_get_iclass(xedd);
+  return XED_ICLASS_CALL_FAR == iclass && XED_OPERAND_PTR == op_name;
+}
+
 static bool IsIndirectFunctionCall(const xed_decoded_inst_t *xedd) {
   auto xedi = xed_decoded_inst_inst(xedd);
   auto xedo = xed_inst_operand(xedi, 0);
   auto op_name = xed_operand_name(xedo);
   auto iclass = xed_decoded_inst_get_iclass(xedd);
-  return (XED_ICLASS_CALL_NEAR == iclass && XED_OPERAND_RELBR != op_name) ||
-         XED_ICLASS_CALL_FAR == iclass;
+  return XED_ICLASS_CALL_NEAR == iclass && XED_OPERAND_RELBR != op_name;
+}
+
+static bool IsIndirectFunctionCallFar(const xed_decoded_inst_t *xedd) {
+  auto xedi = xed_decoded_inst_inst(xedd);
+  auto xedo = xed_inst_operand(xedi, 0);
+  auto op_name = xed_operand_name(xedo);
+  auto iclass = xed_decoded_inst_get_iclass(xedd);
+  return XED_ICLASS_CALL_FAR == iclass && XED_OPERAND_MEM0 == op_name;
 }
 
 static bool IsDirectJump(const xed_decoded_inst_t *xedd) {
@@ -114,14 +129,29 @@ static bool IsDirectJump(const xed_decoded_inst_t *xedd) {
   return XED_ICLASS_JMP == iclass && XED_OPERAND_RELBR == op_name;
 }
 
+static bool IsDirectJumpFar(const xed_decoded_inst_t *xedd) {
+  auto xedi = xed_decoded_inst_inst(xedd);
+  auto xedo = xed_inst_operand(xedi, 0);
+  auto op_name = xed_operand_name(xedo);
+  auto iclass = xed_decoded_inst_get_iclass(xedd);
+  return XED_ICLASS_JMP_FAR == iclass && XED_OPERAND_PTR == op_name;
+}
+
 static bool IsIndirectJump(const xed_decoded_inst_t *xedd) {
   auto xedi = xed_decoded_inst_inst(xedd);
   auto xedo = xed_inst_operand(xedi, 0);
   auto op_name = xed_operand_name(xedo);
   auto iclass = xed_decoded_inst_get_iclass(xedd);
   return (XED_ICLASS_JMP == iclass && XED_OPERAND_RELBR != op_name) ||
-         XED_ICLASS_JMP_FAR == iclass ||
          XED_ICLASS_XEND == iclass || XED_ICLASS_XABORT == iclass;
+}
+
+static bool IsIndirectJumpFar(const xed_decoded_inst_t *xedd) {
+  auto xedi = xed_decoded_inst_inst(xedd);
+  auto xedo = xed_inst_operand(xedi, 0);
+  auto op_name = xed_operand_name(xedo);
+  auto iclass = xed_decoded_inst_get_iclass(xedd);
+  return XED_ICLASS_JMP_FAR == iclass && XED_OPERAND_MEM0 == op_name;
 }
 
 static bool IsNoOp(const xed_decoded_inst_t *xedd) {
@@ -188,7 +218,12 @@ static Instruction::Category CreateCategory(const xed_decoded_inst_t *xedd) {
   } else if (IsInterruptReturn(xedd)) {
     return Instruction::kCategoryAsyncHyperCall;
 
-  } else if (IsNoOp(xedd)) {
+  } else if (IsDirectJumpFar(xedd) || IsIndirectJumpFar(xedd) ||
+             IsDirectFunctionCallFar(xedd) || IsIndirectFunctionCallFar(xedd)) {
+    return Instruction::kCategoryAsyncHyperCall;
+
+  }
+  else if (IsNoOp(xedd)) {
     return Instruction::kCategoryNoOp;
 
   } else {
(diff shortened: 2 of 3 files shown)


remill/Arch/Instruction.h
```diff
@@ -36,7 +36,8 @@ class Operand {
     kTypeRegister,
     kTypeShiftRegister,
     kTypeImmediate,
-    kTypeAddress
+    kTypeAddress,
+    kTypePointer
```
pgoodman: What is the value-added of a `kTypePointer`? It seems like we can use either a `kTypeImmediate` or `kTypeAddress`. Really, though, introducing a new type means updating McSema :-P

remill/Arch/Runtime/HyperCall.h
```diff
@@ -87,6 +87,9 @@ class AsyncHyperCall {
     kX86SysEnter,
     kX86SysExit,
 
+    // Far jumps.
```
pgoodman: Let's add a comment that the `CS` segment is updated before this hyper call.

A more general todo is to comment all hyper calls on their pre-conditions / what can be assumed.

remill/Arch/Instruction.h
tathanhdinh: It's similar (though not identical) to a unsigned `kTypeImmediate`:  
- `InstructionLifter::LiftPointerOperand` is mostly identical to `InstructionLifter::LiftImmediateOperand`), but
- `Arch.cpp::DecodePointer` is different from `Arch.cpp::DecodeImmediate`

I still don't know whether merging them is better or not.

pgoodman: Well what I mean is that DecodePointer can produce an immediate.

tathanhdinh: Yes, it's possible.
Do you want me to modify the PR?

pgoodman: Yes.

tathanhdinh: I will do this tomorrow because it's late now (in France).

pgoodman: À demain.

tathanhdinh: Merci, à demain.

tathanhdinh: @pgoodman I've changed the PR to remove `kTypePointer` and use `kTypeImmediate` instead.

remill/Arch/Runtime/HyperCall.h
tathanhdinh: Updated.

remill/Arch/X86/Arch.cpp
```diff
@@ -477,35 +477,46 @@ static void DecodeImmediate(Instruction &inst,
...
+  op.action = Operand::kActionRead;
+
+  if (XED_OPERAND_PTR == op_name) {
+    auto ptr_size = xed_decoded_inst_get_branch_displacement_width_bits(xedd);
+    CHECK(ptr_size <= operand_size)
+      << "Pointer size is greater than effective operand size at "
+      << std::hex << inst.pc << ".";
+    op.size = ptr_size;
+
+    val = xed_decoded_inst_get_branch_displacement(xedd);
```
pgoodman: So the branch displacement here is actually a value for `PC`, and doesn't need to be added to `PC`? This function returns an `int32_t`, but `val` is a `uint64_t`. In this instance I like there to be explicit casting (via `static_cast`) so that what happens (sign extension, zero extension) is an obvious and deliberate choice by the programming, rather than implicitly added behaviour by the compiler.

remill/Arch/X86/Arch.cpp
```diff
@@ -477,35 +477,46 @@ static void DecodeImmediate(Instruction &inst,
...
 
-  CHECK(imm_size <= operand_size)
+  Operand op = {};
+  op.type = Operand::kTypeImmediate;
+  op.action = Operand::kActionRead;
+
+  if (XED_OPERAND_PTR == op_name) {
+    auto ptr_size = xed_decoded_inst_get_branch_displacement_width_bits(xedd);
+    CHECK(ptr_size <= operand_size)
+      << "Pointer size is greater than effective operand size at "
```
pgoodman: Four space indent the `<<`, so it's like:
```c++
CHECK(...)
    << "hiiii";
```

remill/Arch/X86/Arch.cpp
pgoodman: If `PC` needs to be added, then I wonder if this should be decoded as a relative branch, so the operand comes out as something like `(ADD (REG PC) IMM)`...

remill/Arch/X86/Arch.cpp
```diff
@@ -477,35 +477,46 @@ static void DecodeImmediate(Instruction &inst,
...
+      << std::hex << inst.pc << ".";
+    op.size = ptr_size;
+
+    val = xed_decoded_inst_get_branch_displacement(xedd);
+  } else {
+    auto imm_size = xed_decoded_inst_get_immediate_width_bits(xedd);
+    CHECK(imm_size <= operand_size)
       << "Immediate size is greater than effective operand size at "
```
pgoodman: Four space indent `<<` after a `CHECK`.

remill/Arch/X86/Semantics/UNCOND_BR.cpp
```diff
@@ -22,6 +22,28 @@ DEF_SEM(JMP, T target_pc) {
...
 
+template <typename T>
+DEF_SEM(JMP_FAR_MEM, T target_seg_pc) {
+  HYPER_CALL = AsyncHyperCall::kX86JmpFar;
+  uint64_t target_fword = UShr(UShl(Read(target_seg_pc), 0xf), 0xf);
+  auto pc = static_cast<uint32_t>(target_fword);
+  auto seg = static_cast<uint16_t>(UShr(target_fword, 32));
+  WriteZExt(REG_PC, pc);
+  Write(REG_CS.flat, seg);
+  return __remill_async_hyper_call(state, REG_PC, memory);
```
pgoodman: Here you want is to just `return memory`. Then use of a `__remill_async_hyper_call` should be done by a higher level, i.e. the code doing the lifting. You'll need to modify the `CreateCategory`, and possibly `IsIndirectJump` and `IsIndirectCall`, maybe adding some new function that checks for the far calls/jumps only, and returns `kCategoryAsyncHyperCall`.

The idea here is that, it's hard to make a function call inside of a semantics function into a terminating tail call after all optimizations are said and done. Semantics should be mostly come in and come out, with the actual redirecting of control flow happing between/outside of them. At that level, you can say "this is an async hyper call, so call `intrinsics.async_hyper_call`".

remill/Arch/X86/Semantics/UNCOND_BR.cpp
```diff
@@ -22,6 +22,28 @@ DEF_SEM(JMP, T target_pc) {
...
+}
+
+template <typename S1, typename S2>
+DEF_SEM(JMP_FAR_PTR, S1 src1, S2 src2) {
+  HYPER_CALL = AsyncHyperCall::kX86JmpFar;
+  auto pc = Read(src1);
+  auto seg = Read(src2);
+  WriteZExt(REG_PC, pc);
+  Write(REG_CS.flat, seg);
+  return __remill_async_hyper_call(state, REG_PC, memory);
```
pgoodman: Ditto here.

remill/Arch/X86/Arch.cpp
tathanhdinh: @pgoodman Thanks you for the feedback.

The `PC` isn't need to be added, in case of a `jmp far` the address [is absolute and given in the operand](https://c9x.me/x86/html/file_module_x86_id_147.html). The following example taken from a real trace:

```Assembly
...
0x7715b940    64 8b 15 30 00 00 00                         mov edx, fs:[0x30]
0x7715b947    8b 92 64 04 00 00                            mov edx, [edx+0x464]
0x7715b94d    f7 c2 02 00 00 00                            test edx, 0x2
0x7715b953    74 03                                        jz 0x7715b958

0x7715b958    ea 5f b9 15 77 33 00                         jmp far 0x33:0x7715b95f

0x7715b95f    41 ff a7 f8 00 00 00                         jmp [r15+0xf8]

0x63fb1c46    49 87 e6                                     xchg r14, rsp
0x63fb1c49    45 8b 06                                     mov r8d, [r14]
0x63fb1c4c    49 83 c6 04                                  add r14, 0x4
0x63fb1c50    45 89 45 3c                                  mov [r13+0x3c], r8d
0x63fb1c54    45 89 75 48                                  mov [r13+0x48], r14d
...
```
The `jmp far` is used to switch from compatibility (i.e. `32-bit` protected) mode to `64-bit` mode. After this jump, the `PC` becomes `0x7715b95f`  and, (in the CPU context I've found that) the `cs` is updated to `0x33`.

tathanhdinh: > This function returns an `int32_t`, but `val` is a `uint64_t`. In this instance I like there to be explicit casting (via `static_cast`) so that what happens (sign extension, zero extension) is an obvious and deliberate choice by the programming, rather than implicitly added behaviour by the compiler.

You're right, I will modify the PR.

remill/Arch/X86/Arch.cpp
tathanhdinh: Done.

remill/Arch/X86/Arch.cpp
tathanhdinh: Done.

remill/Arch/X86/Semantics/UNCOND_BR.cpp
tathanhdinh: That makes sense to me, PR updated.

remill/Arch/X86/Semantics/UNCOND_BR.cpp
tathanhdinh: Done.

PR #346, opened 2019-04-03
added vpbroadcastb semantic to x86 and added a test for it. Also changed the name of MMX0-9 to MM0-9 to fit testing 

svv232:


tests/X86/AVX/VPBROADCASTB.S
```diff
@@ -0,0 +1,30 @@
...
+ * limitations under the License.
+ */
+
+#if HAS_FEATURE_AVX512
+ #define VPBROADCASTB_INPUTS_64 \
+      0x0000000000000000, 0x0000000000000000,  /* sanity check */\
+      0xfedcba9876543210, 0x0000000000000000,  /* arbitrarily chosen values */\
+      0xfedcba9876543210, 0xffffffffffffffff   /* other arbitrarily chosen values */\
+
+TEST_BEGIN_64(VPBROADCASTB_YMMqq_XMMb, 3)
```
pgoodman: Can you make this test case use `ymm` instead of `zmm`, and then make it conditions on `HAS_FEATURE_AVX`

svv232: changes made

PR #353, opened 2019-07-16
added tests and instructions for MOVBE_GPRv_MEMv_16/32

svv232:
I know it's implemented slightly unconventionally according to the macros, I'll change that if necessary

### DIFF
--- a/remill/Arch/X86/Semantics/DATAXFER.cpp
+++ b/remill/Arch/X86/Semantics/DATAXFER.cpp
@@ -45,6 +45,15 @@ DEF_SEM(MOVBE32, D dst, const S src) {
   return memory;
 }
 
+
+#if 64 == ADDRESS_SIZE_BITS
+  template <typename D, typename S>
+  DEF_SEM(MOVBE64, D dst, const S src) {
+    Write(dst, __builtin_bswap64(Read(src)));
+    return memory;
+  }
+#endif 
+
 template <typename D, typename S>
 DEF_SEM(MOVQ, D dst, S src) {
   UWriteV64(dst, UExtractV64(UReadV64(src), 0));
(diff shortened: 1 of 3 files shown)


tests/X86/DATAXFER/MOVBE.S
```diff
@@ -0,0 +1,28 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+TEST_BEGIN_64(MOVBE_GPRv_MEMv_16, 1)
+TEST_INPUTS(
+    0)
```
pgoodman: would recommend putting other inputs that should distinguish working from not working, then do
`push ARG1_64` or something like that (look at other tests), and tehn `movbe ..., [rsp]`

remill/Arch/X86/Semantics/DATAXFER.cpp
```diff
@@ -33,6 +33,18 @@ DEF_SEM(XCHG, D1 dst, S1 dst_val, D2 src, S2 src_val) {
...
+  WriteZExt(dst, __builtin_bswap16(Read(src)));
+  return memory;
+}
+
+template <typename D, typename S>
+DEF_SEM(MOVBE32, D dst, const S src) {
+  WriteZExt(dst, __builtin_bswap32(Read(src)));
+  return memory;
+}
+
```
pgoodman: Add a 64-bit implementation, guarded by `#if 64 == ADDRESS_SIZE_BITS`

remill/Arch/X86/Semantics/DATAXFER.cpp
```diff
@@ -143,6 +155,8 @@ DEF_ISEL(MOV_GPR8_IMMb_C6r0) = MOV<R8W, I8>;
 DEF_ISEL(MOV_MEMb_IMMb) = MOV<M8W, I8>;
 DEF_ISEL_RnW_In(MOV_GPRv_IMMz, MOV);
 DEF_ISEL_MnW_In(MOV_MEMv_IMMz, MOV);
+DEF_ISEL(MOVBE_GPRv_MEMv_16) = MOVBE16<R16W, M16>;
+DEF_ISEL(MOVBE_GPRv_MEMv_32) = MOVBE32<R32W, M32>;
 DEF_ISEL(MOV_GPR8_GPR8_88) = MOV<R8W, R8>;
```
pgoodman: Add 64-bit iself, wrapped in IF_64BIT(...)`

Issue #342, opened 2019-03-27
Missing AMD64 opcode (SHUFPD_XMMpd_XMMpd_IMMb)


pgarba:
Missing semantics for instruction (AMD64 402d5d 5 (BYTES 66 0f c6 e2 01) SHUFPD_XMMpd_XMMpd_IMMb (WRITE_OP (REG_128 XMM4)) (READ_OP (REG_128 XMM4)) (READ_OP (REG_128 XMM2)) (READ_OP (IMM_8 1)))

PR #363, opened 2019-10-26
implemented SHUFPD_XMMpd_XMMpd_IMMb

Toizi:
Implements the missing instruction from #342

It is my first instruction so let me know if there is something I could have done better.
I tried to implement the `SEM` in a way that it works for the AVX instruction set as well but figured I should start small so I only implemented the exact instruction.

The tests are just a copy and paste from `SHUFPS` plus an added line in `SHUFPDv128v128imm8_many` to also test the case of source and destination being the same register


pgoodman:
Thanks!!

PR #369, opened 2019-10-29
Unfold devel

Aiethel:
Pass that unfolds `State` structure into separate variables that are passed around by value.
Needs functions to be annotated to work properly (via `remill/BC/Annotate.*`)
Still experimental.


remill/BC/UnfoldUtils.h
```diff
@@ -177,28 +146,25 @@ struct _EscapeArgument {
...
-    );
+    if (auto phi = llvm::dyn_cast<llvm::PHINode>(val)) {
+      bool acc = SpreadForward(phi);
+      for (auto &use: phi->incoming_values()) {
+        acc &= Backward(use);
+      }
+      return acc;
+    }
+
+    if (auto call = llvm::dyn_cast<llvm::CallInst>(val)) {
```
pgoodman: `} else if`

PR #396, opened 2020-01-22
Added get Architecture from Module

aaronyoo:
A few fixed that are needed for anvill/bitcode-specifier to work. Created a function that allows you to get an architecture from a module.

### DIFF
--- a/remill/Arch/Arch.cpp
+++ b/remill/Arch/Arch.cpp
@@ -247,6 +247,21 @@ const Arch* GetTargetArch() {
   return Arch::Build(nullptr, GetOSName(FLAGS_os), GetArchName(FLAGS_arch)).get();
 }
 
+const remill::Arch *Arch::GetModuleArch(const llvm::Module &module) {
+  auto triple = llvm::Triple(module.getTargetTriple());
+  std::string arch_name = triple.getArchName().data();
+  std::string os_name = triple.getOSName().data();
+  // Since llvm::Triple.getArchName() incorrectly returns the entire triple,
+  // just extract the first part of the triple ourselves.
+  arch_name = arch_name.substr(0, arch_name.find('-'));
+  if (arch_name.empty() || os_name.empty()) {
+    return nullptr;
+  }
+  std::cout << os_name << " " << arch_name << std::endl;
+  return remill::Arch::Get(module.getContext(), remill::GetOSName(os_name),
+                           remill::GetArchName(arch_name));
+}
+
 bool Arch::IsX86(void) const {
   switch (arch_name) {
     case remill::kArchX86:
--- a/remill/Arch/Arch.cpp
+++ b/remill/Arch/Arch.cpp
@@ -249,8 +249,8 @@ const Arch* GetTargetArch() {
 
 remill::Arch::ArchPtr Arch::GetModuleArch(const llvm::Module &module) {
   const llvm::Triple triple = llvm::Triple(module.getTargetTriple());
-  return remill::Arch::Build(&module.getContext(), GetOSName(triple.getOS()),
-                             GetArchName(triple.getArch()));
+  return remill::Arch::Build(&module.getContext(), GetOSName(triple),
+                             GetArchName(triple));
 }
 
 bool Arch::IsX86(void) const {
--- a/remill/Arch/Arch.cpp
+++ b/remill/Arch/Arch.cpp
@@ -255,7 +255,8 @@ const remill::Arch *Arch::GetModuleArch(const llvm::Module &module) {
   // just extract the first part of the triple ourselves.
   arch_name = arch_name.substr(0, arch_name.find('-'));
   if (arch_name.empty() || os_name.empty()) {
-    return nullptr;
+    LOG(FATAL) << "Empty arch or os name: " << triple.getArchName().data()
+               << " " << triple.getOSName().data();
   }
   return remill::Arch::Get(module.getContext(), remill::GetOSName(os_name),
                            remill::GetArchName(arch_name));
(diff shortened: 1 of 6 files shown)


remill/Arch/Arch.cpp
```diff
@@ -255,6 +255,21 @@ const Arch *GetTargetArch(llvm::LLVMContext &context) {
...
+  auto triple = llvm::Triple(module.getTargetTriple());
+  std::string arch_name = triple.getArchName().data();
+  std::string os_name = triple.getOSName().data();
+  // Since llvm::Triple.getArchName() incorrectly returns the entire triple,
+  // just extract the first part of the triple ourselves.
+  arch_name = arch_name.substr(0, arch_name.find('-'));
+  if (arch_name.empty() || os_name.empty()) {
+    return nullptr;
+  }
+  std::cout << os_name << " " << arch_name << std::endl;
```
pgoodman: Remove `std::cout` logging here.

remill/Arch/Arch.cpp
```diff
@@ -255,6 +255,21 @@ const Arch *GetTargetArch(llvm::LLVMContext &context) {
...
 
+const remill::Arch *Arch::GetModuleArch(const llvm::Module &module) {
+  auto triple = llvm::Triple(module.getTargetTriple());
+  std::string arch_name = triple.getArchName().data();
+  std::string os_name = triple.getOSName().data();
+  // Since llvm::Triple.getArchName() incorrectly returns the entire triple,
+  // just extract the first part of the triple ourselves.
+  arch_name = arch_name.substr(0, arch_name.find('-'));
+  if (arch_name.empty() || os_name.empty()) {
+    return nullptr;
```
pgoodman: This returns `nullptr`, but `remill::Arch::Get` will `LOG(FATAL)` on an invalid os or arch name. I'd recommend having a more helpful yet fatal log message here where you log out the `triple.getArchName().data()` and `triple.getOSName().data()`.

remill/Arch/Arch.h
```diff
@@ -171,6 +172,9 @@ class Arch {
   // Get the LLVM DataLayout for this architecture.
   virtual llvm::DataLayout DataLayout(void) const = 0;
 
+  // Get the architecture related to a module.
+  static const remill::Arch *GetModuleArch(const llvm::Module&);
```
pgoodman: Space before the `&`

remill/Arch/Name.cpp
```diff
@@ -28,7 +28,7 @@ ArchName GetArchName(const std::string &arch_name) {
   } else if (arch_name == "x86_avx512") {
     return kArchX86_AVX512;
 
-  } else if (arch_name == "amd64") {
+  } else if (arch_name == "amd64" || arch_name == "x86_64") {
```
pgoodman: This logic should go in `GetModuleArch`. Then, you can also inspect the attributes (target features) of individual functions [1] to infer the `_avx` or `_avx512` suffixes. This could mean that in anvill, you would need to get the target arch on a per-function basis, possibly with some caching on attribute sets or the like.

[1] https://github.com/lifting-bits/remill/blob/master/remill/Arch/Arch.cpp#L835-L852

remill/Arch/Arch.cpp
aaronyoo: Fixed b61450c

remill/Arch/Arch.cpp
```diff
@@ -255,6 +255,21 @@ const Arch *GetTargetArch(llvm::LLVMContext &context) {
   return arch;
 }
 
+const remill::Arch *Arch::GetModuleArch(const llvm::Module &module) {
+  auto triple = llvm::Triple(module.getTargetTriple());
+  std::string arch_name = triple.getArchName().data();
```
pgoodman: Ideally I'd like you to use a `switch (triple.getArch()` and use that to select an arch, rather than doing string parrsing. Ditto for OS/environment/etc.

Aiethel:
Also please use `Arch::Build` https://github.com/lifting-bits/remill/blob/master/remill/Arch/Arch.h#L191, since `Arch::Get` is deprecated.

remill/Arch/Arch.cpp
aaronyoo: Fixed in c7b0e04

remill/OS/OS.cpp
```diff
@@ -24,6 +25,19 @@ DEFINE_string(os, REMILL_OS, "Operating system name of the code being "
 
 namespace remill {
 
+OSName GetOSName(const llvm::Triple::OSType &os) {
```
pgoodman: Lets make this take in the environment or file type here too, as a backup. So for example, if the file type is ELF, a.out, or COFF, default to linux, Mach-O for macOS, PE, etc. for Windows.

aaronyoo: Fixed in f12d641

remill/OS/OS.cpp
```diff
@@ -24,6 +25,31 @@ DEFINE_string(os, REMILL_OS, "Operating system name of the code being "
...
+    case llvm::Triple::OSType::Win32:
+      return kOSWindows;
+    default:
+      break;
+  }
+
+  switch (triple.getObjectFormat()) {
+    case llvm::Triple::ObjectFormatType::MachO:
+      return kOSmacOS;
+    case llvm::Triple::ObjectFormatType::COFF:
```
pgoodman: I think COFF ends up being windows.

aaronyoo: Fixed 7ed4bb4

PR #398, opened 2020-01-25
Arch struct improvement

Aiethel:
Improves `remill::Arch` in two ways:

1) Improves backward compatibility by allowing `Arch` to be constructed over empty `llvm::LLVMContext *`. This is mainly due to how KLEE-Native uses the `GetTargetArch()` function (all over the place without simpe way of fixing it).

2) Changes memory managements of these structures.
* Before: Pointers were "cached" in static variable of the getter based on `llvm::LLVMContext *`
* Now: Getter method returns `std::unique_ptr<const Arch>` and caller is responsible for lifetime of the object. Therefore call to the `Arch::Build` twice with the same arguments will return different objects. (Before it would return pointer to the same memory)
Old way is marked as deprecated but it is still present in API, since random removal can break a lot of stuff.

And I added one patch from `Annotate` by an accident. (But it is a valid enhacement) 

EDIT: Typos


remill/Arch/Arch.cpp
```diff
@@ -232,27 +214,54 @@ const Register *Arch::RegisterByName(const std::string &name) const {
   }
 }
 
-const Arch *Arch::GetMips(llvm::LLVMContext &, OSName, ArchName) {
+auto Arch::GetMips(llvm::LLVMContext *, OSName, ArchName) -> ArchPtr {
```
pgoodman: Lets remove `GetMips` and all MIPS related things while we're at it. We don't support it.

pgoodman:
Can you submit a PR to anvill that will change it, if necessary?

PR #402, opened 2020-02-17
Remill running on the web via Emscripten

TrevorSundberg:
See the README.md I added for notes about building and running. I'm very open to make changes or answer questions :)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -20,6 +20,29 @@ endif ()
 project(remill)
 cmake_minimum_required(VERSION 3.2)
 
+if(EMSCRIPTEN)
+  set(CMAKE_BUILD_TYPE Release)
+  set(gflags_DIR "$ENV{REPOS}/gflags/build")
+  set(glog_DIR "$ENV{REPOS}/glog/build")
+  set(LLVM_DIR "$ENV{REPOS}/llvm-project/llvm/build/lib/cmake/llvm/")
+  set(remill_DIR "$ENV{REPOS}/remill/build/")
+
+  # We want to manually invoke main for Lift instead of Emscripten calling it for us
+  # So we disable INVOKE_RUN, export callMain (and FS for files).
+  set(CMAKE_MODULE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/web/src")
+  set(CMAKE_CXX_FLAGS "\
+    ${CMAKE_CXX_FLAGS} \
+    $ENV{EM_CXX_FLAGS} \
+    -D__i386__ \
+    -DADDRESS_SIZE_BITS=32 \
+    -s ALLOW_MEMORY_GROWTH=1 \
+    --no-heap-copy \
+    -s INVOKE_RUN=0 \
+    -s EXTRA_EXPORTED_RUNTIME_METHODS='[\"callMain\",\"FS\"]' \
+    -s ASSERTIONS=1 \
+    --embed-file ${CMAKE_CURRENT_SOURCE_DIR}/web/build/remill/Arch/X86/Runtime/x86.bc@/share/remill/10.0/semantics/x86.bc")
+endif()
+
 include("${CMAKE_CURRENT_SOURCE_DIR}/cmake/settings.cmake")
 include("${CMAKE_CURRENT_SOURCE_DIR}/cmake/utils.cmake")
 include("${CMAKE_CURRENT_SOURCE_DIR}/cmake/BCCompiler.cmake")
@@ -227,7 +250,7 @@ if("${LIBRARY_REPOSITORY_ROOT}" STREQUAL "" OR NOT EXISTS "${LIBRARY_REPOSITORY_
   message("InstallExternalTarget: Found llvm-link executable: ${llvmlink_location}")
 
   find_library("libllvm_location" "${dynamic_lib_prefix}LLVM-${REMILL_LLVM_VERSION}.${dynamic_lib_extension}")
-  if("${libllvm_location}" STREQUAL "libllvm_location-NOTFOUND")
+  if("${libllvm_location}" STREQUAL "libllvm_location-NOTFOUND" AND NOT EMSCRIPTEN)
     message(FATAL_ERROR "InstallExternalTarget: Failed to locate the libLLVM dynamic library")
   endif()
   
@@ -313,7 +336,7 @@ add_subdirectory(remill/Arch/AArch64/Runtime)
 add_subdirectory(tools)
 
 # tests
-if ("${CMAKE_C_COMPILER_ID}" STREQUAL "Clang" OR "${CMAKE_C_COMPILER_ID}" STREQUAL "AppleClang")
+if ("${CMAKE_C_COMPILER_ID}" STREQUAL "Clang" OR "${CMAKE_C_COMPILER_ID}" STREQUAL "AppleClang" AND NOT EMSCRIPTEN)
   add_custom_target(test_dependencies)
 
   if(NOT "${PLATFORM_NAME}" STREQUAL "windows")
--- a/remill/OS/OS.h
+++ b/remill/OS/OS.h
@@ -24,7 +24,7 @@
 #   define REMILL_ON_LINUX 0
 #   define REMILL_ON_WINDOWS 0
 #   define REMILL_OS "macos"
-# elif defined(__linux__)
+# elif defined(__linux__) || defined(EMSCRIPTEN)
 #   define REMILL_ON_MACOS 0
 #   define REMILL_ON_LINUX 1
 #   define REMILL_ON_WINDOWS 0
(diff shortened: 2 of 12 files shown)


CMakeLists.txt
```diff
@@ -20,6 +20,29 @@ endif ()
 project(remill)
 cmake_minimum_required(VERSION 3.2)
 
+if(EMSCRIPTEN)
```
alessandrogario: Can this be defined under cmake/settings.cmake as `option(REMILL_EMSCRIPTEN "description")`?
This makes sure the option will show up as a configuration setting in the project (i.e. make edit_cache)

CMakeLists.txt
```diff
@@ -20,6 +20,29 @@ endif ()
...
+  set(CMAKE_BUILD_TYPE Release)
+  set(gflags_DIR "$ENV{REPOS}/gflags/build")
+  set(glog_DIR "$ENV{REPOS}/glog/build")
+  set(LLVM_DIR "$ENV{REPOS}/llvm-project/llvm/build/lib/cmake/llvm/")
+  set(remill_DIR "$ENV{REPOS}/remill/build/")
+
+  # We want to manually invoke main for Lift instead of Emscripten calling it for us
+  # So we disable INVOKE_RUN, export callMain (and FS for files).
+  set(CMAKE_MODULE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/web/src")
```
alessandrogario: This variable may contain a value already; can this be changed to an APPEND instead?
Example: `list(APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/web/src")`

CMakeLists.txt
```diff
@@ -20,6 +20,29 @@ endif ()
 project(remill)
 cmake_minimum_required(VERSION 3.2)
 
+if(EMSCRIPTEN)
+  set(CMAKE_BUILD_TYPE Release)
```
alessandrogario: Can this be moved at the top of the cmake/settings.cmake file? If this new feature does not support other build configurations (such as Debug), we could add a warning message and overwrite the setting.

Example:
```
if(REMILL_EMSCRIPTEN)
  ..
  if(NOT "${CMAKE_BUILD_TYPE}" STREQUAL "Release")
    message(WARNING "remill: The REMILL_EMSCRIPTEN overwrites the build type to Release")
    set(CMAKE_BUILD_TYPE "Release" CACHE STRING "Build type" FORCE)
  endif()
endif()
```

CMakeLists.txt
```diff
@@ -20,6 +20,29 @@ endif ()
...
+  set(gflags_DIR "$ENV{REPOS}/gflags/build")
+  set(glog_DIR "$ENV{REPOS}/glog/build")
+  set(LLVM_DIR "$ENV{REPOS}/llvm-project/llvm/build/lib/cmake/llvm/")
+  set(remill_DIR "$ENV{REPOS}/remill/build/")
+
+  # We want to manually invoke main for Lift instead of Emscripten calling it for us
+  # So we disable INVOKE_RUN, export callMain (and FS for files).
+  set(CMAKE_MODULE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/web/src")
+  set(CMAKE_CXX_FLAGS "\
```
alessandrogario: Is it possible to move these settings in cmake/settings.cmake and use GLOBAL_CXX_FLAGS instead of CMAKE_CXX_FLAGS?

CMakeLists.txt
```diff
@@ -20,6 +20,29 @@ endif ()
...
+    ${CMAKE_CXX_FLAGS} \
+    $ENV{EM_CXX_FLAGS} \
```
alessandrogario: Can this setting be moved away from the environment variable?

Example, in cmake/settings.cmake
```
set(REMILL_EMSCRIPTEN_CXXFLAGS "default_value" CACHE STRING "Semicolon separated list of flags")
```

This will make the flag list show up as a configurable setting (i.e.: make edit_cache)

remill/OS/OS.h
```diff
@@ -24,7 +24,7 @@
 #   define REMILL_ON_LINUX 0
 #   define REMILL_ON_WINDOWS 0
 #   define REMILL_OS "macos"
-# elif defined(__linux__)
+# elif defined(__linux__) || defined(EMSCRIPTEN)
```
alessandrogario: I was looking for this definition, but I couldn't find it in this PR. Is this a non documented value provided by the toolchain? Docs: https://emscripten.org/docs/compiling/Building-Projects.html#detecting-emscripten-in-preprocessor

CMakeLists.txt
```diff
@@ -20,6 +20,29 @@ endif ()
...
+    -D__i386__ \
```
pgoodman: Maybe you want `-m32` here? Also, probably `-mno-sse`, `-mno-avx`, and perhaps `-freestanding`?

CMakeLists.txt
TrevorSundberg: Woops! Good find :)

CMakeLists.txt
TrevorSundberg: Just to be certain are you suggesting replacing all instances of `EMSCRIPTEN` with `REMILL_EMSCRIPTEN` and then defining that in the `generate.sh` script?

I can certainly make this change, but it seems prone to error since if you try to build with the EMSCRIPTEN.cmake toolchain it will certainly fail without defining `REMILL_EMSCRIPTEN`, and visa versa. I haven't actually used make edit_cache before, but if the purpose is to change the cached variables on the fly it seems like that would break in this case since we defined the toolchain when we called cmake. I'll take your advice on this one however.

remill/OS/OS.h
TrevorSundberg: Yeah it looks like it's undocumented, but a good find since they now have the documented version __EMSCRIPTEN__. I'll change it to using that :)

As a word of caution, Emscripten's documentation is notoriously bad. I think it comes from how fast that project evolved that the written documentation was quickly broken by new features. For the source of truth I usually clone the emscripten repo and search through it rather than look at docs:

https://github.com/emscripten-core/emscripten/blob/430c2911dbe0104beab264cd19a971a2a7c6b793/emcc.py#L1264

TrevorSundberg:
This is my first real GitHub PR (I'm used to Azure Devops), when I commit and push the comments that were made just show `Outdated` and the old code. In Devops when you push the new code is shown next to the comment so the reviewers can verify that it was changed. Is this normal GitHub behavior or is there something I'm supposed to do to tell the PR to use the latest commit? Seeing the `Outdated` just seems weird.

CMakeLists.txt
TrevorSundberg: Yeah I love this idea!

CMakeLists.txt
TrevorSundberg: I'll certainly try. I don't see any reference to `GLOBAL_CXX_FLAGS` in CMake docs, or used anywhere in Remill. I also don't see it being respected anywhere in Emscripten. Is this a feature I'm unaware of?

CMakeLists.txt
TrevorSundberg: With `-ffreestanding` it breaks the closure compiler (a Java tool that Emscripten runs to minify and elide JS functions). I have a feeling this option is untested under Emscripten.

As for the rest, with those specified it does compile, although I don't truly know if they have any effect. Emscripten for the longest time already had SSE disabled (and probably AVX), which I think is up-streamed into clang for the WASM backend. Let me know if you'd like me to check this change in anyways.

CMakeLists.txt
alessandrogario: I initially thought it could be turned on and off with a setting, thanks for explaining how it works! It seems like your approach is the most correct one

CMakeLists.txt
alessandrogario: The goal is trying to not edit the CMake CXX variables by hand, and append them to the variables used there (which will eventually use property-based target settings such as `target_compile_options`)

I think it is fine to append them directly with something similar to

```cmake
if(EMSCRIPTEN)
  list(GLOBAL_CXX_FLAGS APPEND
    flag1
    flag2
  )
endif()
```
but you can also create a setting such as `REMILL_EMSCRIPTEN_CXXFLAGS` if you think it may be useful

pgoodman:
I think we might need some changes to cxx-common to add wasm as a supported target.

pgoodman:
Hey @TrevorSundberg , just as an FYI, we've all been on a company trip hence the lack of movement. What do you think of us using a cxx-common built clang for the compiler to produce WASM targets? It also seems like LLVM has some kind of support for wasm with 64-bit pointers [1].

[1] https://code.woboq.org/llvm/llvm/include/llvm/ADT/Triple.h.html#llvm::Triple::wasm64

CMakeLists.txt
pgoodman: Has this been resolved?

PR #404, opened 2020-03-24
Fix IDIV instruction source operands not being shifted properly.

Aiethel:


remill/Arch/X86/Semantics/BINARY.cpp
```diff
@@ -474,7 +474,7 @@ IF_64BIT(MAKE_DIVxax(rdxrax, REG_RAX, REG_RDX, REG_RAX, REG_RDX))
       auto lhs_high = ZExt(Read(src2)); \
       auto rhs = SExt(Read(src3)); \
       auto shift = ZExt(BitSizeOf(src3)); \
-      auto lhs = Signed(UOr(UShl(lhs_high, shift), lhs_low)); \
+      auto lhs = Signed(UOr(UShr(lhs_high, shift), lhs_low)); \
```
pgoodman: Why is this switched to a `UShr`?

remill/Arch/X86/Semantics/BINARY.cpp
```diff
@@ -491,7 +491,7 @@ IF_64BIT(MAKE_DIVxax(rdxrax, REG_RAX, REG_RDX, REG_RAX, REG_RDX))
 
 MAKE_IDIVxax(ax, REG_AL, REG_AH, REG_AL, REG_AH)
 MAKE_IDIVxax(dxax, REG_AX, REG_DX, REG_AX, REG_DX)
-MAKE_IDIVxax(edxeax, REG_EAX, REG_EDX, REG_XAX, REG_XDX)
+MAKE_IDIVxax(edxeax, REG_EAX, REG_EDX, REG_EAX, REG_EDX)
```
pgoodman: In 64-bit, the dest regs need to be `rax` and `rdx`, hence the use of `REG_XAX` and `REG_XDX`.

pgoodman:
I am closing this because I don't think it's right.

PR #424, opened 2020-05-24
General API improvements

pgoodman:


artemdinaburg:
Pretty massive commit :). The failures look like tests fails with one specific thing, i think it may be FPU data related.

PR #461, opened 2020-11-19
New arch support

sschriner:


pgoodman (approved):

pgoodman:
uhh I clicked approve too soon

lib/Arch/AArch32/Semantics/BINARY.cpp
```diff
@@ -0,0 +1,540 @@
...
+DEF_ISEL(UMLALS) = UMULLS;
+DEF_ISEL(SMULL) = SMULL;
+DEF_ISEL(SMULLS) = SMULLS;
+DEF_ISEL(SMLAL) = SMULL;
+DEF_ISEL(SMLALS) = SMULLS;
+
+// Halfword Multiply and Accumulate
+namespace {
+DEF_COND_SEM(SMLAh, R32W dst, R32 src1, R32 src2, R32 src3) {
+  auto rhs = SExt<uint64_t>(Read(src2));
```
pgoodman, line 275: Normally I use `SExtTo` when specifying a destination type. In this case, you've specified `uint64_t` (an unsigned type), but the operator will convert to `int64_t` via `WST` here: https://github.com/lifting-bits/remill/blob/master/include/remill/Arch/Runtime/Operators.h#L733

I think by specifying `uint64_t` here, what's happening is a sign extension to an `int128_t`; see this: https://github.com/lifting-bits/remill/blob/master/include/remill/Arch/Runtime/Types.h#L664-L666

Is this the intention?

pgoodman (changes requested):

pgoodman (changes requested):
Only change requested is to run `scripts/format-files` (or whatever the code formatter is called)

:-)

Don't merge this at the end of a day; we'll need to quickly follow up with changes to Anvill and Circuitous.

artemdinaburg:
Is this ready for merging :) ?

pgoodman (approved):

lib/BC/TraceLifter.cpp
```diff
@@ -571,6 +571,7 @@ bool TraceLifter::Impl::Lift(
             goto direct_func_call;
           }
           try_add_delay_slot(true, block);
+          try_add_delay_slot(false, block);
```
pgoodman, line 574: This isn't quite right, what's needed is someting like:

```
block:
  br condition, true_block, false_block

true_block:
  lift delayed true
  call function
  br not_taken_pc

false_block:
  lift delayed false
  br not_taken_pc

```

pgoodman (changes requested):

PR #465, opened 2020-12-15
Replace old build system entirely with vcpkg

ekilmer:
DO NOT MERGE

Update README build directions

TODO:

- [ ] Double-check README is update appropriately with links, directions, etc.
- [x] Docker support---Currently only builds 1 combination (LLVM 11 and Ubuntu:20.04), since we test the other combinations already.
- [ ] Releases---We want to be able to upload an archive of the built installation directory to be used in other similar environments (Docker)


ekilmer:
@alessandrogario I'm fine with this PR being superseded by anything you have time to work on with respect to cleaning up the CMake more 🙂

PR #469, opened 2020-12-28
Llvm 7 build

Aiethel:
Build for llvm-7 (and possibly others lower than 8) version fails because of some API changes. This PR tries to fix that.

### DIFF
--- a/include/remill/BC/Compat/CallSite.h
+++ b/include/remill/BC/Compat/CallSite.h
@@ -1,0 +1,104 @@
+/*
+ * Copyright (c) 2021 Trail of Bits, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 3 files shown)


pgoodman (approved):

artemdinaburg:
@Aiethel it looks like CallSite has been removed in LLVM-11; relevant error:

```
/__w/remill/remill/lib/BC/DeadStoreEliminator.cpp:1191:20: error: no member named 'CallSite' in namespace 'llvm'
  auto val = llvm::CallSite(&inst).getCalledValue()->stripPointerCasts();
             ~~~~~~^
3 errors generated.
```

Its replacement is [AbstractCallSite](https://llvm.org/doxygen/classllvm_1_1AbstractCallSite.html) with a slightly different API. This would require hacking in some compatibility headers.

Aiethel:
Yeah, I thought that `llvm::AbstractCallSite` was added along the original `llvm::CallSite`. 

I wrapped the compatibility class into `remill::compat::llvm` namespace to make the compatibility glue more easily recognizable.

Aiethel:
Failing test case seems to be unrelated to the changes?

ekilmer:
> Failing test case seems to be unrelated to the changes?

Yes, there is a flaky test on Mac, and the Mac OS 11.0 runners are very inconsistent. CI looks good!

PR #470, opened 2021-01-13
Update README

artemdinaburg:
* Update README explaining how to do full source builds against vcpkg-based cxx-common
* Explain how to solve a frustrating build issue

### DIFF
--- a/README.md
+++ b/README.md
@@ -148,7 +148,7 @@ mkdir build
 cd build
 cmake  \
   -DCMAKE_INSTALL_PREFIX="<path where remill will install>" \
-  -DVCPKG_ROOT="<path to cxx-common diretory>/vcpkg"  \
+  -DVCPKG_ROOT="<path to cxx-common directory>/vcpkg"  \
   -G Ninja  \
   ..
 cmake --build .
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -138,3 +138,39 @@ cd ./remill-build
...
+### Full Source Builds
+
+Sometimes, you want to build everything from source, including the [cxx-common](https://github.com/trailofbits/cxx-common) libraries remill depends on. To build against a custom cxx-common location, you can use the following `cmake` invocation:
+
+```sh
+mkdir build
+cd build
+cmake  \
+  -DCMAKE_INSTALL_PREFIX="<path where remill will install>" \
+  -DVCPKG_ROOT="<path to cxx-common diretory>/vcpkg"  \
```
ekilmer, line 151:
```suggestion
  -DVCPKG_ROOT="<path to cxx-common directory>/vcpkg"  \
```

README.md
```diff
@@ -138,3 +138,39 @@ cd ./remill-build
...
+  -G Ninja  \
+  ..
+cmake --build .
+cmake --build . --target install
+```
+
+The output may produce some cmake warnings about policy CMP0003. These warnings are safe to ignore.
```
ekilmer, line 158:
```suggestion
The output may produce some CMake warnings about policy CMP0003. These warnings are safe to ignore.
```

README.md
```diff
@@ -138,3 +138,39 @@ cd ./remill-build
...
+
+Then you need to install 32-bit libstdc++ headers and libraries. On a Debian/Ubuntu based distribution, You would want to do something like this:
+
+```sh
+sudo dpkg --add-architecture i386
+sudo apt-get update
+sudo apt-get install libc6-dev:i386 install libstdc++-10-dev:i386 g++-multilib
+```
+
+This error happens because the SPARC32 semantics are built as 32-bit code, but 32-bit development libraries are not installed by default. A similar situation occurs when building on arm64 Linux. In that case, you want to follow a similar workflow, except the architecture that you'd add would be `armhf` instead of `i386`. Another alternative is to not build SPARC32 semantics.
```
ekilmer, line 176: > Another alternative is to not build SPARC32 semantics.

I think if we're going to mention this, we should tell people how to exclude the SPARC32 semantics.

ekilmer (changes requested):
Thank you for the update to the docs! Just a few comments and typos.

Also, I am planning to rework this README and the vcpkg.README.md to combine them into a single file in #465 but we can merge this PR first if we want.

README.md
pgoodman: Can you try with the msp430 branch? It has a bunch of compatbility headers that might get things closer to working out.

artemdinaburg: Let me write an edit on how to do it; this is a good idea

artemdinaburg: This should now be addressed with a new cmake option and an explanation for how to use it.

ekilmer (approved):
LGTM!

PR #480, opened 2021-02-08
Allow to build on Windows with Debug build type

ekilmer:


artemdinaburg (commented):
From an initial glance, shouldn't this be more like:

```
"$<$<CONFIG:Debug>:/MDd>$<$<CONFIG:Release>:/MD>"
```

?

ekilmer:
> From an initial glance, shouldn't this be more like:
> 
> ```
> "$<$<CONFIG:Debug>:/MDd>$<$<CONFIG:Release>:/MD>"
> ```
> 
> ?

Hmmm. Yeah, I think this should be equivalent 👍

artemdinaburg (approved):

PR #486, opened 2021-02-18
CMake: Add packaging support

alessandrogario:


ekilmer (approved):
Haven't tested it, but looks very similar to what is done in anvill 👍 

Aside from running some smoketests (testing if executables work, testing if CMake config/logic works for projects using remill as a dependency) in CI after installing the packaging artifacts, this looks good to me.

PR #487, opened 2021-02-19
CI: Automatically generate releases when pushing tags

alessandrogario:


scripts/build.sh
```diff
@@ -27,6 +27,7 @@ LLVM_VERSION=llvm-9
 OS_VERSION=
 ARCH_VERSION=
 BUILD_FLAGS=
+CXX_COMMON_VERSION="v0.1.1"
```
ekilmer, line 30: I really should have done this at the beginning. Hopefully it wasn't too much headache 😬 

I _think_ `v0.1.2` should be all done now. Can we try that bump here?

.github/workflows/vcpkg_ci.yml
```diff
@@ -127,4 +127,105 @@ jobs:
...
+        uses: actions/create-release@v1
+
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
+        with:
+          tag_name: ${{ github.ref }}
+          release_name: Version ${{ github.ref }}
+          draft: true
+          prerelease: true
```
ekilmer, line 156: Do these need to change?

alessandrogario: This was done on purpose, so that we still have a chance to edit the release before it goes public

ekilmer (approved):
LGTM

PR #498, opened 2021-03-20
Use _WIN32 instead of WIN32 to fix Windows builds

mrexodia:
See https://docs.microsoft.com/en-us/cpp/preprocessor/predefined-macros?view=msvc-160

With this fix I can compile remill with clang-cl and with `-DVCPKG_ROOT=` https://github.com/trailofbits/cxx-common/releases/download/v0.1.2/vcpkg_windows-vs2019_llvm-11_amd64.7z

### DIFF
--- a/lib/BC/Util.cpp
+++ b/lib/BC/Util.cpp
@@ -23,7 +23,7 @@
 #include <utility>
 #include <vector>
 
-#ifndef WIN32
+#ifndef _WIN32
 #  include <sys/stat.h>
 #  include <unistd.h>
 #endif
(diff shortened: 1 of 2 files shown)


ekilmer (approved):
Thank you for the fixes! 

Not sure how I didn't run into this earlier. CMake seems to set `WIN32` when compiling on Windows, but I think it's better to use the standard define 👍 

https://github.com/Kitware/CMake/blob/6d4cbf7416f8ab4d001ca08ca4e108615ea5dfb1/Modules/Platform/Windows-MSVC.cmake#L224

mrexodia:
I didn’t see that before, interesting... In my case I was using clang-cl so
maybe that’s why 🤷🏻‍♂️ Either way, thanks for the quick merge!

PR #499, opened 2021-03-31
PC relative handling and instr category for AArch32

kumarak:
The changes fix the PC relative handling and return type instruction category for AArch32 support.


lib/Arch/AArch32/Decode.cpp
```diff
@@ -999,7 +999,8 @@ static bool DecodeCondition(Instruction &inst, uint32_t cond) {
 }
 
 std::optional<uint64_t> EvalReg(const Instruction &inst,
-                                const Operand::Register &op) {
+                                const Operand::Register &op, bool &is_linkreg) {
+  is_linkreg = (op.name == kIntRegName[kLRRegNum] || op.name == "LR");
```
pgoodman, line 1003: This needs to be saturating:
```c++
if (!is_linkreg) {
  is_linkreg = (op.name == kIntRegName[kLRRegNum] || op.name == "LR");
}
```

pgoodman: Also rename to `uses_linkreg`, as `uses` communicates saturation, whereas `is` does not.

lib/Arch/AArch32/Decode.cpp
```diff
@@ -1120,10 +1127,31 @@ static bool EvalPCDest(Instruction &inst, const bool s, const unsigned int rd,
       inst.category = Instruction::kCategoryError;
       return false;
     } else {
-      auto src1 = EvalOperand(inst, inst.operands[1]);
-      auto src2 = EvalOperand(inst, inst.operands[2]);
 
-      if (!src1 || !src2) {
+      // check if the number of operands in instruction is 5
```
pgoodman, line 1131: Explain this check in a comment, it's not obvious why `5` is significant or always right for all users of `EvalPCDest`.

lib/Arch/AArch32/Decode.cpp
```diff
@@ -2253,8 +2302,13 @@ static bool TryLogicalArithmeticRRRI(Instruction &inst, uint32_t bits) {
     AddImmOp(inst, ~0u);
   }
 
-  AddShiftRegImmOperand(inst, enc.rm, enc.type, enc.imm5, enc.s);
-  return EvalPCDest(inst, enc.s, enc.rd, kLogArithEvaluators[enc.opc >> 1u],
+  if (enc.type) {
+	  AddShiftRegImmOperand(inst, enc.rm, enc.type, enc.imm5, enc.s);
```
pgoodman, line 2306: Run code formatter to fix indentation; `./scripts/format-files` I believe. Make sure to copy the `format-files` from anvill first; I think it's script is more up-to-date.

pgoodman (changes requested):

pgoodman (approved):

lib/Arch/AArch32/Decode.cpp
kumarak: Updated comments about the operand count check. This is a hack and I will relook the implementation of `EvalPCDest`. The previous implementation was not doing anything because it was not evaluating the source operands and also not handling the expression operands.

PR #507, opened 2021-04-15
Fix FCOMPmem and FUCOMPmem

### LINKED ISSUE #504: Infinite recursion in the semantics
It seems that the implementation of the FCOMP (and FUCOMP) with memory argument results in an infinite recursion:

```
template <typename S2>
DEF_FPU_SEM(FUCOMPmem, RF80 src1, S2 src2) {
  SetFPUDp(src2);
  return FUCOMPmem(memory, state, src1, src2, pc, fop);
}

template <typename S2>
DEF_FPU_SEM(FCOMPmem, RF80 src1, S2 src2) {
  SetFPUDp(src2);
  return FCOMPmem(memory, state, src1, src2, pc, fop);
}
```

Probably some other function should have been called...

DCNick3:
Closes #504

### DIFF
--- a/lib/Arch/X86/Semantics/X87.cpp
+++ b/lib/Arch/X86/Semantics/X87.cpp
@@ -1112,17 +1112,13 @@ DEF_FPU_SEM(FCOMP, RF80 src1, S2 src2) {
 template <typename S2>
 DEF_FPU_SEM(FUCOMPmem, RF80 src1, S2 src2) {
   SetFPUDp(src2);
-  memory = FUCOMP(memory, state, src1, src2, pc, fop);
-  (void) POP_X87_STACK();
-  return memory;
+  return FUCOMP(memory, state, src1, src2, pc, fop);
 }
 
 template <typename S2>
 DEF_FPU_SEM(FCOMPmem, RF80 src1, S2 src2) {
   SetFPUDp(src2);
-  memory = FCOMP(memory, state, src1, src2, pc, fop);
-  (void) POP_X87_STACK();
-  return memory;
+  return FCOMP(memory, state, src1, src2, pc, fop);
 }
 
 DEF_FPU_SEM(DoFUCOMPP) {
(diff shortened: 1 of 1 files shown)


lib/Arch/X86/Semantics/X87.cpp
```diff
@@ -1112,13 +1112,17 @@ DEF_FPU_SEM(FCOMP, RF80 src1, S2 src2) {
 template <typename S2>
 DEF_FPU_SEM(FUCOMPmem, RF80 src1, S2 src2) {
   SetFPUDp(src2);
-  return FUCOMPmem(memory, state, src1, src2, pc, fop);
+  memory = FUCOMP(memory, state, src1, src2, pc, fop);
+  (void) POP_X87_STACK();
```
pgoodman, line 1116: Should a double pop be happening? `FUCOMP` pops already, I believe.

lib/Arch/X86/Semantics/X87.cpp
```diff
@@ -1112,13 +1112,17 @@ DEF_FPU_SEM(FCOMP, RF80 src1, S2 src2) {
...
+  return memory;
 }
 
 template <typename S2>
 DEF_FPU_SEM(FCOMPmem, RF80 src1, S2 src2) {
   SetFPUDp(src2);
-  return FCOMPmem(memory, state, src1, src2, pc, fop);
+  memory = FCOMP(memory, state, src1, src2, pc, fop);
+  (void) POP_X87_STACK();
```
pgoodman, line 1124: Should a double pop be happening? `FCOMP` pops already, I believe.

lib/Arch/X86/Semantics/X87.cpp
DCNick3: Yeah, you are right

lib/Arch/X86/Semantics/X87.cpp
DCNick3: Yeah, you are right

DCNick3: Yeah, you are right

DCNick3:
Does not seem to me that this CI failure is related to my changes...

Intel manual says this about the mismatching leaf:
image

Seems to me this is environment-specific, probably should be ignored in the test case

pgoodman (approved):

Issue #527, opened 2021-06-25
Emit native architecture size FPU types


artemdinaburg:
Right now remill uses `double` as the FPU type, but different architectures have higher precision native FPUs. For example, x86 has an 80-bit native FPU type, and SPARC has 128-bit FPUs.

Emit native-sized FPU types based on target architecture.

PR #536, opened 2021-08-16
WIP: Fix FPU types

kumarak:
Fix FPU types for 80-bits floats. Some of the test cases are failing and need to be fixed. 

Ref #527, https://github.com/lifting-bits/anvill/issues/222

### DIFF
--- a/include/remill/Arch/Runtime/Operators.h
+++ b/include/remill/Arch/Runtime/Operators.h
@@ -1540,8 +1540,8 @@ MAKE_BUILTIN(FSin, __builtin_sin)
 MAKE_BUILTIN(FTan, __builtin_tan)
 MAKE_BUILTIN(FAtan,__builtin_atan)
 MAKE_BUILTIN(FSqrt,__builtin_sqrt)
-MAKE_BUILTIN(FExp2,__builtin_exp2)
-MAKE_BUILTIN(FLog2,__builtin_log2)
+MAKE_BUILTIN(Exp2,__builtin_exp2)
+MAKE_BUILTIN(Log2,__builtin_log2)
 
 MAKE_BUILTIN(FRoundUsingMode, __builtin_nearbyint);
 MAKE_BUILTIN(FTruncTowardZero, __builtin_trunc);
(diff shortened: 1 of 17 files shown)


include/remill/Arch/Runtime/Types.h
```diff
@@ -64,12 +65,56 @@ static_assert(4 == sizeof(float32_t), "Invalid `float32_t` size.");
...
+
+  // TODO: float80_t is backed by float64_t since we would
+  //       not like to deal with converting long double to
+  //       extended-precision format. It will be error prone
+  //       and may pollute semantics.
+  //
+  union{
+    uint16_t data[5];
+    struct {
+      float64_t f64;
```
kumarak, line 87: I am using `fp64` as the backing for `float80_t`. This is because cutting the padding bits from long double causes most of the test cases to fail. In some cases the ST registers output mismatch and in other cases the FPU exception flags are not getting set correctly. 

The FPU status flag differences are due to the extended-precision format. And some of them can be resolved by fixing the utility functions.

kumarak: One approach is to extend `float80_t` to the size of `long double` and let extra bits run into the paddings. This will not require cutting off the bits and causing issues. @pgoodman might be suggesting the same approach earlier.

include/remill/Arch/Runtime/Operators.h
```diff
@@ -560,6 +576,13 @@ ALWAYS_INLINE uint8_t issignaling(float64_t x) {
   return x_nan.exponent == 0x7FFU && !x_nan.is_quiet_nan && x_nan.payload;
 }
 
+ALWAYS_INLINE uint8_t issignaling(float80_t x) {
```
pgoodman, line 579: Return `bool`. Why not decompose the float 80 directly?? Downcasting will presumably signal if it is a signalling NaN...?

pgoodman: Lower down in the code there is ```ALWAYS_INLINE static uint8_t IsSignalingNaN(float80_t x) {
  const nan80_t x_nan = {x};
  return x_nan.exponent == 0x7FFFU && !x_nan.is_quiet_nan && x_nan.payload && x_nan.interger_bit;
}``` which looks more correct.

pgoodman (changes requested):

artemdinaburg (approved):

PR #539, opened 2021-09-02
fix remill intrinsics for float80

kumarak:
Fix float80 intrinsics to avoid casting issue due to struct unrolling

### DIFF
--- a/include/remill/Arch/Runtime/Intrinsics.h
+++ b/include/remill/Arch/Runtime/Intrinsics.h
@@ -57,7 +57,8 @@ __remill_write_memory_64(Memory *, addr_t, uint64_t);
 [[gnu::used, gnu::const]] extern float64_t __remill_read_memory_f64(Memory *,
                                                                     addr_t);
 
-[[gnu::used]] extern float80_t __remill_read_memory_f80(Memory *, addr_t);
+[[gnu::used]] extern Memory* __remill_read_memory_f80(Memory *, addr_t,
+                                                      native_float80_t&);
 
 [[gnu::used]] extern float128_t __remill_read_memory_f128(Memory *, addr_t);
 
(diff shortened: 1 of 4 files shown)


include/remill/Arch/Runtime/Intrinsics.h
```diff
@@ -68,7 +69,7 @@ __remill_write_memory_f32(Memory *, addr_t, float32_t);
 __remill_write_memory_f64(Memory *, addr_t, float64_t);
 
 [[gnu::used]] extern Memory *__remill_write_memory_f80(Memory *, addr_t,
-                                                       float80_t);
+                                                       native_float80_t);
```
pgoodman, line 72: `const native_float80_t &`

pgoodman: Maybe? Dunno if this actually matters.

kumarak: I will make the changes but it should not matter. It will require additional fixes in Lower memory op pass to handle this variant of write intrinsics.

include/remill/Arch/Runtime/Operators.h
```diff
@@ -157,6 +157,24 @@ MAKE_MREAD(128, 128, uint, 128)
...
+  } \
+\
+  ALWAYS_INLINE static type_prefix##ret_size##_t _Read( \
+      Memory *&memory, MnW<type_prefix##size##_t> op) { \
+	  native_##type_prefix##ret_size##_t val; \
+    memory = __remill_read_memory_##access_suffix(memory, op.addr, val); \
+    return val; \
+  }
+
 MAKE_MREAD(80, 80, float, f80)
```
pgoodman, line 178: Do you need a different parameter here, or maybe just manually implement this, so that the `val` is typed as `native_float80_t`?

kumarak: The `val` type needed to be `native_float80_t` for the intrinsics. Manually implement this: Do you men not use the intrinsics for the read operator?

pgoodman (approved):

Issue #541, opened 2021-09-09
Uniform handling of undefined aflags


lkorenc:
After some instructions, value inside some arithmetic flags may be undefined. X86 semantics currently handle this non-uniformly by either using `__remill_undefined_*` intrinsics, storing arbitrary values, or ignoring the undefined value altogether.

Ideal solution would be to always use `__remill_undefined_*` but add an extra argument, that can be used to indicate a preferred value.
Unfortunately a list of all affected instructions is rather long. Non-exhaustive list of instructions to check:

- [ ] IDIV
- [ ] ROR, RCL, ...
- [ ] BTS, BTR, BTC, ...

PR #543, opened 2021-09-21
Explicitly call __remill_undef for undefined arith flags.

lkorenc:
Implements part of #541


lkorenc:
@pgoodman Should be good to go. Failing test case is instruction (`cpuid`) this PR does not touch so I am not sure why that happened.

pgoodman (approved):

Issue #554, opened 2021-10-21
Attributes for memory read/write intrinsics


tathanhdinh:
In lifting some instructions which read/write memory, e.g. `mov rax, [rcx]`:

```
define dso_local %struct.Memory* @sub_0(%struct.State* noalias nonnull align 1 %state, i64 %pc, %struct.Memory* noalias %memory) local_unnamed_addr #0 {
  %RCX_ptr = getelementptr inbounds %struct.State, %struct.State* %state, i64 0, i32 6, i32 5, i32 0, i32 0
  %RAX_ptr = getelementptr inbounds %struct.State, %struct.State* %state, i64 0, i32 6, i32 1, i32 0, i32 0
  %PC_ptr = getelementptr inbounds %struct.State, %struct.State* %state, i64 0, i32 6, i32 33, i32 0, i32 0
  %1 = add i64 %pc, 3
  %RCX_ = load i64, i64* %RCX_ptr, align 8
  %2 = tail call i64 @__remill_read_memory_64(%struct.Memory* %memory, i64 %RCX_) #3
  store i64 %2, i64* %RAX_ptr, align 8
  store i64 %1, i64* %PC_ptr, align 8
  %3 = tail call %struct.Memory* @__remill_missing_block(%struct.State* %state, i64 %1, %struct.Memory* %memory)
  ret %struct.Memory* %3
}

; Function Attrs: noduplicate noinline nounwind optnone readnone willreturn
declare dso_local i64 @__remill_read_memory_64(%struct.Memory*, i64) #1
```
I suppose that `__remill_read_memory_64` will be implemented (by some program which uses remill) when needed and inlined, so no special attribute is needed here. But I still wonder, if it would be better if we add some attributes for such an intrinsic (which may help the optimizer!?), e.g.

```
; Function Attrs: noduplicate noinline nounwind optnone readonly willreturn
declare dso_local i64 @__remill_read_memory_64(%struct.Memory* readonly, i64) #1
```

PR #555, opened 2021-10-21
Add attributes for memory intrinsics

tathanhdinh:
Hello all,

This is not a ready PR (does not compile yet with LLVM below 3.9, an only for a single 64-bit read intrinsic), just for discussion at this [issue](https://github.com/lifting-bits/remill/issues/554).

I still do not know the interference between `readnone`, set by `FindPureIntrinsic` for dead store elimination:

https://github.com/lifting-bits/remill/blob/e737070f4827368b49879695bf375d9885e9a2b7/lib/BC/IntrinsicTable.cpp#L58-L60

 and `readonly` and `argmemonly`.

Many thanks for any review.

### DIFF
--- a/lib/BC/IntrinsicTable.cpp
+++ b/lib/BC/IntrinsicTable.cpp
@@ -61,6 +61,12 @@ static llvm::Function *FindPureIntrinsic(llvm::Module *module,
   return function;
 }
 
+static llvm::Function *SetMemoryReadNone(llvm::Function *func) {
+  auto arg = remill::NthArgument(func, 0);
+  arg->addAttr(llvm::Attribute::ReadNone);
+  return func;
+}
+
 }  // namespace
 
 IntrinsicTable::IntrinsicTable(llvm::Module *module)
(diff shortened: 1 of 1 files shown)


pgoodman:
@tathanhdinh we've stopped supporting llvm versions less than 12, and we plan to move to llvm 13 soon. Do you rely on older versions?

tathanhdinh:
> 
> 
> @tathanhdinh we've stopped supporting llvm versions less than 12, and we plan to move to llvm 13 soon. Do you rely on older versions?

@pgoodman Ah, I don't know that. I did not follow the development of remill for so long (nearly 2 years!?). So it would compile just fine, I think. A minor remark, I still see:

https://github.com/lifting-bits/remill/blob/e737070f4827368b49879695bf375d9885e9a2b7/include/remill/BC/Version.h#L32-L34



> 
> 
> Also, for the read-modify-write intrinsics, e.g. cmpxchg, fetch_add, etc. perhaps a `__restrict` keyword could be used as well.

Sure, that's just because I still not sure what I do is correct or not (from your response in the issue).

Thank you for the feedback.

PR #564, opened 2021-11-30
Fix TryDecodeLoadStoreDualHalfSignedBIL

sschriner:


lib/Arch/AArch32/Decode.cpp
```diff
@@ -3631,7 +3635,7 @@ bool AArch32Arch::DecodeInstruction(uint64_t address,
 
   auto ret = decoder(inst, bits);
 
-  //  LOG(ERROR) << inst.Serialize();
+  LOG(ERROR) << inst.Serialize();
```
pgoodman, line 3675: Remove

pgoodman (changes requested):

lib/Arch/AArch32/Decode.cpp
```diff
@@ -2074,17 +2102,54 @@ static bool TryDecodeLoadStoreDualHalfSignedBReg(Instruction &inst,
...
+  //   If t == 15 || t2 == 15, then one of the following behaviors must occur:
+  //     - The instruction is undefined.
+  //     - The instruction executes as NOP.
+  //     - The store instruction performs the store using the specified
+  //      addressing mode but the value corresponding to R15 is unknown.
+  //
+  // Permitted UNPREDICTABLE behavior for STRDp only when rt is r15 only
+  if (enc.rt == kPCRegNum) {
+    // The instruction executes with the additional decode: t2 = t.
+    CHECK(!enc.o1 && enc.op2 == 0b11u)
```
pgoodman, line 2148: Turn this into an invalid instruction instead of a `CHECK` failure.

pgoodman (changes requested):

pgoodman (approved):

PR #570, opened 2021-12-22
Build via CMakePresets.json

artemdinaburg:
Add support for optionally building via CMakePresets.json.

* Update old build.sh script to fetch new cxx-common
* CMakePresets.json
* Script to build with CMakePresets.json
* Test the presets build in CI

### DIFF
--- a/CMakePresets.json
+++ b/CMakePresets.json
@@ -1,0 +1,172 @@
+{
+	"version": 3,
+	"cmakeMinimumRequired": {
+		"major": 3,
+		"minor": 19,
+		"patch": 0
+	},
+	"configurePresets": [
+		{
+			"name": "debug-flags",
+			"hidden": true,
+			"cacheVariables": {
+				"CMAKE_BUILD_TYPE": "Debug",
+				"CMAKE_C_FLAGS": "-fno-omit-frame-pointer -fno-optimize-sibling-calls -gfull -O0",
+				"CMAKE_CXX_FLAGS": "-fno-omit-frame-pointer -fno-optimize-sibling-calls -gfull -O0"
+			},
+      "environment": {
+        "VCPKG_SUFFIX": ""
+      }
+		},
+		{
+			"name": "release-flags",
+			"hidden": true,
+			"cacheVariables": {
+				"CMAKE_BUILD_TYPE": "Release"
+			},
+      "environment": {
+        "VCPKG_SUFFIX": "-rel"
+      }
+		},
+    {
+      "name": "arm64",
+      "hidden": true,
+      "environment": {
+        "VCPKG_ARCH": "arm64"
+      },
+      "architecture": {
+        "value": "arm64",
+        "strategy": "external"
+      }
+    },
+    {
+      "name": "x86_64",
+      "hidden": true,
+      "environment": {
+        "VCPKG_ARCH": "x64"
+      },
+      "architecture": {
+        "value": "x64",
+        "strategy": "external"
+      }
+    },
+		{
+			"name": "linux",
+			"hidden": true,
+			"cacheVariables": {
+				"CMAKE_C_COMPILER": "$env{VCPKG_ROOT}/installed/x64-linux$env{VCPKG_SUFFIX}/tools/llvm/clang",
+				"CMAKE_CXX_COMPILER": "$env{VCPKG_ROOT}/installed/x64-linux$env{VCPKG_SUFFIX}/tools/llvm/clang++"
+			},
+      "environment": {
+        "VCPKG_TARGET": "$env{VCPKG_ARCH}-linux"
+      },
+			"condition": {
+				"type": "equals",
+				"lhs": "Linux",
+				"rhs": "${hostSystemName}"
+			}
+		},
+		{
+			"name": "macos",
+			"hidden": true,
+			"cacheVariables": {
+				"CMAKE_C_COMPILER": "/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/clang",
+				"CMAKE_CXX_COMPILER": "/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/clang++"
+			},
+      "environment": {
+        "VCPKG_TARGET": "$env{VCPKG_ARCH}-osx"
+      },
+			"condition": {
+				"type": "equals",
+				"lhs": "Darwin",
+				"rhs": "${hostSystemName}"
+			}
+		},
+		{
+			"name": "vcpkg-debug",
+      "hidden": true,
+			"inherits": ["debug-flags", "linux", "macos"],
+			"binaryDir": "$env{INSTALL_DIR}/build/remill",
+			"generator": "Ninja",
+			"cacheVariables": {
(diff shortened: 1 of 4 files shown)


CMakePresets.json
```diff
@@ -0,0 +1,172 @@
...
+				"lhs": "Linux",
+				"rhs": "${hostSystemName}"
+			}
+		},
+		{
+			"name": "macos",
+			"hidden": true,
+			"cacheVariables": {
+				"CMAKE_C_COMPILER": "/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/clang",
+				"CMAKE_CXX_COMPILER": "/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/clang++"
```
ekilmer: Is this really necessary?

Shouldn't we just let CMake pick the compiler on its own?

I think the only issue we have is with GCC, but we should mention that in the README.

CMakePresets.json
```diff
@@ -0,0 +1,172 @@
...
+        "value": "x64",
+        "strategy": "external"
+      }
+    },
+		{
+			"name": "linux",
+			"hidden": true,
+			"cacheVariables": {
+				"CMAKE_C_COMPILER": "$env{VCPKG_ROOT}/installed/x64-linux$env{VCPKG_SUFFIX}/tools/llvm/clang",
+				"CMAKE_CXX_COMPILER": "$env{VCPKG_ROOT}/installed/x64-linux$env{VCPKG_SUFFIX}/tools/llvm/clang++"
```
ekilmer: I'm not sure we really want to rely on the compiler built by vcpkg for building the tool itself. Just use whatever CMake finds.

CMakePresets.json
```diff
@@ -0,0 +1,172 @@
...
+        {
+            "name": "debug-flags",
+            "hidden": true,
+            "cacheVariables": {
+                "CMAKE_BUILD_TYPE": "Debug",
+                "CMAKE_C_FLAGS": "-fno-omit-frame-pointer -fno-optimize-sibling-calls -gfull -O0",
+                "CMAKE_CXX_FLAGS": "-fno-omit-frame-pointer -fno-optimize-sibling-calls -gfull -O0"
+            },
+            "environment": {
+                "VCPKG_SUFFIX": ""
```
ekilmer, line 18: This means we would use the host system's default vcpkg triplet, which cxx-common does not distribute. Is that your intention?

Technically, we are still be able to build remill with debug symbols without requiring everything else to have debug symbols.

ekilmer (approved):
Looks like a good start!

CMakePresets.json
pgoodman: @ekilmer I would prefer to be able to specify something like `which clang` and `which clang++`, if possible, due to asan working with that on macOS, but not with vcpkg clang on macOS.

ekilmer: That should go in the `CMakeUserPresets.json` because your clang could have a different path than someone else.

pgoodman: Artem copied this absolute path from me because I was trying to diagnose a gRPC issue. Normally I do `which clang` and `which clang++`; another alternative for unix systems would be `which cc` and `which c++`. I guess if default CMake on macOS with vcpkg finds the system compiler then I'd be happy.

artemdinaburg: I have not tested a MacOS build but would be ok eliminating this; I was worried about it finding the "correct" clang to build & link; although I guess it could be fine to leave this out.

CMakePresets.json
artemdinaburg: So I personally really like picking the specific compiler from vcpkg! I found that for me, the default cmake always makes a *very poor* decision regarding picking a compiler, and then I have to remember to delete my build and re-set `CC` & `CXX`.

Is there a reason we would not want to default to the cxx-common compiler, given that we already ship a working compiler anyway? Afaik we can always override CMakePresets with explicit directives to CMake.

CMakePresets.json
artemdinaburg: I think I am missing something here and need a bit more of an explanation about the issue.

CMakePresets.json
pgoodman: How about `/usr/bin/cc` and `/usr/bin/c++`?

artemdinaburg: Also -- this just sets the default; it should still be possible to override it via `-DCMAKE_C_COMPILER=...`... need to see if just setting CC and CXX would also override it.

CMakePresets.json
ekilmer: cxx-common only builds the `x64-{linux,osx}-rel` triplet (value for `VCPKG_TARGET_TRIPLET`), so if a user uses this preset, they would need to have built cxx-common themselves in order to use the `x64-{linux,osx}` triplet that the empty `VCPKG_SUFFIX` will generate.

I worry that people will get very confused when vcpkg spits out an error that it can't find things when trying to build debug versions of remill.

CMakePresets.json
ekilmer: I know there have been some previous issues with the vcpkg-built compiler not being able to compile with an arbitrary sanitizer enabled, and I haven't tested it recently, but it sounds like a simple compile is currently working.

Yes, setting a different compiler for CMake on the CLI will work. 👍

pgoodman (approved):

CMakePresets.json
```diff
@@ -0,0 +1,172 @@
...
+            "cacheVariables": {
+                "VCPKG_TARGET_TRIPLET": "$env{VCPKG_TARGET}$env{VCPKG_SUFFIX}",
+                "VCPKG_ROOT": "$env{VCPKG_ROOT}",
+                "CMAKE_INSTALL_PREFIX": "$env{INSTALL_DIR}"
+            }
+        },
+        {
+            "name": "vcpkg-release",
+            "hidden": true,
+            "inherits": ["release-flags", "linux", "macos"],
```
ekilmer, line 100: Unfortunately, this sort of inheritance doesn't work on macOS with the conditional presets. It reads the `linux` preset first, finds that it's disabled and errors out. It works on macOS if you change the ordering like this though
```suggestion
            "inherits": ["release-flags", "macos", "linux"],
```
but then it won't work on Linux

ekilmer (changes requested):

CMakePresets.json
artemdinaburg: Thanks for looking into this! I was able to repro and working on a workaround where VCPKG_TARGET_TRIPLET is manually specified. I am still thinking about the best way to deal with VCPKG_TARGET_TRIPLET; probably will add some brains to the script to try and do the right thing.

artemdinaburg: Ok, I think I have a workaround that works for the common case (release vcpkg) but lets people handle a debug vcpkg build. Will push soon.

CMakePresets.json
ekilmer: Since we now take care of the triplet in the script, should we get rid of the `VCPKG_SUFFIX` in the presets file?

artemdinaburg: Yes 🙏

ekilmer (approved):
LGTM!

We now build twice in CI, but if you're okay with that, then I am too.

PR #579, opened 2022-02-28
Issue 578

pgoodman:


lib/BC/InstructionLifter.cpp
```diff
@@ -290,9 +291,25 @@ InstructionLifter::LoadRegAddress(llvm::BasicBlock *block,
     reg_ptr_it->second = reg_ptr;
     return reg_ptr;
 
+  // Try to find it as a global variable.
+  } else if (auto gvar = module->getGlobalVariable(reg_name)) {
+    return gvar;
+
+  // Invent a fake one and keep going.
   } else {
```
lkorenc, line 299: Drop `else` branch there is no reason to have it as previous branches always end with `return`.

lib/BC/InstructionLifter.cpp
```diff
@@ -290,9 +291,25 @@ InstructionLifter::LoadRegAddress(llvm::BasicBlock *block,
     reg_ptr_it->second = reg_ptr;
     return reg_ptr;
 
+  // Try to find it as a global variable.
+  } else if (auto gvar = module->getGlobalVariable(reg_name)) {
```
lkorenc, line 295: `if` is enough, there is `return` in branch before.

lib/BC/InstructionLifter.cpp
```diff
@@ -290,9 +291,25 @@ InstructionLifter::LoadRegAddress(llvm::BasicBlock *block,
...
   } else {
-    LOG(FATAL) << "Could not locate variable or register " << reg_name_;
-    return nullptr;
+    std::stringstream unk_var;
+    unk_var << "__remill_unknown_register_" << reg_name;
+    auto unk_var_name = unk_var.str();
+    auto var = module->getGlobalVariable(unk_var_name);
+    if (!var) {
+      LOG(ERROR)
+          << "Could not locate variable or register " << reg_name_;
```
lkorenc, line 306: `LOG(ERROR)` is problematic as user cannot do anything about it (method returns nothing about it) and user cannot configure the stream. We will need to fix them on whole project level anyway though, but if this is not something we want to propagate up to the user eventually, I would just remove it.

lkorenc: This can be simplified right?

```
if (auto var = module->getGlobalVariable(unk_var_name)) {
  return var;
}
return new llvm::GlobalVariable( ... )
```

lib/BC/InstructionLifter.cpp
lkorenc: Or maybe not as I am looking in the code above - simplify for the control flow to be easily trackable. There is no reason to have huge if-else trees when you will never continue after them, just `return`.

lib/BC/InstructionLifter.cpp
lkorenc: (See above)

lkorenc (changes requested):

lib/BC/InstructionLifter.cpp
lkorenc: Add a comment then that will give us a hint what should be done once `LOG` is refactored?

lib/BC/InstructionLifter.cpp
```diff
@@ -290,27 +291,28 @@ InstructionLifter::LoadRegAddress(llvm::BasicBlock *block,
...
-
-      var = new llvm::GlobalVariable(
-          *module, impl->word_type, false, llvm::GlobalValue::ExternalLinkage,
-          llvm::UndefValue::get(impl->word_type), unk_var_name);
-    }
+  std::stringstream unk_var;
+  unk_var << "__remill_unknown_register_" << reg_name;
+  auto unk_var_name = unk_var.str();
+  auto var = module->getGlobalVariable(unk_var_name);
+  if (var) {
```
lkorenc, line 306: `if( auto var = ... )` or is there any reason for it to have a it's own line (and therefore longer lifespan?)

PR #584, opened 2022-03-25
Filter for only loads that modify pc

2over12:
This adds filtering to prevent invalidating next_pc for stores that store the value of pc. Not sure if there are any implications im missing with these cases


sschriner (approved):

pgoodman:
This was merged, but were any semantics affected, or were they previously assuming no write-able operand?

PR #585, opened 2022-03-30
Ian/sleigh continuation

2over12:
Implements support for thumb via sleigh


lib/BC/SleighLifter.cpp
```diff
@@ -404,70 +419,30 @@ class SleighLifter::PcodeToLLVMEmitIntoBlock : public PcodeEmit {
...
-        break;
+        return this->LiftUnOpWithFloatIntrinsic(bldr, llvm::Intrinsic::round,
+                                                outvar, input_var);
       }
       case OpCode::CPUI_FLOAT_NAN: {
         auto nan_inval = this->LiftInParam(
             bldr, input_var, llvm::Type::getFloatTy(this->context));
         if (nan_inval.has_value()) {
-          // LLVM trunk has an `isnan`intrinsic but to support older versions, I think we need to do this.
+          // LLVM trunk has an `isnan` intrinsic but to support older versions, I think we need to do this.
```
pgoodman, line 445: Is it supported in LLVM 13 or 14? If 14 then we'll move to that reasonably soon, so file an issue to add its use in.

tetsuo-cpp: Oh, interesting. The intrinsic got reverted in https://github.com/llvm/llvm-project/commit/3f1f08f0ed6aa1ebd4678583cbbf5026ca5cbdf8.

I'll still make an issue to check whether it makes its way into 15.

tetsuo-cpp:
I think I've fixed one issue in [a9fc22a](https://github.com/lifting-bits/remill/pull/585/commits/a9fc22a872d2b91ed06a74e1f9f026835066ced2). Now we're generating this for the BLX instruction that you were using:
```
I0418 19:23:49.282938 294978944 Arch.cpp:80] Pcode: INT_OR(register,0x58,4):lr = (const,0x4,4)(const,0x1,4)
I0418 19:23:49.282960 294978944 Arch.cpp:80] Pcode: COPY(register,0xb0,1):ISAModeSwitch = (const,0x0,1)
I0418 19:23:49.282975 294978944 Arch.cpp:80] Pcode: COPY(register,0x69,1):TB = (register,0xb0,1):ISAModeSwitch
I0418 19:23:49.282985 294978944 Arch.cpp:80] Pcode: CALL(ram,0xfff0f2d4,4)

  store i32 5, i32* %LR, align 4
  %2 = alloca i8, align 1
  store i8 0, i8* %2, align 1
  %3 = load i8, i8* %2, align 1
  %4 = alloca i8, align 1
  store i8 %3, i8* %4, align 1
  store i32 -986412, i32* %PC, align 4
  %5 = load i32, i32* %PC, align 4
  store i32 %5, i32* %NEXT_PC, align 4
```
I think this is closer to what we want because now we're correctly generating the store to set `ISAModeSwitch` to 0. The part where we do `load` on `%2` seems wrong though since we're using `ISAModeSwitch` as a memory address for some reason. Still figuring out what to do about that.

pgoodman:
@tetsuo-cpp One thing I'd like is for ISA mode switches to be covered in `remill::Instruction::branch_taken_arch_name`. This probably means analyzing the p-code to detect the write to `ISAModeSwitch` and then recovering that into the `remill::Instruction`. My understanding is that the branch-not-taken / fall-through paths never change the architecture.

PR #588, opened 2022-04-11
Remove gflags

lkorenc:
Still links against `gflags` as I am not sure what relationship `gflags` and `glog` have (and we still use `glog`).


lib/BC/Util.cpp
```diff
@@ -381,10 +376,22 @@ llvm::GlobalVariable *FindGlobaVariable(llvm::Module *module,
...
   auto arch_name = GetArchName(arch->arch_name);
-  auto path = FindSemanticsBitcodeFile(arch_name);
-  DLOG(INFO) << "Loading " << arch_name << " semantics from file " << path;
-  auto module = LoadModuleFromFile(arch->context, path);
+  // If `sem_dirs` does not contain the dir, fallback to compiled in paths.
+  auto path = FindSemanticsBitcodeFile(arch_name, sem_dirs, true);
+  // TODO(lukas): We can propagate error up, but we should first check each callsite
+  //              properly checks for possible error (this could not return pointer
+  //              without value before).
+  if (!path)
```
pgoodman, line 390: `LOG_IF` or use `{` squigglies!

lib/BC/Util.cpp
```diff
@@ -559,59 +566,61 @@ namespace {
...
-  }
+  if (fallback_to_defaults)
+    return _FindSemanticsBitcodeFile(arch, DefaultSemanticsSearchPaths());
+
+  return {};
+}
 
-  LOG(FATAL) << "Cannot find path to " << arch << " semantics bitcode file.";
-  return "";
+std::optional<std::string> FindSemanticsBitcodeFile(std::string_view arch) {
```
pgoodman, line 619: Return `std::optional<std::filesystem::path>`?

lib/BC/Util.cpp
```diff
@@ -559,59 +566,61 @@ namespace {
...
-    pp << FLAGS_semantics_search_paths;
-    for (std::string sem_dir; std::getline(pp, sem_dir, ':');) {
-      std::stringstream ss;
-      ss << sem_dir << "/" << arch << ".bc";
-      if (auto sem_path = ss.str(); FileExists(sem_path)) {
-        return sem_path;
-      }
-    }
+std::optional<std::string> FindSemanticsBitcodeFile(std::string_view arch,
+                                     const std::vector< std::string > &dirs,
```
pgoodman, line 607: Take in a vector of `std::filesystsem::path`?

pgoodman (changes requested):

lib/BC/Util.cpp
lkorenc: Will use `{` as `LOG` macro will be removed next.

lib/BC/Util.cpp
lkorenc: To make transition easier I wanted to expose older type of API, but I guess I can change to return `path` as well in one swoop.

PR #591, opened 2022-05-03
Fix LLVM 14 related warnings.

xlauko:


include/remill/BC/Compat/PointerType.h
```diff
@@ -0,0 +1,30 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#pragma once
+
+#include "remill/BC/Version.h"
+
+namespace remill {
+
+inline static llvm::Type *getPointerElementType(llvm::PointerType *type) {
```
pgoodman, line 22: Rename to `PointerElementType`. Is a checked needed for `type->isOpaque()` and have that return `nullptr`?

pgoodman (changes requested):

include/remill/BC/Compat/PointerType.h
xlauko: Isn't it better to have the same name as LLVM does? So you don't have to always think whether it is with or without `get` prefix?

It is a precondition of both versions that the type is not opaque pointer. There is an assertion on that in both implementations.

pgoodman: If you're putting it in the `llvm` namespace, use `get`. In the remill namespace, drop `get` or change to `Get`. Is it an assertion only enabled in debug builds, or all builds? If the former, make it return `nullptr`.

xlauko: Fixed in c0b87c1ec6b4d2ff8d627a703ec663b4cf0f7ba8 and 9e87c962b3fccec6a4897341aa7c747e99ecf041

pgoodman (approved):

PR #596, opened 2022-05-31
Support opaque pointers

tetsuo-cpp:


lib/BC/Util.cpp
```diff
@@ -1465,6 +1464,9 @@ llvm::GlobalVariable *DeclareVarInModule(llvm::GlobalVariable *var,
     return llvm::dyn_cast<llvm::GlobalVariable>(moved_var);
   }
 
+  // NOTE(alex): This looks difficult to reproduce without querying `PointerElementType`.
```
tetsuo-cpp, line 1467: This one seems a bit tricky. The variables seem to be pointers but we want to move the underlying variable that the pointers is pointing at to the other module. That information just doesn't seem to be there with opaque pointers.

tetsuo-cpp: Seems like I can do `llvm::GlobalValue::getValueType` as a replacement.

lib/BC/InstructionLifter.cpp
```diff
@@ -328,7 +328,8 @@ llvm::Value *InstructionLifter::LoadRegValue(llvm::BasicBlock *block,
                                              std::string_view reg_name) const {
   auto ptr = LoadRegAddress(block, state_ptr, reg_name);
   CHECK_NOTNULL(ptr);
-  auto ptr_ty = ptr->getType()->getPointerElementType();
+  // NOTE(alex): This isn't right. Not sure how to solve this right now.
+  auto ptr_ty = impl->word_type;
```
pgoodman, line 332: Then change the API to fix this. Overall the `InstructionLifter` is totally evil and I regret it and so much more. We know that we can get the value type by `reg->type` if we can find `reg` via its name. However, we may also be dealing with "local var" regs, e.g. `BRANCH_TAKEN`. Those are `alloca`s for which the value type can be deduced. Finally, there is the case that t might find a global. I think instead of `auto ptr = LoadRegAddress(...)`, there should be something like `auto [ptr, val_type] = LoadRegAddress(...)`.

pgoodman (changes requested):

lib/BC/Util.cpp
```diff
@@ -2437,6 +2433,7 @@ llvm::Value *BuildPointerToOffset(llvm::IRBuilder<> &ir, llvm::Value *ptr,
     }
   }
 
+  // NOTE(alex): Same here
   const auto dest_elem_type = PointerElementType(dest_elem_ptr_type);
```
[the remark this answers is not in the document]
tetsuo-cpp: One problem area is here: https://github.com/lifting-bits/anvill/blob/master/lib/Lifters/DataLifter.cpp#L112
If we want to pass the underlying element types through to this function, we're going to need them here. However, one of the pointer types comes from a `VariableDecl` which I believe actually gets parsed from the JSON format. So we can't really embed the element type into the `VariableDecl` because we purposely removed that information from the JSON spec format.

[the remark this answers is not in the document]
pgoodman: I'd like to say that the anvill side of things should stop doing that, whatever "that is." And by this, I mean: I think we should stop lifting data and global vars in anvill, and try make that function return `nullptr`, or pass in a `nullptr` type or something.

But, that is beyond the scope of this change, so I would err on the side of: maybe the subsequent code should stop trying to be clever, and instead just do byte-granularity indexing, instead of pointer element type size-granularity indexing.

include/remill/Arch/Arch.h
```diff
@@ -125,7 +125,7 @@ struct Register {
   // The directly enclosed registers.
   std::vector<const Register *> children;
 
-  void CompteGEPAccessors(const llvm::DataLayout &dl, llvm::Type *state_type);
+  void ComputeGEPAccessors(const llvm::DataLayout &dl, llvm::Type *state_type);
```
pgoodman, line 128: `llvm::StructType *state_type`

lib/Arch/Arch.cpp
```diff
@@ -237,7 +237,7 @@ llvm::StructType *Arch::StateStructType(void) const {
 llvm::PointerType *Arch::StatePointerType(void) const {
```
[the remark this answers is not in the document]
pgoodman: Whoops, has to be in the header.

pgoodman (changes requested):

lib/BC/Util.cpp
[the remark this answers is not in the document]
tetsuo-cpp: I'm interpreting that to mean that I should change `BuildPointerToOffset` index at the byte-level instead of doing clever things with the element size as opposed to writing something custom to replace that invocation of `BuildPointerToOffset`.

I have some code to do this and am currently debugging it.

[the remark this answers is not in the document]
tetsuo-cpp: [080969e](https://github.com/lifting-bits/remill/pull/596/commits/080969e0b00fe3ea80d04792b26d668bf602a352)

include/remill/Arch/Arch.h
tetsuo-cpp: [e266fa9](https://github.com/lifting-bits/remill/pull/596/commits/e266fa93ee66e7a4605a82eafdbe7efe2e28e58a)

lib/Arch/Arch.cpp
```diff
@@ -539,22 +533,22 @@ FinishAddressOf(llvm::IRBuilder<> &ir, const llvm::DataLayout &dl,
...
     return ir.CreateBitCast(gep, goal_ptr_type);
   }
 }
 
 }  // namespace
 
-void Register::CompteGEPAccessors(const llvm::DataLayout &dl,
-                                  llvm::Type *state_type) {
+void Register::ComputeGEPAccessors(const llvm::DataLayout &dl,
+                                   llvm::Type *state_type) {
```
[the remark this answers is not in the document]
tetsuo-cpp: [e266fa9](https://github.com/lifting-bits/remill/pull/596/commits/e266fa93ee66e7a4605a82eafdbe7efe2e28e58a)

lib/Arch/Arch.cpp
```diff
@@ -482,27 +481,22 @@ static uint64_t TotalOffset(const llvm::DataLayout &dl, llvm::Value *base,
 
 static llvm::Value *
 FinishAddressOf(llvm::IRBuilder<> &ir, const llvm::DataLayout &dl,
-                llvm::Type *state_ptr_type, size_t state_size,
+                llvm::Type *state_type, size_t state_size,
```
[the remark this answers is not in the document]
tetsuo-cpp: [e266fa9](https://github.com/lifting-bits/remill/pull/596/commits/e266fa93ee66e7a4605a82eafdbe7efe2e28e58a)

lib/Arch/Arch.cpp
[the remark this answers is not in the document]
tetsuo-cpp: [4cc8253](https://github.com/lifting-bits/remill/pull/596/commits/4cc82534b6006ec625eda88a340ecd9058b65635)

lib/Arch/Arch.cpp
```diff
@@ -803,10 +795,7 @@ void Arch::InitFromSemanticsModule(llvm::Module *module) const {
   const auto &dl = module->getDataLayout();
   const auto basic_block = module->getFunction("__remill_jump");
   CHECK_NOTNULL(basic_block);
-  const auto state_ptr_type =
-      NthArgument(basic_block, kStatePointerArgNum)->getType();
-  const auto state_type =
-      llvm::dyn_cast<llvm::StructType>(state_ptr_type->getPointerElementType());
+  const auto state_type = llvm::StructType::getTypeByName(module->getContext(), "struct.State");
```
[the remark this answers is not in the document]
tetsuo-cpp: I'm trying to figure out how this should look. In `X86/Runtime/BasicBlock.cpp` I see some declarations for registers marked with `extern` but I don't see where they're instantiated. If I do `module->getGlobalVariable(...)` for those register names, they don't seem to be accessible.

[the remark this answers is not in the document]
pgoodman: So you'd declare these global variables, mark them as used, then look at `__remill_intrinsics` and try to mark them used there too. The purpose is just that these variables exist as kind of "anchors" providing us a dependable way of accessing these structure types (as the global variable value type). These variables will otherwise never be defined/used, only declared.

Those variables that you're seeing are debugging registers that are global, whereas a `State` is "thread local." It's up to a runtime (code that links against lifted bitcode) to define those register global variables.

lib/BC/Util.cpp
```diff
@@ -208,38 +207,41 @@ llvm::CallInst *AddTerminatingTailCall(llvm::BasicBlock *source_block,
...
 }
 
 // Find a local variable defined in the entry block of the function. We use
 // this to find register variables.
-llvm::Value *FindVarInFunction(llvm::Function *function, std::string_view name_,
-                               bool allow_failure) {
+std::pair<llvm::Value *, llvm::Type *>
+FindVarInFunction(llvm::Function *function, std::string_view name_,
+                  bool allow_failure) {
+  // TODO(alex): Figure out how to to get types for the two cases below.
```
tetsuo-cpp, line 221: > However, we may also be dealing with "local var" regs, e.g. BRANCH_TAKEN. Those are allocas for which the value type can be deduced

@pgoodman This part wasn't clear to me. What did you mean by deducing the type from allocas?

lib/BC/InstructionLifter.cpp
tetsuo-cpp: [7fdcb51](https://github.com/lifting-bits/remill/pull/596/commits/7fdcb510f2c64d4aa70bc21df414538bb1b16159)

lib/Arch/Arch.cpp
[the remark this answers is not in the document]
tetsuo-cpp: [f29e920](https://github.com/lifting-bits/remill/pull/596/commits/f29e920a128d0cf2d4386b5e73de20bbd119c4cd)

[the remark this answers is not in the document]
tetsuo-cpp: [20bc423](https://github.com/lifting-bits/remill/pull/596/commits/20bc423b7a170443ef2e85586a23c44524b20caa)

lib/Arch/SPARC64/Arch.cpp
```diff
@@ -376,7 +376,7 @@ void SPARC64Arch::FinishLiftedFunctionInitialization(
   //            a structure type, so we can check that.
   const auto prev_window_link = this->RegisterByName("PREV_WINDOW_LINK");
   CHECK(prev_window_link->type->isPointerTy());
-  const auto window_type = prev_window_link->type->getPointerElementType();
+  const auto window_type = llvm::StructType::getTypeByName(context, "RegisterWindow");
```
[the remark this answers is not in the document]
tetsuo-cpp: [67c5fd0](https://github.com/lifting-bits/remill/pull/596/commits/67c5fd064f757fd15329f2e9094c96e3a4ce0d40)

lib/Arch/SPARC32/Arch.cpp
```diff
@@ -404,7 +404,7 @@ void SPARC32Arch::FinishLiftedFunctionInitialization(
   //            a structure type, so we can check that.
   const auto prev_window_link = RegisterByName("PREV_WINDOW_LINK");
   CHECK(prev_window_link->type->isPointerTy());
-  const auto window_type = prev_window_link->type->getPointerElementType();
+  const auto window_type = llvm::StructType::getTypeByName(context, "RegisterWindow");
```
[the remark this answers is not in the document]
tetsuo-cpp: [67c5fd0](https://github.com/lifting-bits/remill/pull/596/commits/67c5fd064f757fd15329f2e9094c96e3a4ce0d40)

lib/BC/Util.cpp
pgoodman: If you cast the thing to an `AllocaInst *`, and it works, then from the alloca inst you can get the `llvm::Type *` for the allocated value.

tetsuo-cpp: What if it isn't an `AllocaInst` though? Just skip it? That's what I'm intending to do but it seems like it will find less variables this way.

tetsuo-cpp: [669166d](https://github.com/lifting-bits/remill/pull/596/commits/669166d5725e51dc045efbe6ae05fd6d61262175)

From my test program, I saw that it can either be an alloca or a GEP so I handled those two cases. I also didn't see the code ever hit the function arg branch so I just removed it entirely.

lib/Arch/X86/Runtime/BasicBlock.cpp
```diff
@@ -44,6 +48,12 @@ extern CR3Reg CR3;
 extern CR4Reg CR4;
 extern CR8Reg CR8;
 
+extern State __remill_state;
+
 #pragma clang diagnostic pop
 
+[[gnu::used]] void __remill_state_used(void) {
```
tetsuo-cpp, line 55: This needs to get added for each arch (in addition to the register window struct where relevant).

pgoodman: Ooooh sorry it would be easier to modify `Intrinsics.cpp` and `Instrinsics.h`!! Then you get it done for all of the architectures at once. I would also update the `IntrinsicTable` class to go and be the thing responsible for getting the state structure and register window structure (or a nullptr).

lib/BC/Util.cpp
pgoodman: 👍

lib/Arch/X86/Runtime/BasicBlock.cpp
tetsuo-cpp: I was thinking of that but I don't think we could put `__remill_register_window` in there since not all the runtimes have that struct.

tetsuo-cpp: I guess I'll put `__remill_state` in `Intrinsics.cpp` once for all architectures. Then the `__remill_register_window` can go in each arch's `BasicBlock.cpp` where applicable.

tetsuo-cpp:
@pgoodman 
Ah the tests are a bit of a headache 🤦‍♂️ . So part of our building process for the test dependencies involves invoking Clang on some test bitcode that came from Remill (I think) and generating some textual assembly from it. However, what I've realised is that opaque pointers are supported in LLVM 14 but not in Clang 14 if that makes sense. So when the build invokes the Clang that came with the LLVM 14 toolchain, it barfs on our bitcode because it contains typeless pointers.

lib/Arch/X86/Runtime/BasicBlock.cpp
tetsuo-cpp: Oh actually, I can check `REMILL_ARCH` and only add `RegisterWindow` when it's a SPARC arch.

tetsuo-cpp: [385150a](https://github.com/lifting-bits/remill/pull/596/commits/385150a857ae069698a62a683f1a0579912cf8fb)

pgoodman:
@tetsuo-cpp We'll leave fixing the tests to the next PR. I think you could probably use the vcpkg `llc`, perhaps.

pgoodman (approved):

tetsuo-cpp:
Cool! As discussed, I'm going to get this in + the Anvill change and deal with the test fallout in a follow up fix.

To anyone curious, the issue is that our testing involves feeding some Remill-generated bitcode into Clang and trying to lower into some textual assembly. Unfortunately, Clang 14 doesn't support opaque pointers yet (15 will use opaque pointers by default and have a `-no-opaque-pointers` flag) and currently barfs on Remill's bitcode. I think the fix will involve using `llc` to convert the bitcode to assembly since there's no reason we need to use Clang.

PR #605, opened 2022-06-27
Pass opaque pointer flag to Clang so it can handle Remill's IR

### LINKED ISSUE #604: tests: Clang complains about bitcode with typeless pointers generated by Remill
As of #596, CI is broken. The issue is that part of our testing involves feeding some Remill-generated IR into Clang. Unfortunately, Clang 14 doesn't support opaque pointers yet (15 will use opaque pointers by default and have a `-no-opaque-pointers` flag) and currently errors on Remill's bitcode. I think the fix will involve using `llc` to convert the bitcode to assembly since there's no reason we need to use Clang.

tetsuo-cpp:
Closes #604

### DIFF
--- a/lib/BC/Optimizer.cpp
+++ b/lib/BC/Optimizer.cpp
@@ -56,7 +56,10 @@ void OptimizeModule(const remill::Arch *arch, llvm::Module *module,
   TLI->disableAllFunctions();  // `-fno-builtin`.
 
   llvm::PassManagerBuilder builder;
-  builder.OptLevel = 3;
+  // TODO(alex): Some of the optimization passes that the builder adds still rely on typed pointers
+  // so we cannot use them. We should switch to using the new pass manager and choose which passes
+  // we want.
+  builder.OptLevel = 0;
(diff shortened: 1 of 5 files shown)


lib/BC/Optimizer.cpp
```diff
@@ -56,7 +56,10 @@ void OptimizeModule(const remill::Arch *arch, llvm::Module *module,
   TLI->disableAllFunctions();  // `-fno-builtin`.
 
   llvm::PassManagerBuilder builder;
-  builder.OptLevel = 3;
+  // TODO(alex): Some of the optimization passes that the builder adds still rely on typed pointers
+  // so we cannot use them. We should switch to using the new pass manager and choose which passes
+  // we want.
+  builder.OptLevel = 0;
```
tetsuo-cpp, line 62: This isn't ideal but it looks like Anvill doesn't actually use this. It's just `remill-lift` that calls into it so I'm assuming the impact of this is minimal.

pgoodman (approved):

PR #606, opened 2022-06-28
Remove compatibility for LLVM < 14

### LINKED ISSUE #603: Remove obsolete LLVM version checks
As of the opaque pointers PR, we'll only be supporting LLVM 14+. This means that a lot of the preprocessor checks that check `LLVM_VERSION` are no longer relevant and should be removed.

tetsuo-cpp:
Closes #603


include/remill/BC/Compat/CTypes.h
```diff
@@ -1,30 +0,0 @@
...
- */
-
-#pragma once
-
-#if __has_include(<llvm-c/Types.h>)
-#  include <llvm-c/Types.h>
-
-namespace llvm {
-
-// TODO(pag): This is a rather ugly hack; had some issues with anvill not
```
tetsuo-cpp, line 24: I don't see this file included anywhere in Anvill so I'm assuming this doesn't apply anymore.

pgoodman (approved):

PR #607, opened 2022-06-28
Ian/sleigh support rebased

2over12:
This pull request implements lifting to llvm via sleigh semantics. The PR adds the lifting-bits sleigh repo as a dependency.


bin/differential_tester_x86/LiftAndCompare.cpp
```diff
@@ -0,0 +1,811 @@
...
+    auto second_handler = std::make_unique<MemoryHandler>(
+        this->endian, mem_handler->GetUninitializedReads());
+    ExecuteLiftedFunction(f2, insn_length, func2_state, second_handler.get());
+
+
+    auto memory_state_eq =
+        mem_handler->GetMemory() == second_handler->GetMemory();
+
+    if (!memory_state_eq) {
+      LOG(ERROR) << "Memory state differs";
```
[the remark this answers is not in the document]
2over12: These differences are what we are trying to collect with this tester. Each error represents a problem in either or semantics or sleigh lifting

bin/differential_tester_x86/LiftAndCompare.cpp
```diff
@@ -0,0 +1,811 @@
...
+
+  const remill::Arch::ArchPtr &GetArch() {
+    return this->arch;
+  }
+};
+
+static constexpr auto kFlagIntrinsicPrefix = "__remill_flag_computation";
+static constexpr auto kCompareFlagIntrinsicPrefix = "__remill_compare";
+
+bool flag_computation_stub(bool res, ...) {
```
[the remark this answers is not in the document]
lkorenc: Well, just make this comment a comment in a code then.

lib/BC/SleighLifter.cpp
```diff
@@ -0,0 +1,1292 @@
...
+#include <remill/BC/ABI.h>
+#include <remill/BC/IntrinsicTable.h>
+#include <remill/BC/SleighLifter.h>
+#include <remill/BC/Util.h>
+
+#include <cassert>
+#include <unordered_map>
+#include <unordered_set>
+
+#include "../Arch/Sleigh/Arch.h"
```
lkorenc, line 27: Use absolute path?

lib/BC/SleighLifter.cpp
```diff
@@ -0,0 +1,1292 @@
...
+    if (op_func) {
+      // TODO(alex): I think we need some helper here to achieve something similar to what `llvm::IntegerType::get`
+      // gives us, except for floating point types.
+      //
+      // So we need to check the size of the node and return either a 32-bit float, brain float, double, etc.
+      auto lifted_lhs =
+          this->LiftInParam(bldr, lhs, llvm::Type::getFloatTy(this->context));
+      auto lifted_rhs =
+          this->LiftInParam(bldr, lhs, llvm::Type::getFloatTy(this->context));
+      if (lifted_lhs.has_value() && lifted_rhs.has_value()) {
```
lkorenc, line 842: Copy & paste.

lib/BC/SleighLifter.cpp
```diff
@@ -0,0 +1,1292 @@
...
+      return std::make_pair(LiftStatus::kLiftedUnsupportedInstruction, false);
+    }
+
+    return std::make_pair(LiftStatus::kLiftedUnsupportedInstruction, false);
+  }
+
+  LiftStatus LiftTernOp(llvm::IRBuilder<> &bldr, OpCode opc,
+                        VarnodeData *outvar, VarnodeData param0,
+                        VarnodeData param1, VarnodeData param2) {
+    switch (opc) {
```
lkorenc, line 956: `param0-2` is not very helpful naming - it is a ternary op, those operands have meaning - `cond, true_val, false_val` for example.

lib/BC/SleighLifter.cpp
```diff
@@ -0,0 +1,1292 @@
...
+      case OpCode::CPUI_STORE: {
+        auto addr_operand = param1;
+        auto lifted_addr_offset = this->LiftInParam(
+            bldr, addr_operand, this->insn_lifter_parent.GetWordType());
+
+        if (lifted_addr_offset) {
```
lkorenc, line 962: early return.

lib/BC/SleighLifter.cpp
```diff
@@ -0,0 +1,1292 @@
...
+      this->UpdateStatus(
+          this->LiftVariadicOp(bldr, current_target.opc, current_target.outvar,
+                               current_target.vars, current_target.isize),
+          current_target.opc);
+      return false;
+    }
+
+    if (current_target.opc == OpCode::CPUI_CALLOTHER) {
+      if (current_target.isize == 3 &&
+          current_target.vars[0].offset < this->user_op_names.size() &&
```
lkorenc, line 1055: when you are using dot two times to ked to some magic attribute it is a signal abstractions could be leaky.

lib/BC/SleighLifter.cpp
```diff
@@ -0,0 +1,1292 @@
...
+  }
+
+  LiftStatus GetStatus() {
+    return this->status;
+  }
+};  // namespace remill
+
+std::unordered_set<OpCode>
+    SleighLifter::PcodeToLLVMEmitIntoBlock::INTEGER_COMP_OPS = {
+        CPUI_INT_EQUAL,   CPUI_INT_NOTEQUAL,  CPUI_INT_LESS,
```
lkorenc, line 1109: Define it where declared?

lib/BC/SleighLifter.cpp
```diff
@@ -0,0 +1,1292 @@
...
+         [](llvm::Value *lhs, llvm::Value *rhs, llvm::IRBuilder<> &bldr) {
+           auto new_rhs = rhs;
+           if (lhs->getType() != rhs->getType()) {
+             new_rhs = bldr.CreateZExt(rhs, lhs->getType());
+           }
+           return bldr.CreateShl(lhs, new_rhs);
+         }},
+        {OpCode::CPUI_INT_RIGHT,
+         [](llvm::Value *lhs, llvm::Value *rhs, llvm::IRBuilder<> &bldr) {
+           auto new_rhs = rhs;
```
lkorenc, line 1137: Why is this done? Pointer is taken by value?

lib/BC/SleighLifter.cpp
```diff
@@ -0,0 +1,1292 @@
...
+  }
+};  // namespace remill
+
+std::unordered_set<OpCode>
+    SleighLifter::PcodeToLLVMEmitIntoBlock::INTEGER_COMP_OPS = {
+        CPUI_INT_EQUAL,   CPUI_INT_NOTEQUAL,  CPUI_INT_LESS,
+        CPUI_INT_SLESS,   CPUI_INT_LESSEQUAL, CPUI_INT_SLESSEQUAL,
+        CPUI_INT_SBORROW, CPUI_INT_SCARRY,    CPUI_INT_CARRY};
+
+std::map<OpCode, SleighLifter::PcodeToLLVMEmitIntoBlock::BinaryOperator>
```
lkorenc, line 1113: What about
```
struct MyBuilder
{
   using builder_t = llvm::IRBuilder;
   builder_t &irb;

   auto translate(auto opcode, const std::vector< llvm::Value * > &ops)
   {
      switch(opcode)
      {
        case CPUI_INT_AND: return simple_op< 2 >(&builder_t::CreateAnd, ops);
      }
   }
};

auto translate(opcode, bldr,  ops) { return MyBuilder(bldr).translate( ... ); } // or some other wrapper fn.
```
I feel this has less bolierplate and since it is a type, it can be easily swapped if someone desires so.

lib/BC/SleighLifter.cpp
```diff
@@ -0,0 +1,1292 @@
...
+           return bldr.CreateZExt(bldr.CreateICmpULE(lhs, rhs),
+                                  llvm::IntegerType::get(bldr.getContext(), 8));
+         }},
+        {OpCode::CPUI_INT_CARRY,
+         [](llvm::Value *lhs, llvm::Value *rhs, llvm::IRBuilder<> &bldr) {
+           llvm::Type *overloaded_types[1] = {lhs->getType()};
+           llvm::Function *uadd_intrinsic = llvm::Intrinsic::getDeclaration(
+               bldr.GetInsertBlock()->getModule(),
+               llvm::Intrinsic::uadd_with_overflow, overloaded_types);
+           llvm::Value *uadd_args[] = {lhs, rhs};
```
lkorenc, line 1213: `std::array`

lib/BC/SleighLifter.cpp
```diff
@@ -0,0 +1,1292 @@
...
+           return bldr.CreateExtractValue(sadd_val, {1});
+         }},
+        {OpCode::CPUI_INT_SBORROW,
+         [](llvm::Value *lhs, llvm::Value *rhs, llvm::IRBuilder<> &bldr) {
+           llvm::Type *overloaded_types[1] = {lhs->getType()};
+           llvm::Function *ssub_intrinsic = llvm::Intrinsic::getDeclaration(
+               bldr.GetInsertBlock()->getModule(),
+               llvm::Intrinsic::ssub_with_overflow, overloaded_types);
+           llvm::Value *ssub_args[] = {lhs, rhs};
+           llvm::Value *ssub_val = bldr.CreateCall(ssub_intrinsic, ssub_args);
```
lkorenc, line 1236: copy & paste.

lib/BC/SleighLifter.cpp
```diff
@@ -0,0 +1,1292 @@
...
+  arch_->InitializeSleighContext(*sleigh_context);
+}
+
+LiftStatus
+SleighLifter::LiftIntoBlock(Instruction &inst, llvm::BasicBlock *block,
+                            llvm::Value *state_ptr, bool is_delayed) {
+
+  if (!inst.IsValid()) {
+    LOG(ERROR) << "Invalid function" << inst.Serialize();
+    inst.operands.clear();
```
lkorenc, line 1256: We are calling `LiftIntoBlock` why does the instruction gets stealth erased?

lib/BC/SleighLifter.cpp
```diff
@@ -0,0 +1,1292 @@
...
+  ir.CreateStore(curr_eip, pc_ref);
+
+
+  SleighLifter::PcodeToLLVMEmitIntoBlock lifter(
+      block, state_ptr, inst, *this, this->sleigh_context->getUserOpNames());
+  auto res = sleigh_context->oneInstruction(inst.pc, lifter, inst.bytes);
+
+  lifter.finalize();
+
+  (void) res;
```
lkorenc, line 1279: `std::ignore = ...` is the way to do it, although it is a red flag unless good reason (there is no comment right now).

bin/differential_tester_x86/LiftAndCompare.cpp
```diff
@@ -0,0 +1,811 @@
...
+  static DifferentialModuleBuilder
+  Create(remill::OSName os_name_1, remill::ArchName arch_name_1,
+         remill::OSName os_name_2, remill::ArchName arch_name_2) {
+    // it is expected that compatible arches share a semantics module.
+    std::unique_ptr<llvm::LLVMContext> context =
+        std::make_unique<llvm::LLVMContext>();
+    auto tmp_arch = remill::Arch::Build(context.get(), os_name_1, arch_name_1);
+    auto semantics_module = remill::LoadArchSemantics(tmp_arch.get());
+    tmp_arch->PrepareModule(semantics_module);
+    auto l1 = LiftingTester(semantics_module.get(), os_name_1, arch_name_1);
```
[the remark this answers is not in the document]
2over12: Not possible because the semantics module is moved in the ctor

bin/differential_tester_x86/LiftAndCompare.cpp
```diff
@@ -0,0 +1,811 @@
...
+class MemoryHandler {
+ private:
+  std::unordered_map<uint64_t, uint8_t> uninitialized_reads;
+  std::unordered_map<uint64_t, uint8_t> state;
+
+  random_bytes_engine rbe;
+  llvm::support::endianness endian;
+
+ public:
+  MemoryHandler(llvm::support::endianness endian_) : endian(endian_) {}
```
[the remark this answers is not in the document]
2over12: it's just an enum of big little native

bin/differential_tester_x86/LiftAndCompare.cpp
```diff
@@ -0,0 +1,811 @@
...
+
+    auto tgt_mod = llvm::CloneModule(*func->getParent());
+    tgt_mod->setTargetTriple("");
+    tgt_mod->setDataLayout(llvm::DataLayout(""));
+    llvm::InitializeNativeTarget();
+    llvm::InitializeNativeTargetAsmParser();
+    llvm::InitializeNativeTargetAsmPrinter();
+    llvm::InitializeAllTargetMCs();
+
+    auto res = remill::VerifyModuleMsg(tgt_mod.get());
```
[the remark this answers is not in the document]
2over12: JIT initialization? doesnt look like it

bin/differential_tester_x86/LiftAndCompare.cpp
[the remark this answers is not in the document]
lkorenc: Would leave a comment about that in the code as it really looks half like debug messages/"normal" logs.

lib/BC/SleighLifter.cpp
```diff
@@ -0,0 +1,1292 @@
...
+    return LiftStatus::kLiftedUnsupportedInstruction;
+  }
+  using BinaryOperator = std::function<llvm::Value *(
+      llvm::Value *, llvm::Value *, llvm::IRBuilder<> &)>;
+  static std::map<OpCode, BinaryOperator> INTEGER_BINARY_OPS;
+  static std::unordered_set<OpCode> INTEGER_COMP_OPS;
+
+
+  std::pair<LiftStatus, bool>
+  LiftIntegerBinOp(llvm::IRBuilder<> &bldr, OpCode opc, VarnodeData *outvar,
```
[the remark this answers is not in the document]
pgoodman: I believe it will lead to unintuitive clashes.

lib/BC/SleighLifter.cpp
```diff
@@ -0,0 +1,1292 @@
...
+              *trunc_inval,
+              llvm::IntegerType::get(this->context, outvar->size * 8));
+          return this->LiftStoreIntoOutParam(bldr, converted, outvar);
+        }
+        break;
+      }
+      case OpCode::CPUI_POPCOUNT: {
+        auto ctpop_inval = this->LiftIntegerInParam(bldr, input_var);
+        if (ctpop_inval.has_value()) {
+          llvm::Type *overloaded_types[1] = {(*ctpop_inval)->getType()};
```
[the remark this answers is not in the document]
pgoodman: Does `llvm::ArrayRef` take a `std::array`? If not, then that is one instance where a C array may be desirable.

lib/BC/SleighLifter.cpp
2over12: I think this is a bad function name. It is just handling operands that have three operands, it's not a ternary conditional

lib/BC/SleighLifter.cpp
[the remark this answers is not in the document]
lkorenc: Even if not, you can always ask `std::array` to give you the raw pointer iirc.

lib/BC/SleighLifter.cpp
lkorenc: Then take an `std::tuple` or `std::array`? If they have no special meaning by themselves, why do they even need names?

bin/differential_tester_x86/LiftAndCompare.cpp
```diff
@@ -0,0 +1,541 @@
...
+DEFINE_bool(stop_on_fail, false, "Stop on first failure");
+
+
+class DiffModule {
+ private:
+  std::unique_ptr<llvm::Module> mod;
+  llvm::Function *f1;
+  llvm::Function *f2;
+  std::string f1_insn_name;
+  std::string f2_insn_name;
```
lkorenc, line 48: `std::vector` of tuples? Having 4 attrs like this is a lot of copy and paste - `GetF1` and `GetF2` can be easily unified as `GetF< N >` or `GetF( n )` - same for names.

bin/differential_tester_x86/LiftAndCompare.cpp
```diff
@@ -0,0 +1,541 @@
...
+
+
+class MappTypeRemapper : public llvm::ValueMapTypeRemapper {
+ private:
+  const remill::TypeMap &tmap;
+
+ public:
+  MappTypeRemapper(const remill::TypeMap &tmap_) : tmap(tmap_) {}
+
+  virtual llvm::Type *remapType(llvm::Type *SrcTy) override {
```
lkorenc, line 89: `SrcTy` -> `srcTy`?

bin/differential_tester_x86/LiftAndCompare.cpp
```diff
@@ -0,0 +1,541 @@
...
+    }
+
+    auto f1_and_name = *maybe_f1;
+    auto f2_and_name = *maybe_f2;
+
+    auto f1 = f1_and_name.first;
+    auto f2 = f2_and_name.first;
+    llvm::verifyFunction(*f1, &llvm::errs());
+
+    llvm::verifyFunction(*f2, &llvm::errs());
```
lkorenc, line 165: Copy & paster

bin/differential_tester_x86/LiftAndCompare.cpp
```diff
@@ -0,0 +1,541 @@
...
+
+
+    auto tst = f1->getParent();
+
+    for (const auto &f : tst->getFunctionList()) {
+      if (llvm::verifyFunction(f, &llvm::errs())) {
+
+        f.dump();
```
lkorenc, line 173: Is `dump()` still supported? I recall it being removed in favor of `print`.

bin/differential_tester_x86/LiftAndCompare.cpp
```diff
@@ -0,0 +1,541 @@
...
+        LOG(INFO) << "Num basic block: " << f.getBasicBlockList().size();
+        LOG(FATAL) << "Error in " << f.getName().str();
+      }
+    }
+
+    assert(remill::VerifyModule(tst));
```
lkorenc, line 179: `CHECK`?

bin/differential_tester_x86/LiftAndCompare.cpp
```diff
@@ -0,0 +1,541 @@
...
+
+    auto cloned = llvm::CloneModule(*tst);
+
+    auto maybe_message = remill::VerifyModuleMsg(cloned.get());
+    if (maybe_message.has_value()) {
```
lkorenc, line 184: `if (auto maybe_message = ...)`

bin/differential_tester_x86/LiftAndCompare.cpp
```diff
@@ -0,0 +1,541 @@
...
+      cloned->getFunction(f1->getName())->dump();
```
lkorenc, line 185:
```
for (const auto &name : { f1->getName(), f2->getName(), sleigh_name })
  if (auto fn = cloned->getFunction(name))
    fn->dump()
```
Something like this maybe?

bin/differential_tester_x86/LiftAndCompare.cpp
```diff
@@ -0,0 +1,541 @@
...
+      cloned->getFunction(f2->getName())->dump();
+      auto insn_func =
+          cloned->getFunction("sleigh_remill_instruction_function");
```
lkorenc, line 188: Magic constant - pull out as attribute?

bin/differential_tester_x86/LiftAndCompare.cpp
```diff
@@ -0,0 +1,541 @@
...
+      if (insn_func) {
+        insn_func->dump();
```
lkorenc, line 190: See above.

bin/differential_tester_x86/LiftAndCompare.cpp
```diff
@@ -0,0 +1,541 @@
...
+    func1_state->aflag.pf = test_runner::random_boolean_flag(this->rbe);
+    func1_state->aflag.sf = test_runner::random_boolean_flag(this->rbe);
+    func1_state->aflag.zf = test_runner::random_boolean_flag(this->rbe);
+
+    if (isel_name.rfind("REP_") != std::string::npos) {
+      LOG(INFO) << "setting ecx to 1";
+      func1_state->gpr.rcx.dword = 1;
+    }
+
+    auto func2_state = (X86State *) alloca(sizeof(X86State));
```
lkorenc, line 327: I am confused by this -> why don't I just allocate it as "normal" variable?

bin/differential_tester_x86/LiftAndCompare.cpp
```diff
@@ -0,0 +1,541 @@
...
+
+    auto init_state = this->DumpState(func1_state);
+
+    std::memcpy(func2_state, func1_state, sizeof(X86State));
+
+    assert(std::memcmp(func1_state, func2_state, sizeof(X86State)) == 0);
+
+    auto mem_handler =
+        std::make_unique<test_runner::MemoryHandler>(this->endian);
+    std::function<uint64_t(X86State * st)> pc_fetch = [](X86State *st) {
```
lkorenc, line 337: `auto pc_fetch = []`

bin/differential_tester_x86/LiftAndCompare.cpp
```diff
@@ -0,0 +1,541 @@
...
+  ComparisonRunner comp_runner(end);
+
+  if (FLAGS_should_dump_functions) {
+    LOG(INFO) << remill::LLVMThingToString(diff_mod->GetF1());
+    LOG(INFO) << remill::LLVMThingToString(diff_mod->GetF2());
+  }
+
+  for (uint64_t i = 0; i < FLAGS_num_iterations; i++) {
+    auto tc_result = comp_runner.SingleCmpRun(
+        tc.bytes.size(), diff_mod->GetF1(), diff_mod->GetF2(), whitelist,
```
lkorenc, line 431: `diff_mod->GetFns` feels like a better design - it can return a two element tuple for example, so inside the callee (possibly a few callees later) you can unpack it.

bin/differential_tester_x86/LiftAndCompare.cpp
```diff
@@ -0,0 +1,541 @@
...
+
+      llvm::json::operator<<(o, llvm::json::Value(std::move(arr)));
+    }
+
+    if (!succeeded_tot && FLAGS_stop_on_fail) {
+      return 2;
+    }
+  }
+
+  if (succeeded_tot) {
```
lkorenc, line 536: I would opt for a ternary operator here.

bin/differential_tester_x86/Whitelist.cpp
```diff
@@ -0,0 +1,146 @@
...
+
+                    if (target_ptr) {
+                      *target_ptr = 0;
+                      return;
+                    }
+                    std::string s(target);
+                    throw std::runtime_error(std::string("Unknown reg: ") + s);
+                  }
+
+                 }
```
lkorenc, line 55: Looks like indent is slightly off?

bin/differential_tester_x86/Whitelist.cpp
```diff
@@ -0,0 +1,146 @@
...
+  if (accessors.find(this->section) != accessors.end()) {
+    accessors.find(this->section)->second(state, this->target_name);
+  } else {
+    throw std::runtime_error(std::string("Couldnt find section ") +
+                             std::string(this->section));
+  }
+}
+
+bool Accessor::fromJSON(const llvm::json::Value &E, llvm::json::Path P) {
+
```
lkorenc, line 71: empty lines without much impact?

lkorenc (changes requested):

lib/BC/SleighLifter.cpp
```diff
@@ -247,7 +249,8 @@ class SleighLifter::PcodeToLLVMEmitIntoBlock : public PcodeEmit {
   PcodeToLLVMEmitIntoBlock(llvm::BasicBlock *target_block,
                            llvm::Value *state_pointer, const Instruction &insn,
                            SleighLifter &insn_lifter_parent,
-                           std::vector<std::string> user_op_names_)
+                           std::vector<std::string> user_op_names_,
+                           llvm::BasicBlock *exit_block_)
```
lkorenc, line 267: This is starting to have too much args, consider grouping them up in structures.

lib/BC/SleighLifter.cpp
```diff
@@ -429,16 +433,20 @@ class SleighLifter::PcodeToLLVMEmitIntoBlock : public PcodeEmit {
       case OpCode::CPUI_CALL: {
         // directs dont read the address of the variable, the offset is the jump
         // TODO(Ian): handle other address spaces
-
+        if (input_var.getAddr().getSpace()->constant_space_index ==
```
lkorenc, line 450: This feels like a really leaky abstraction, since we need to go through a lot of random calls to get some info - provide a helper function?

lib/BC/SleighLifter.cpp
```diff
@@ -449,7 +457,9 @@ class SleighLifter::PcodeToLLVMEmitIntoBlock : public PcodeEmit {
         if (copy_inval.has_value()) {
           auto pc_reg = this->LiftNormalRegister(bldr, "PC");
           assert(pc_reg.has_value());
-          return (*pc_reg)->StoreIntoParam(bldr, *copy_inval);
+          auto res = (*pc_reg)->StoreIntoParam(bldr, *copy_inval);
+          this->TerminateBlock();
+          return res;
```
lkorenc, line 476: copy & paste.

lib/BC/SleighLifter.cpp
```diff
@@ -429,16 +433,20 @@ class SleighLifter::PcodeToLLVMEmitIntoBlock : public PcodeEmit {
...
 
         auto input_val = this->replacement_cont.LiftOffsetOrReplace(
             bldr, input_var,
             llvm::IntegerType::get(this->context, input_var.size * 8));
         auto pc_reg = this->LiftNormalRegister(bldr, "PC");
         assert(pc_reg.has_value());
-        return (*pc_reg)->StoreIntoParam(bldr, input_val);
-
-        break;
+        auto res = (*pc_reg)->StoreIntoParam(bldr, input_val);
```
lkorenc, line 461: `pc_reg.Store...`?

lib/BC/SleighLifter.cpp
```diff
@@ -585,6 +595,21 @@ class SleighLifter::PcodeToLLVMEmitIntoBlock : public PcodeEmit {
   static std::unordered_set<OpCode> INTEGER_COMP_OPS;
 
 
+  LiftStatus TerminateBlockWithCondition(llvm::Value *condition) {
+    llvm::IRBuilder<> ir(this->target_block);
+    this->target_block = llvm::BasicBlock::Create(
```
lkorenc, line 614: drop `this->`

lib/BC/SleighLifter.cpp
```diff
@@ -585,6 +595,21 @@ class SleighLifter::PcodeToLLVMEmitIntoBlock : public PcodeEmit {
...
+        this->context, "continuation", this->target_block->getParent());
+    ir.CreateCondBr(condition, this->exit_block, this->target_block);
+    return LiftStatus::kLiftedInstruction;
+  }
+
+  void TerminateBlock() {
```
lkorenc, line 620: Name is kinda misleading - I would try to say that control flow is also redirected somewhere, because just `TerminateBlock` can also mean a simple `return` or other terminator is there.

lib/BC/SleighLifter.cpp
```diff
@@ -593,10 +618,18 @@ class SleighLifter::PcodeToLLVMEmitIntoBlock : public PcodeEmit {
     if (opc == OpCode::CPUI_CBRANCH) {
       auto should_branch = this->LiftInParam(
           bldr, rhs, llvm::IntegerType::get(this->context, rhs.size * 8));
+
+      if (lhs.getAddr().getSpace()->constant_space_index ==
+          lhs.getAddr().getSpace()->getIndex()) {
```
lkorenc, line 637: Copy & paste, so setting a helper fn is probably a good idea.

lib/BC/SleighLifter.cpp
```diff
@@ -687,47 +687,45 @@ class SleighLifter::PcodeToLLVMEmitIntoBlock : public PcodeEmit {
...
-  LiftStatus LiftBoolBinOp(llvm::IRBuilder<> &bldr, OpCode opc,
-                           VarnodeData *outvar, VarnodeData lhs,
-                           VarnodeData rhs) {
-    std::function<llvm::Value *(llvm::Value *, llvm::Value *,
-                                llvm::IRBuilder<> &)>
-        op_func;
+
+  std::optional<llvm::Value *>
+  GetResultOfBoolOpFunc(llvm::IRBuilder<> &bldr, OpCode opc,
+                        llvm::Value *in_lhs, llvm::Value *in_rhs) {
```
lkorenc, line 707: Maybe take params as a collection instead of two separate arguments?

lib/BC/SleighLifter.cpp
```diff
@@ -687,47 +687,45 @@ class SleighLifter::PcodeToLLVMEmitIntoBlock : public PcodeEmit {
...
+  }
+  LiftStatus LiftBoolBinOp(llvm::IRBuilder<> &bldr, OpCode opc,
+                           VarnodeData *outvar, VarnodeData lhs,
+                           VarnodeData rhs) {
+
+
+    auto lifted_lhs = this->LiftInParam(
+             bldr, lhs, llvm::IntegerType::get(this->context, 8)),
+         lifted_rhs = this->LiftInParam(
+             bldr, rhs, llvm::IntegerType::get(this->context, 8));
```
lkorenc: Copy & paste - also I would advise to avoid using `,` as it can be overloaded.

lib/BC/SleighLifter.cpp
```diff
@@ -687,47 +687,45 @@ class SleighLifter::PcodeToLLVMEmitIntoBlock : public PcodeEmit {
...
+    if (!lifted_lhs.has_value() || !lifted_rhs.has_value()) {
+      return LiftStatus::kLiftedUnsupportedInstruction;
     }
-    return LiftStatus::kLiftedUnsupportedInstruction;
+
+    auto computed_value =
+        this->GetResultOfBoolOpFunc(bldr, opc, *lifted_lhs, *lifted_rhs);
+    if (!computed_value.has_value()) {
```
lkorenc, line 737: No need to use `has_value()` - this is multiple times in the codebase and while I do not explicitly mind I just don't see the gain of having it there.

lib/BC/SleighLifter.cpp
```diff
@@ -28,6 +28,20 @@
...
+namespace {
+
+static llvm::Value *ExtractOverflowBitFromCallToIntrinsic(
+    llvm::Intrinsic::IndependentIntrinsics intrinsic, llvm::Value *lhs,
+    llvm::Value *rhs, llvm::IRBuilder<> &bldr) {
+  llvm::Type *overloaded_types[1] = {lhs->getType()};
+  llvm::Function *target_instrinsic = llvm::Intrinsic::getDeclaration(
+      bldr.GetInsertBlock()->getModule(), intrinsic, overloaded_types);
+  std::array<llvm::Value *, 2> intrinsic_args = {lhs, rhs};
+  llvm::Value *res_val = bldr.CreateCall(target_instrinsic, intrinsic_args);
```
lkorenc, line 40: I am pretty sure you can just do `bld.CreateCall(target_intrinsic, { lhs, rhs }` - it even more underlines the fact that they should have been a collection in the arguments as well.

lib/BC/SleighLifter.cpp
```diff
@@ -28,6 +28,20 @@
 
 namespace remill {
 
+namespace {
+
+static llvm::Value *ExtractOverflowBitFromCallToIntrinsic(
+    llvm::Intrinsic::IndependentIntrinsics intrinsic, llvm::Value *lhs,
+    llvm::Value *rhs, llvm::IRBuilder<> &bldr) {
+  llvm::Type *overloaded_types[1] = {lhs->getType()};
```
lkorenc, line 36: No need for array, just use `{}` in the callsite.

tests/Thumb/TestLifting.cpp
```diff
@@ -24,7 +24,61 @@
...
+    auto accessor = reg_to_accessor.find(reg);
+    if (accessor != reg_to_accessor.end()) {
+      CHECK_EQ(accessor->second(state), value);
+    }
+  }
+
+ public:
+  TestOutputSpec(
+      std::string target_bytes, remill::Instruction::Category expected_category,
+      std::vector<std::pair<std::string, uint32_t>> register_preconditions,
```
lkorenc, line 84: Use type alias.

test_runner_lib/include/test_runner/TestRunner.h
```diff
@@ -64,9 +64,11 @@ void StubOutFlagComputationInstrinsics(llvm::Module *mod,
 
 void *MissingFunctionStub(const std::string &name);
 
-template <class T>
-void ExecuteLiftedFunction(llvm::Function *func, size_t insn_length, T *state,
-                           test_runner::MemoryHandler *handler) {
+template <class T, typename P>
```
lkorenc, line 70: Why is one `class` and other `typename`?

test_runner_lib/include/test_runner/TestRunner.h
```diff
@@ -114,13 +116,23 @@ void ExecuteLiftedFunction(llvm::Function *func, size_t insn_length, T *state,
...
   // run until we terminate and exit pc
-  while (state->gpr.rip.dword == orig_pc) {
-    returned(state, state->gpr.rip.dword, handler);
+  while (program_counter_fetch(state) == orig_pc) {
+    returned(state, program_counter_fetch(state), handler);
   }
 }
 
+template <typename T>
+void RandomizeState(T *state, random_bytes_engine &rbe) {
```
lkorenc, line 130: Why is `state` a pointer? Does `nullptr` make sense here? If not -> use ref instead.

lkorenc (changes requested):

bin/differential_tester_x86/LiftAndCompare.cpp
2over12: I tried to refactor this to handle f1 and f2 agnostically ie. refactor everything to handle n lifters, n lifted instructions. There is too much code that depends on this being a compare between two functions at this time, will try to reduce code duplication without refactoring to n functions.

lib/BC/SleighLifter.cpp
2over12: I mean it's sleigh's abstraction not ours :P,  but ive added a helper function that determines if a vnode is in the constant address space

CMakeLists.txt
```diff
@@ -120,15 +121,32 @@ target_link_libraries(thirdparty_gflags INTERFACE
   gflags
 )
 
+set(sleigh_ENABLE_TESTS OFF)
+set(sleigh_ADDITIONAL_PATCHES "${CMAKE_CURRENT_SOURCE_DIR}/patches/sleigh/x86-ia.patch;${CMAKE_CURRENT_SOURCE_DIR}/patches/sleigh/arm-thumb.patch")
+
+# GHIDRA SLEIGH
+FetchContent_Declare(sleigh
+  GIT_REPOSITORY https://github.com/lifting-bits/sleigh.git
+  GIT_TAG master
```
ekilmer, line 130: You should really use a git hash here so you don't have any phantom build errors from broken updates on `master` branch of sleigh.

I know having to manually update a git hash can be annoying, but I would really suggest using a hash to an unreleased commit until there's a release that includes your fix(es) and make an issue to keep track and then change to the release tag if/when sleigh is updated.

lib/BC/SleighLifter.cpp
pgoodman: Can just be `if (!computed_value)`.

lib/BC/SleighLifter.cpp
pgoodman: This could be a new issue.

lib/BC/SleighLifter.cpp
```diff
@@ -0,0 +1,1292 @@
...
+        break;
+      }
+      case OpCode::CPUI_COPY:
+      case OpCode::CPUI_CAST: {
+        auto copy_inval = this->LiftInParam(
+            bldr, input_var,
+            llvm::IntegerType::get(this->context, input_var.size * 8));
+        if (copy_inval.has_value()) {
+          return this->LiftStoreIntoOutParam(bldr, *copy_inval, outvar);
+        }
```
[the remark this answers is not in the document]
pgoodman: This is too big of a change to handle in this PR. The code "works" so I'd prefer to lock in those gains, so to speak. File an issue for creating an improved setup.

lib/BC/SleighLifter.cpp
```diff
@@ -0,0 +1,1196 @@
...
+          bldr.GetInsertBlock()->getModule(), intrinsic_id);
+      llvm::Value *intrinsic_args[] = {*inval};
+      return this->LiftStoreIntoOutParam(
+          bldr, bldr.CreateCall(intrinsic, intrinsic_args), outvar);
+    }
+    return LiftStatus::kLiftedUnsupportedInstruction;
+  }
+
+  LiftStatus LiftUnOp(llvm::IRBuilder<> &bldr, OpCode opc, VarnodeData *outvar,
+                      VarnodeData input_var) {
```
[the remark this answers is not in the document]
pgoodman: Agree, but it also seems to work. I think we should track a refactor with a new issue, and mostly leave this huge function alone for now.

include/remill/Arch/Arch.h
```diff
@@ -62,13 +65,50 @@ enum OSName : uint32_t;
 enum ArchName : uint32_t;
 
 class Arch;
-class ArchImpl;
 class Instruction;
 
+// An RAII locker for handling issues related to SLEIGH.
+class ArchLocker {
```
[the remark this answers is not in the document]
lkorenc: I absolutely do not see how this comment justifies existence of this class that seems to be a re-implementation of `std::lock_guard` (it can also just take something without locking it).

And `Arch` is an interface for which each implementation knows whether it should or should not lock, so I am confused what are we gaining by this at all.

lib/Arch/Arch.cpp
```diff
@@ -195,17 +237,9 @@ auto Arch::Build(llvm::LLVMContext *context_, OSName os_name_,
       ret = GetSPARC64(context_, os_name_, arch_name_);
       break;
     }
-
-    case kArchThumb2LittleEndian: {
-      DLOG(WARNING) << "Using architecture: Aarch32/Thumb2. WARNING: not fully implemented at this point.";
-      //TODO(artem): Fix this once Thumb2 fully supported
-      ret = GetAArch32(context_, os_name_, arch_name_);
-      break;
-    }
```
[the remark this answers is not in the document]
lkorenc: This comment was targeting the currently used codestyle.

lib/Arch/Arch.cpp
```diff
@@ -271,17 +300,16 @@ const Register *Arch::RegisterAtStateOffset(uint64_t offset) const {
...
 // Return information about a register, given its name.
-const Register *Arch::RegisterByName(std::string_view name_) const {
+const Register *ArchBase::RegisterByName(std::string_view name_) const {
   std::string name(name_.data(), name_.size());
-  auto [curr_val_it, added] =
-      impl->reg_by_name.emplace(std::move(name), nullptr);
+  auto [curr_val_it, added] = reg_by_name.emplace(std::move(name), nullptr);
   if (added) {
     return nullptr;
   } else {
```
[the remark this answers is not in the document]
lkorenc: Or junk.

lib/Arch/Arch.cpp
```diff
@@ -735,13 +767,14 @@ void Arch::PrepareModule(llvm::Module *mod) const {
...
-                                  const char *parent_reg_name) const {
+const Register *ArchBase::AddRegister(const char *reg_name_,
+                                      llvm::Type *val_type, size_t offset,
+                                      const char *parent_reg_name) const {
+
   CHECK_NOTNULL(val_type);
 
   const std::string reg_name(reg_name_);
-  auto &reg = impl->reg_by_name[reg_name];
+  auto &reg = reg_by_name[reg_name];
```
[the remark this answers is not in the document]
lkorenc: Maybe let's not do that and properly use a store into collection once we need to insert a new value?

bin/differential_tester_x86/LiftAndCompare.cpp
```diff
@@ -0,0 +1,811 @@
...
+  llvm::Module *semantics_module;
+  remill::InstructionLifter::LifterPtr lifter;
+  std::unique_ptr<remill::IntrinsicTable> table;
+  remill::Arch::ArchPtr arch;
+
+ public:
+  LiftingTester(llvm::Module *semantics_module_, remill::OSName os_name,
+                remill::ArchName arch_name)
+      : semantics_module(semantics_module_) {
+    this->arch = remill::Arch::Build(&semantics_module_->getContext(), os_name,
```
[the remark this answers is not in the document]
2over12: Ive moved this where possible, but sometimes we do need to call init from semantic module etc

bin/differential_tester_x86/LiftAndCompare.cpp
```diff
@@ -0,0 +1,811 @@
...
+
+    llvm::EngineBuilder builder(std::move(tgt_mod));
+
+
+    std::string estr;
+    auto eptr = builder.setEngineKind(llvm::EngineKind::JIT)
+                    .setErrorStr(&estr)
+                    .create();
+
+    if (eptr == nullptr) {
```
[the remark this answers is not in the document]
2over12: Want to check if we should print the error out param

bin/differential_tester_x86/Whitelist.cpp
```diff
@@ -0,0 +1,142 @@
...
+                    throw std::runtime_error(std::string("Unknown reg: ") + s);
+                  }},
+                 {"aflags",
+                  [](X86State *state, std::string_view target) {
+                    uint8_t *target_ptr = nullptr;
+                    if (target == "af") {
+                      target_ptr = &state->aflag.af;
+                    }
+
+                    if (target == "zf") {
```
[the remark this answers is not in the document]
2over12: This is a pain to implement for the entire state... id love to have this

lib/Arch/Sleigh/X86Arch.cpp
```diff
@@ -0,0 +1,596 @@
...
+  }
+
+  void PopulateRegisterTable(void) const final {
+    reg_by_offset.resize(sizeof(X86State));
+
+    CHECK_NOTNULL(context);
+
+    bool has_avx = false;
+    bool has_avx512 = false;
+    switch (arch_name) {
```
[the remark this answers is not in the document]
2over12: Need to come back and address this, this is going to get looped into the larger change to share code between x86 -> x86 sleigh and thumsleigh aarch32

lib/BC/SleighLifter.cpp
2over12: Doesnt work with an internal header like this, maybe if i add it to the cmake as a private header then i could do that?

lib/Arch/Sleigh/X86Arch.cpp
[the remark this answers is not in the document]
2over12: This and the following have been addressed in unifying portions of the x86 arch

bin/differential_tester_x86/LiftAndCompare.cpp
```diff
@@ -305,44 +278,42 @@ class ComparisonRunner {
...
-    func1_state->addr.es_base.dword = 0;
-    func1_state->addr.cs_base.dword = 0;
-    func1_state->aflag.af = test_runner::random_boolean_flag(this->rbe);
-    func1_state->aflag.cf = test_runner::random_boolean_flag(this->rbe);
-    func1_state->aflag.df = test_runner::random_boolean_flag(this->rbe);
-    func1_state->aflag.of = test_runner::random_boolean_flag(this->rbe);
-    func1_state->aflag.pf = test_runner::random_boolean_flag(this->rbe);
-    func1_state->aflag.sf = test_runner::random_boolean_flag(this->rbe);
-    func1_state->aflag.zf = test_runner::random_boolean_flag(this->rbe);
+    X86State func1_state = {};
```
lkorenc, line 241: `= {}` is not needed, default ctor should be called even if you just do `T val;`

lib/BC/SleighLifter.cpp
```diff
@@ -308,6 +308,18 @@ class SleighLifter::PcodeToLLVMEmitIntoBlock : public PcodeEmit {
     }
   }
 
+  ParamPtr LiftNormalRegisterOrCreateUnique(llvm::IRBuilder<> &bldr,
+                                            std::string reg_name,
+                                            VarnodeData target_vnode) {
+    if (auto res = this->LiftNormalRegister(bldr, reg_name)) {
+      return *res;
+    } else {
```
lkorenc, line 318: else after return.

lkorenc (approved):

PR #610, opened 2022-07-25
Implement GET_PC idiom

tetsuo-cpp:
This is an initial attempt at implementing the CALL+POP idiom that occurs in x86 code to assign the PC to a register.

### DIFF
--- a/lib/Arch/X86/Arch.cpp
+++ b/lib/Arch/X86/Arch.cpp
@@ -827,7 +827,6 @@ static bool TryDecodeIdioms(Instruction &inst) {
     assert(pop_inst.operands.size() == 1);
     inst.operands = std::move(pop_inst.operands);
 
-    inst.pc = call_pc;
     inst.category = Instruction::kCategoryNormal;
     inst.function = std::string(kGetPCISelPrefix) +
                     std::to_string(inst.operands.front().size);
--- a/lib/Arch/X86/Arch.cpp
+++ b/lib/Arch/X86/Arch.cpp
@@ -788,7 +788,8 @@ static void DecodeOperand(Instruction &inst, const xed_decoded_inst_t *xedd,
   }
 }
 
-static uint64_t BytesToBits(const uint8_t *bytes) {
+static uint64_t BytesToCallBits(const uint8_t *bytes) {
+  // The pattern we're looking for is 5 bytes long.
   uint64_t bits = 0;
   bits = (bits << 8) | static_cast<uint32_t>(bytes[0]);
   bits = (bits << 8) | static_cast<uint32_t>(bytes[1]);
@@ -801,13 +802,13 @@ static uint64_t BytesToBits(const uint8_t *bytes) {
 static bool TryDecodeIdioms(Instruction &inst) {
   // Check for CALL+POP idiom used to retrieve PC.
   if (inst.bytes.size() == 6) {
-    const auto call_pc = inst.pc;
     const auto bytes = reinterpret_cast<const uint8_t *>(inst.bytes.data());
-    const auto bits1 = BytesToBits(bytes);
+    const auto call_pc = inst.pc;
+    const auto call_bits = BytesToCallBits(bytes);
 
     // Check that we have a CALL pointing one instruction ahead of where we are
     // now.
-    if (bits1 != 0xe800000000) {
+    if (call_bits != 0xe800000000) {
       return false;
     }
 
--- a/lib/Arch/X86/Arch.cpp
+++ b/lib/Arch/X86/Arch.cpp
@@ -799,55 +799,61 @@ static uint64_t BytesToCallBits(const uint8_t *bytes) {
   return bits;
 }
 
-static bool TryDecodeIdioms(Instruction &inst) {
+static bool TryDecodeGetPCIdiom(Instruction &inst) {
   // Check for CALL+POP idiom used to retrieve PC.
-  if (inst.bytes.size() == 6) {
-    const auto bytes = reinterpret_cast<const uint8_t *>(inst.bytes.data());
-    const auto call_bits = BytesToCallBits(bytes);
-
-    // Check that we have a CALL pointing one instruction ahead of where we are
-    // now.
-    if (call_bits != 0xe800000000) {
-      return false;
-    }
-
-    // Check whether we have a POP instruction to retrieve the return address
-    // that we just pushed to the stack.
-    Instruction pop_inst;
-    if (!inst.arch->DecodeInstruction(inst.pc + 5, &inst.bytes[5], pop_inst) ||
-        pop_inst.function.substr(0, 3) != "POP") {
-      return false;
-    }
-
-    // We've matched the idiom. Now let's call our custom intrinsic that assigns
-    // the PC to the operand register.
-
-    // The first operand should be the register that we're popping into. Just
-    // take the operand from the POP instruction.
-    CHECK(pop_inst.operands.size() == 1)
-        << "Unexpected number of POP operands, expected=1, got="
-        << pop_inst.operands.size();
-    inst.operands = std::move(pop_inst.operands);
-
-    // The second operand is the address of the POP instruction.
-    inst.operands.emplace_back();
-    auto &addr_op = inst.operands.back();
-    addr_op.type = Operand::kTypeAddressExpression;
-    auto *pc_reg =
-        inst.EmplaceRegister(inst.arch->ProgramCounterRegisterName());
-    CHECK(pc_reg);
-    auto *pc_offset = inst.EmplaceConstant(
-        llvm::ConstantInt::get(inst.arch->AddressType(), 5));
-    CHECK(pc_offset);
-    addr_op.expr =
-        inst.EmplaceBinaryOp(llvm::Instruction::Add, pc_reg, pc_offset);
-
-    inst.category = Instruction::kCategoryNormal;
-    inst.function = std::string(kGetPCISelPrefix) +
-                    std::to_string(inst.operands.front().size);
-    return true;
+  if (inst.bytes.size() != 6) {
+    return false;
   }
 
+  const auto bytes = reinterpret_cast<const uint8_t *>(inst.bytes.data());
+  const auto call_bits = BytesToCallBits(bytes);
+
+  // Check that we have a CALL pointing one instruction ahead of where we are
+  // now.
+  if (call_bits != 0xe800000000) {
+    return false;
+  }
+
+  // Check whether we have a POP instruction to retrieve the return address
+  // that we just pushed to the stack.
+  Instruction pop_inst;
+  if (!inst.arch->DecodeInstruction(inst.pc + 5, &inst.bytes[5], pop_inst) ||
+      pop_inst.function.substr(0, 3) != "POP") {
+    return false;
+  }
+
+  // We've matched the idiom. Now let's call our custom intrinsic that assigns
+  // the PC to the operand register.
+
+  // The first operand should be the register that we're popping into. Just
+  // take the operand from the POP instruction.
+  CHECK(pop_inst.operands.size() == 1)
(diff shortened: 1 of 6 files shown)


lib/Arch/X86/Arch.cpp
```diff
@@ -788,6 +788,55 @@ static void DecodeOperand(Instruction &inst, const xed_decoded_inst_t *xedd,
...
+    // Check that we have a CALL pointing one instruction ahead of where we are
+    // now.
+    if (call_bits != 0xe800000000) {
+      return false;
+    }
+
+    // Check whether we have a POP instruction to retrieve the return address
+    // that we just pushed to the stack.
+    Instruction pop_inst;
+    if (!inst.arch->DecodeInstruction(inst.pc + 5, &inst.bytes[5], pop_inst) ||
```
tetsuo-cpp, line 817: Not sure whether this is appropriate since we're within the `DecodeInstruction` path already. I could avoid this (with a bit more mess) by calling `DecodeXED` and using the XED structures directly.

lib/Arch/X86/Semantics/GET_PC.cpp
```diff
@@ -0,0 +1,28 @@
...
+#pragma once
+
+template <typename D>
+DEF_SEM(GET_PC, D dst) {
+  addr_t pc = Read(REG_PC);
+  Write(dst, Read(ReadPtr<D>(pc)));
+  return memory;
+}
+
+DEF_ISEL(GET_PC_16) = GET_PC<M16W>;
```
tetsuo-cpp, line 26: I don't really know what these type names like `M16W` and `RnW` refer to. I just copied this from the POP implementation.

lib/Arch/X86/Arch.cpp
```diff
@@ -788,6 +788,54 @@ static void DecodeOperand(Instruction &inst, const xed_decoded_inst_t *xedd,
...
+      return false;
+    }
+
+    // We've matched the idiom. Now let's call our custom intrinsic that assigns
+    // the PC to the operand register.
+
+    // The operand should be the register that we're popping into. Just take the
+    // operand from the POP instruction.
+    assert(pop_inst.operands.size() == 1);
+    inst.operands = std::move(pop_inst.operands);
```
tetsuo-cpp, line 828: - [ ] Still need to add an address calculation operand to calculate the return address of the `CALL`. So instead of assigning PC to the register, it should be PC + 5.

lib/Arch/X86/Semantics/GET_PC.cpp
tetsuo-cpp: Oh, there's some information in the "life of an instruction" doc. I'll check that out.

lib/Arch/X86/Arch.cpp
tetsuo-cpp: [c2c5654](https://github.com/lifting-bits/remill/pull/610/commits/c2c56541b50263750348f5308ad16c246a823949)

lib/Arch/X86/Semantics/GET_PC.cpp
tetsuo-cpp: [07abb69](https://github.com/lifting-bits/remill/pull/610/commits/07abb69d3bc0daa72c4dcaf277b567c1aa4b7f45)

tetsuo-cpp:
@pgoodman The `State` structure seems to be getting removed from the generated IR for the example we were looking at.

lib/Arch/X86/Arch.cpp
```diff
@@ -788,6 +788,66 @@ static void DecodeOperand(Instruction &inst, const xed_decoded_inst_t *xedd,
   }
 }
 
+static uint64_t BytesToCallBits(const uint8_t *bytes) {
+  // The pattern we're looking for is 5 bytes long.
+  uint64_t bits = 0;
+  bits = (bits << 8) | static_cast<uint32_t>(bytes[0]);
```
lkorenc, line 794: For loop?

lib/Arch/X86/Arch.cpp
```diff
@@ -788,6 +788,66 @@ static void DecodeOperand(Instruction &inst, const xed_decoded_inst_t *xedd,
...
+  bits = (bits << 8) | static_cast<uint32_t>(bytes[1]);
+  bits = (bits << 8) | static_cast<uint32_t>(bytes[2]);
+  bits = (bits << 8) | static_cast<uint32_t>(bytes[3]);
+  bits = (bits << 8) | static_cast<uint32_t>(bytes[4]);
+  return bits;
+}
+
+static bool TryDecodeIdioms(Instruction &inst) {
+  // Check for CALL+POP idiom used to retrieve PC.
+  if (inst.bytes.size() == 6) {
```
lkorenc, line 804: Early return here - if we know already that is not `6` just return `false` - it will simplify the code (because you do not need to scroll all way down for that `return false` and spare an indent.

lib/Arch/X86/Arch.cpp
```diff
@@ -788,6 +788,66 @@ static void DecodeOperand(Instruction &inst, const xed_decoded_inst_t *xedd,
...
+        pop_inst.function.substr(0, 3) != "POP") {
+      return false;
+    }
+
+    // We've matched the idiom. Now let's call our custom intrinsic that assigns
+    // the PC to the operand register.
+
+    // The first operand should be the register that we're popping into. Just
+    // take the operand from the POP instruction.
+    assert(pop_inst.operands.size() == 1);
```
lkorenc, line 827: `CHECK`?

lib/Arch/X86/Arch.cpp
tetsuo-cpp: Do we expect to add more idioms? That's the reason I wrote it like this (and named the function in a more generic way).

lkorenc: That's a valid point, but I would consider extracting this idiom as separate function then? We don't want to have them all inlined.

pgoodman: Start with `inst.bytes.size() >= 6`, as you might be given up to 15 bytes.

lib/Arch/X86/Arch.cpp
pgoodman: It seems like a `memcmp` again the bytes in a static const array could also work.

lib/Arch/X86/Arch.cpp
pgoodman: I think there's an issue here where it's converting a `const char *` to a `std::string_view`, which in turn relies on `strlen`. You want to give it a one-byte view into the `inst.bytes` instead.

pgoodman: I would also recomment being more selective about the `pop_inst.function`, as theoretically this could match on other `POP` variants.

lib/Arch/X86/Arch.cpp
```diff
@@ -788,6 +788,66 @@ static void DecodeOperand(Instruction &inst, const xed_decoded_inst_t *xedd,
...
+    inst.operands = std::move(pop_inst.operands);
+
+    // The second operand is the address of the POP instruction.
+    inst.operands.emplace_back();
+    auto &addr_op = inst.operands.back();
+    addr_op.type = Operand::kTypeAddressExpression;
+    auto *pc_reg = inst.EmplaceRegister("EIP");
```
pgoodman, line 834: Use `"PC"` or call the method on `Arch` that gets the current program counter name.

pgoodman (changes requested):

lib/Arch/X86/Arch.cpp
tetsuo-cpp: [fa2e66e](https://github.com/lifting-bits/remill/pull/610/commits/fa2e66e3b84423d11aafee18150e733cad2cb30c)

lib/Arch/X86/Arch.cpp
tetsuo-cpp: [23d08cd](https://github.com/lifting-bits/remill/pull/610/commits/23d08cdcf4f6d2b8f6db8f9dc666b0be98de390d)

lib/Arch/X86/Arch.cpp
tetsuo-cpp: Good idea.

[5a4470b](https://github.com/lifting-bits/remill/pull/610/commits/5a4470b69bb3023fafdcbd1c61b0268d466243a8)

lib/Arch/X86/Arch.cpp
tetsuo-cpp: > I think there's an issue here where it's converting a `const char *` to a `std::string_view`, which in turn relies on `strlen`. You want to give it a one-byte view into the `inst.bytes` instead.

Why is that an issue? I'm just taking an index into `inst.bytes` which is a `std::string` so that should be a valid C-string.

lib/Arch/X86/Arch.cpp
tetsuo-cpp: [de4ec45](https://github.com/lifting-bits/remill/pull/610/commits/de4ec451c95237019aa5a1c7c9f990f3ae492afb)

lib/Arch/X86/Arch.cpp
tetsuo-cpp: > I would also recomment being more selective about the `pop_inst.function`, as theoretically this could match on other `POP` variants.

What `POP` functions would it not apply for? I had a look at the stuff in `POP.cpp` and from what I can see, they would all have the "get pc" effect when combined with the call.

lib/Arch/X86/Arch.cpp
```diff
@@ -788,36 +788,26 @@ static void DecodeOperand(Instruction &inst, const xed_decoded_inst_t *xedd,
...
-  bits = (bits << 8) | static_cast<uint32_t>(bytes[2]);
-  bits = (bits << 8) | static_cast<uint32_t>(bytes[3]);
-  bits = (bits << 8) | static_cast<uint32_t>(bytes[4]);
-  return bits;
-}
+static unsigned char GET_PC_IDIOM_CALL[] = {0xE8, 0x00, 0x00, 0x00, 0x00};
 
 static bool TryDecodeGetPCIdiom(Instruction &inst) {
   // Check for CALL+POP idiom used to retrieve PC.
   if (inst.bytes.size() != 6) {
```
pgoodman, line 795: This feels overly restrictive. I think it should be saying: if we don't have /at least/ six bytes, then we can't find our idiom. If we have at least 6, then we can. The follow-on issue is that if we have more than 6 bytes, then we might have a different form of POP on our hands. It seems reasonable-ish to me to generically handle this, e.g. have a static const mapping of ISEL names of supported trailing pops, mapping to isel names for the new idiom.

lib/Arch/X86/Arch.cpp
```diff
@@ -838,8 +828,8 @@ static bool TryDecodeGetPCIdiom(Instruction &inst) {
   addr_op.type = Operand::kTypeAddressExpression;
```
pgoodman, line 828: After decoding, you'll want to `inst.bytes.resize()` to `call_size + pop_inst.bytes.size()`.

pgoodman (changes requested):

lib/Arch/X86/Arch.cpp
tetsuo-cpp: [95bfd8d](https://github.com/lifting-bits/remill/pull/610/commits/95bfd8d4ccf49ffa9ebbc92e69983aa5cc051abd)

lib/Arch/X86/Arch.cpp
tetsuo-cpp: @pgoodman

tetsuo-cpp:
There's already some attempt to recognise this idiom in Remill so I'm going to explore patching that first.

tetsuo-cpp:
Ok, superseded by #612.

Issue #173, opened 2017-12-04
Remove preprocessor include guards left in from a previous refactoring


mike-myers-tob:
```shell
$ grep --include *.cpp -Rn '#define.*_H_'
remill/Arch/X86/Semantics/STRINGOP.cpp:18:#define REMILL_ARCH_X86_SEMANTICS_STRINGOP_H_
remill/Arch/X86/Semantics/MMX.cpp:18:#define REMILL_ARCH_X86_SEMANTICS_MMX_H_
remill/Arch/X86/Semantics/ROTATE.cpp:18:#define REMILL_ARCH_X86_SEMANTICS_ROTATE_H_
remill/Arch/X86/Semantics/BINARY.cpp:18:#define REMILL_ARCH_X86_SEMANTICS_BINARY_H_
remill/Arch/X86/Semantics/POP.cpp:18:#define REMILL_ARCH_X86_SEMANTICS_POP_H_
remill/Arch/X86/Semantics/FLAGOP.cpp:18:#define REMILL_ARCH_X86_SEMANTICS_FLAGOP_H_
remill/Arch/X86/Semantics/SHIFT.cpp:18:#define REMILL_ARCH_X86_SEMANTICS_SHIFT_H_
remill/Arch/X86/Semantics/FLAGS.cpp:18:#define REMILL_ARCH_X86_SEMANTICS_FLAGS_H_
remill/Arch/X86/Semantics/CONVERT.cpp:18:#define REMILL_ARCH_X86_SEMANTICS_CONVERT_H_
remill/Arch/X86/Semantics/LOGICAL.cpp:18:#define REMILL_ARCH_X86_SEMANTICS_LOGICAL_H_
remill/Arch/X86/Semantics/PUSH.cpp:18:#define REMILL_ARCH_X86_SEMANTICS_PUSH_H_
remill/Arch/X86/Semantics/XOP.cpp:18:#define REMILL_ARCH_X86_SEMANTICS_XOP_H_
remill/Arch/X86/Semantics/X87.cpp:18:#define REMILL_ARCH_X86_SEMANTICS_X87_H_
remill/Arch/X86/Semantics/CALL_RET.cpp:18:#define REMILL_ARCH_X86_SEMANTICS_CALL_RET_H_
remill/Arch/X86/Semantics/MISC.cpp:18:#define REMILL_ARCH_X86_SEMANTICS_MISC_H_
[...]

PR #611, opened 2022-07-28
Provide definition for `__remill_sync_hyper_call`

tetsuo-cpp:
Partially addresses #173

### DIFF
--- a/lib/Arch/Runtime/HyperCall.cpp
+++ b/lib/Arch/Runtime/HyperCall.cpp
@@ -41,16 +41,37 @@ Memory *__remill_sync_hyper_call(State &state, Memory *mem,
                      "d"(state.gpr.rdx.aword));
       break;
 
-    case SyncHyperCall::kX86LoadGlobalDescriptorTable: break;
+    case SyncHyperCall::kX86LoadGlobalDescriptorTable:
+      // TODO(alex): Where do I get the operand from?
+      asm volatile("lgdt" : : "=m"(__remill_read_memory_64(memory, 0)));
+      break;
 
-    case SyncHyperCall::kX86LoadInterruptDescriptorTable: break;
+    case SyncHyperCall::kX86LoadInterruptDescriptorTable:
+      // TODO(alex): Where do I get the operand from?
+      asm volatile("lidt" : : "m"(__remill_read_memory_64(memory, 0)));
+      break;
 
-    case SyncHyperCall::kX86ReadModelSpecificRegister: break;
+    case SyncHyperCall::kX86ReadModelSpecificRegister:
+      asm volatile("rdmsr"
+                   : "=c"(state.gpr.rcx.dword)
+                   : "a"(state.gpr.rax.dword), "d"(state.gpr.rdx.dword));
+      break;
 
-    case SyncHyperCall::kX86WriteModelSpecificRegister: break;
+    case SyncHyperCall::kX86WriteModelSpecificRegister:
+      asm volatile("wrmsr"
+                   : "=c"(state.gpr.rcx.dword)
+                   : "a"(state.gpr.rax.dword), "d"(state.gpr.rdx.dword));
+      break;
 
-    case SyncHyperCall::kX86WriteBackInvalidate: break;
+    case SyncHyperCall::kX86WriteBackInvalidate:
+      // NOTE(alex): This just clears cache so there's no affect on the state.
+      asm volatile("wbinvd" :);
+      break;
 
+    // TODO(alex): There doesn't seem to be a way to figure out what what value
+    // gets written to these segment registers. Is there a reason we can't
+    // just do a write to these registers in the semantic functions
+    // themselves? I don't really get why these are sync hyper calls.
     case SyncHyperCall::kX86SetSegmentES: break;
 
     case SyncHyperCall::kX86SetSegmentSS: break;
@@ -93,6 +114,7 @@ Memory *__remill_sync_hyper_call(State &state, Memory *mem,
 
 #  endif
 
+// TODO(alex): What variable gets set for ARM32?
 #elif defined(REMILL_ON_AARCH64)
 
     case SyncHyperCall::kAArch64EmulateInstruction: break;
--- a/lib/Arch/Runtime/HyperCall.cpp
+++ b/lib/Arch/Runtime/HyperCall.cpp
@@ -1,0 +1,166 @@
+/*
+ * Copyright (c) 2022 Trail of Bits, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+Memory *__remill_sync_hyper_call(State &state, Memory *mem,
+                                 SyncHyperCall::Name call) {
+  switch (call) {
+
+#if defined(REMILL_ON_X86) or defined(REMILL_ON_AMD64)
+
+    case SyncHyperCall::kX86CPUID:
+      asm volatile("cpuid"
+                   : "=a"(state.gpr.rax.aword), "=b"(state.gpr.rbx.aword),
+                     "=c"(state.gpr.rcx.aword), "=d"(state.gpr.rdx.aword)
+                   : "a"(state.gpr.rax.aword), "b"(state.gpr.rbx.aword),
+                     "c"(state.gpr.rcx.aword), "d"(state.gpr.rdx.aword));
+      break;
+
+    case SyncHyperCall::kX86ReadTSC:
+      asm volatile("rdtsc"
+                   : "=a"(state.gpr.rax.dword), "=d"(state.gpr.rdx.dword));
+      break;
+
+    case SyncHyperCall::kX86ReadTSCP:
+      asm volatile("rdtscp"
+                   : "=a"(state.gpr.rax.aword), "=c"(state.gpr.rcx.aword),
+                     "=d"(state.gpr.rdx.aword)
+                   : "a"(state.gpr.rax.aword), "c"(state.gpr.rcx.aword),
+                     "d"(state.gpr.rdx.aword));
+      break;
+
+    case SyncHyperCall::kX86LoadGlobalDescriptorTable: break;
+
+    case SyncHyperCall::kX86LoadInterruptDescriptorTable: break;
+
+    case SyncHyperCall::kX86ReadModelSpecificRegister: break;
+
+    case SyncHyperCall::kX86WriteModelSpecificRegister: break;
+
+    case SyncHyperCall::kX86WriteBackInvalidate: break;
+
+    case SyncHyperCall::kX86SetSegmentES: break;
+
+    case SyncHyperCall::kX86SetSegmentSS: break;
+
+    case SyncHyperCall::kX86SetSegmentDS: break;
+
+    case SyncHyperCall::kX86SetSegmentFS: break;
+
+    case SyncHyperCall::kX86SetSegmentGS: break;
+
+#  if defined(REMILL_ON_X86)
+
+    case SyncHyperCall::kX86SetDebugReg: break;
+
+    case SyncHyperCall::kX86SetControlReg0: break;
+
+    case SyncHyperCall::kX86SetControlReg1: break;
+
+    case SyncHyperCall::kX86SetControlReg2: break;
+
+    case SyncHyperCall::kX86SetControlReg3: break;
+
+    case SyncHyperCall::kX86SetControlReg4: break;
+
+#  elif defined(REMILL_ON_AMD64)
+
+    case SyncHyperCall::kAMD64SetDebugReg: break;
+
+    case SyncHyperCall::kAMD64SetControlReg0: break;
+
+    case SyncHyperCall::kAMD64SetControlReg1: break;
+
+    case SyncHyperCall::kAMD64SetControlReg2: break;
+
+    case SyncHyperCall::kAMD64SetControlReg3: break;
+
+    case SyncHyperCall::kAMD64SetControlReg4: break;
+
+    case SyncHyperCall::kAMD64SetControlReg8: break;
+
+#  endif
+
+#elif defined(REMILL_ON_AARCH64)
+
+    case SyncHyperCall::kAArch64EmulateInstruction: break;
+
+    case SyncHyperCall::kAArch64Breakpoint: break;
+
+    case SyncHyperCall::kAArch32EmulateInstruction: break;
+
+    case SyncHyperCall::kAArch32CheckNotEL2: break;
+
+#elif defined(REMILL_ON_SPARC32) or defined(REMILL_ON_SPARC64)
+
+    case SyncHyperCall::kSPARCSetAsiRegister: break;
+
+    case SyncHyperCall::kSPARCTagOverflow: break;
+
+    case SyncHyperCall::kSPARCUnimplementedInstruction: break;
+
+    case SyncHyperCall::kSPARCUnhandledDCTI: break;
+
+    case SyncHyperCall::kSPARCWindowUnderflow: break;
+
+    case SyncHyperCall::kSPARCTrapCondA: break;
+
+    case SyncHyperCall::kSPARCTrapCondN: break;
+
+    case SyncHyperCall::kSPARCTrapCondNE: break;
+
+    case SyncHyperCall::kSPARCTrapCondE: break;
+
+    case SyncHyperCall::kSPARCTrapCondG: break;
+
+    case SyncHyperCall::kSPARCTrapCondLE: break;
+
+    case SyncHyperCall::kSPARCTrapCondGE: break;
+
+    case SyncHyperCall::kSPARCTrapCondL: break;
+
+    case SyncHyperCall::kSPARCTrapCondGU: break;
+
+    case SyncHyperCall::kSPARCTrapCondLEU: break;
+
+    case SyncHyperCall::kSPARCTrapCondCC: break;
+
+    case SyncHyperCall::kSPARCTrapCondCS: break;
+
+    case SyncHyperCall::kSPARCTrapCondPOS: break;
+
+    case SyncHyperCall::kSPARCTrapCondNEG: break;
+
+    case SyncHyperCall::kSPARCTrapCondVC: break;
+
+    case SyncHyperCall::kSPARCTrapCondVS: break;
+
+#  if defined(REMILL_ON_SPARC32)
+
+    case SyncHyperCall::kSPARC32EmulateInstruction: break;
+
+#  elif defined(REMILL_ON_SPARC64)
+
+    case SyncHyperCall::kSPARC64EmulateInstruction: break;
+
+#  endif
+
+#endif
+
+    default: abort();
+  }
+
+  return mem;
+}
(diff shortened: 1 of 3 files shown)


tetsuo-cpp:
This is incomplete, I just wanted to ask some questions and check that this approach is looking ok.

lib/Arch/Runtime/HyperCall.cpp
```diff
@@ -0,0 +1,188 @@
...
+                   : "=c"(state.gpr.rcx.dword)
+                   : "a"(state.gpr.rax.dword), "d"(state.gpr.rdx.dword));
+      break;
+
+    case SyncHyperCall::kX86WriteBackInvalidate:
+      // NOTE(alex): This just clears cache so there's no affect on the state.
+      asm volatile("wbinvd" :);
+      break;
+
+    // TODO(alex): There doesn't seem to be a way to figure out what what value
```
tetsuo-cpp, line 71: @pgoodman Do you know the reason behind this?

lib/Arch/Runtime/HyperCall.cpp
```diff
@@ -0,0 +1,188 @@
...
+    case SyncHyperCall::kX86ReadTSCP:
+      asm volatile("rdtscp"
+                   : "=a"(state.gpr.rax.aword), "=c"(state.gpr.rcx.aword),
+                     "=d"(state.gpr.rdx.aword)
+                   : "a"(state.gpr.rax.aword), "c"(state.gpr.rcx.aword),
+                     "d"(state.gpr.rdx.aword));
+      break;
+
+    case SyncHyperCall::kX86LoadGlobalDescriptorTable:
+      // TODO(alex): Where do I get the operand from?
```
tetsuo-cpp, line 45: And this.

lib/Arch/Runtime/HyperCall.cpp
```diff
@@ -0,0 +1,188 @@
...
+
+Memory *__remill_sync_hyper_call(State &state, Memory *mem,
+                                 SyncHyperCall::Name call) {
+  switch (call) {
+
+#if defined(REMILL_ON_X86) or defined(REMILL_ON_AMD64)
+
+    case SyncHyperCall::kX86CPUID:
+      asm volatile("cpuid"
+                   : "=a"(state.gpr.rax.aword), "=b"(state.gpr.rbx.aword),
```
pgoodman, line 25: Here you want to make some local variables, e.g. at the top of the function, then have the inline assembly assign into them, then after the inline assembly, copy the values of those locals into the `State`. The reason for this separation is so that the inline assembly block isn't holding pointers into the state structure, which would trigger a state escape.

lib/Arch/Runtime/HyperCall.cpp
pgoodman: https://github.com/lifting-bits/remill/blob/f096390b95d8528ff28e7e8d178b3577e34e0caa/lib/Arch/X86/Semantics/SYSTEM.cpp#L19-L62

`state.addr_to_load`.

lib/Arch/Runtime/HyperCall.cpp
pgoodman: I am not sure. I'd have to look at the arch docs to see how this instruction works.

pgoodman (changes requested):

lib/Arch/Runtime/HyperCall.cpp
tetsuo-cpp: I tested on the function we looked at that had a `cpuid` instruction on it but didn't see a state escape with this patch. Does this only happen in certain situations?

lib/Arch/Runtime/HyperCall.cpp
tetsuo-cpp: [e58a2f2](https://github.com/lifting-bits/remill/pull/611/commits/e58a2f291cac43263b864e32bee709fe59a4dc35)

lib/Arch/Runtime/HyperCall.cpp
tetsuo-cpp: This one doesn't map to an instruction. It just gets as part of other instructions that eventually write to a segment register. What I don't understand is that the `State` structure has these segment registers.

Here's an instruction where it uses the hyper call:
https://github.com/lifting-bits/remill/blob/master/lib/Arch/X86/Semantics/POP.cpp#L163-L171

And here are the segment registers:
https://github.com/lifting-bits/remill/blob/master/include/remill/Arch/X86/Runtime/State.h#L463-L476

There are also some other registers called `es_base`, `ss_base`, etc. These ones actually get written to in some of the semantic functions:
https://github.com/lifting-bits/remill/blob/master/include/remill/Arch/X86/Runtime/State.h#L672-L685
https://github.com/lifting-bits/remill/blob/master/lib/Arch/X86/Semantics/CALL_RET.cpp#L26

I'm not sure what the distinction between the base and non-base variants is because the segment registers are also meant to hold the "base" address of the segment.

tetsuo-cpp:
I'm trying to figure out what this is doing: https://github.com/lifting-bits/remill/blob/master/cmake/BCCompiler.cmake#L168

It looks like I'll have to add something to here to compile the runtimes for the architecture they're for. At the moment, the inline assembly for non-x86 architectures doesn't work for this reason.

pgoodman:
@tetsuo-cpp that looks like a tricky issue to solve, especially with Apple Clang. That used to be a viable target. Over time, I've tried to add a few things to `Runtime/`, e.g. `Float.h`, that would try to remove or factor out dependence on system libraries, due to issues in trying to do cross-compilation. I think I took it all a bit further even in my unfinished msp430 branch in this repo. At the end of the day, you might have to fix the target triple (what that string is) to something linux based. I think we have methods with hard-coded platform-specific triples in remill, e.g. `Arch::DefaultTriple` or something like that. You could try one of those strings here, and then evaluate what breaks. This might be better suited toward a different issue / PR, though.

lib/Arch/Runtime/HyperCall.cpp
pgoodman: That pleasently surprises me.

pgoodman: What do you see feeding into the `"=a"` operands in the bitcode? Is it `alloca`s?

lib/Arch/Runtime/HyperCall.cpp
pgoodman: From my memory, base address of the segment is "hidden." For practical purposes, we need to know it, though, as that hidden base is used in calculations what involve implicit or explicit segment register use.

```c++
WritePtr<addr_t>(next_sp _IF_32BIT(REG_SS_BASE))
```

That `WritePtr` is sort of a fancy cast, and conditional use of `ss_base` is saying "maybe displace by `ss_base`, i.e. `next_sp + ss_base`, so it's actually read in this instance.

I think I'd need to see the instructions that you're working with to put the picture together and contextualize this a bit more.

lib/Arch/Runtime/HyperCall.cpp
```diff
@@ -44,13 +44,13 @@ Memory *__remill_sync_hyper_call(State &state, Memory *mem,
     case SyncHyperCall::kX86LoadGlobalDescriptorTable:
       asm volatile("lgdt"
                    :
-                   : "=m"(__remill_read_memory_64(memory, state.addr_to_load)));
+                   : "=m"(__remill_read_memory_64(mem, state.addr_to_load)));
```
pgoodman, line 47: I think the `=` in `=m` here is misplaced. See this documentation:

image

From here: https://gcc.gnu.org/onlinedocs/gcc/Extended-Asm.html

It's likely that for the `lgdt`, you would want to declare a local variable, storing its value with the `__remill_read_memory_64`, then reference that variable from the inline asm.

lib/Arch/X86/Runtime/CMakeLists.txt
```diff
@@ -42,6 +42,7 @@ function(add_runtime_helper target_name address_bit_size enable_avx enable_avx51
     BCFLAGS "-std=${required_cpp_standard}"
     INCLUDEDIRECTORIES "${REMILL_INCLUDE_DIR}" "${REMILL_SOURCE_DIR}"
     INSTALLDESTINATION "${REMILL_INSTALL_SEMANTICS_DIR}"
+    TARGET_TRIPLE "x86_64-darwin-macho"
```
pgoodman, line 45: I think the triple you pass in should be a variable that has been set based on some conditions. Another possibility is that it could be possible to use a pretty generic triple.

pgoodman (changes requested):

lib/Arch/X86/Runtime/CMakeLists.txt
tetsuo-cpp: Yeah that's just temporary to get things building. I think I'm going to make it use `<arch>-linux-elf` or `<arch>-darwin-macho` depending on what the host machine is running.

lib/Arch/Runtime/HyperCall.cpp
tetsuo-cpp: [fcd24b2](https://github.com/lifting-bits/remill/pull/611/commits/fcd24b236ca908a41ebe253a849ba6fc738e48c9)

lib/Arch/X86/Runtime/CMakeLists.txt
tetsuo-cpp: I'm not sure that this cross compilation strategy really works. The runtime has too many references to system headers. For example, even the hyper call enumeration itself relies in `<cstdint>`. 

I was going to try to keep the operating system the same and then just substitute in a different arch. This is because Clang seems to use the host system's headers when cross compiling provided that I'm targeting the same operating system. For example, I'm on an M1 but if I cross compile for `x86_64-darwin-macho`, it uses my system headers. However, this doesn't work for architectures like SPARC because Clang obviously doesn't support `sparc-darwin-macho`. So I'm back to the drawing board.

In order for this to work, I'm going to need to build for `<arch>-none-eabi` and remove any references to system headers in the runtime. It might be best to work on top of that `mps430` branch you mentioned since I see you've already got a solution to the `<cstdint>` includes.

lib/Arch/Runtime/HyperCall.cpp
tetsuo-cpp: > I think I'd need to see the instructions that you're working with to put the picture together and contextualize this a bit more.

I'm not seeing these lifted in an example, I'm just going through each hyper call type and trying to implement it.

lib/Arch/X86/Runtime/CMakeLists.txt
pgoodman: OK! Give yourself a limit on how much time you'll spend on it, vs. when it's worth considering if we can get correct-ish results by just using the host architecture target triple (i.e. by not even specifying a triple).

lib/Arch/Runtime/HyperCall.cpp
```diff
@@ -42,15 +42,13 @@ Memory *__remill_sync_hyper_call(State &state, Memory *mem,
       break;
 
     case SyncHyperCall::kX86LoadGlobalDescriptorTable:
-      asm volatile("lgdt"
-                   :
-                   : "=m"(__remill_read_memory_64(mem, state.addr_to_load)));
+      auto read = __remill_read_memory_64(mem, state.addr_to_load);
```
pgoodman: I believe that you'll need parenthesis around the switch case, otherwise you'll get compiler warnings or possibly errors.

pgoodman (changes requested):

lib/Arch/Runtime/HyperCall.cpp
tetsuo-cpp: Do you mean braces around the `case`? It seems to be building fine.

tetsuo-cpp:
Ok, my approach has changed significantly since last time. In order to implement some of these hyper calls (`cpuid` for instance), the only realistic option is to invoke it via inline assembly and then get Clang to give us bitcode for it.

In order to do this, I now have to cross-compile this part of the runtime. Because I'm cross-compiling, I now don't have access to standard library headers which is why there's a lot of code under `lib/Arch/Runtime/` to fill in these gaps. I specifically have chosen to cross-compile just the hyper calls and *not* the entire runtime to reduce the amount of functionality we have to recreate. For example, the runtimes make use of various STL containers like `std::bitset` and stuff from `<algorithm>`. Our build is currently already invoking `llvm-link` to stitch together each of the bitcode files to make a final runtime bitcode file.

tetsuo-cpp:
Seems like I'll need to bring `Limits.h` across too judging from that CI failure. Not sure how it's building fine on my machine. Will look into it more tomorrow.

But the general approach will stay the same.

tetsuo-cpp:
`build_linux` is failing in `master` for the same reason. I think this is good to go.

include/remill/Arch/Runtime/Intrinsics.h
```diff
@@ -303,4 +305,111 @@ __remill_write_io_port_16(Memory *, addr_t, uint16_t);
...
+[[gnu::used, gnu::const]] extern Memory *__remill_x86_set_segment_ds(Memory *);
+
+[[gnu::used, gnu::const]] extern Memory *__remill_x86_set_segment_fs(Memory *);
+
+[[gnu::used, gnu::const]] extern Memory *__remill_x86_set_segment_gs(Memory *);
+
+[[gnu::used, gnu::const]] extern Memory *__remill_x86_set_debug_reg(Memory *);
+
+[[gnu::used, gnu::const]] extern Memory *
+__remill_x86_set_control_reg_0(Memory *);
```
pgoodman, line 322: Set to what?

include/remill/Arch/Runtime/Intrinsics.h
```diff
@@ -303,4 +305,111 @@ __remill_write_io_port_16(Memory *, addr_t, uint16_t);
 [[gnu::used, gnu::const]] extern Memory *
 __remill_write_io_port_32(Memory *, addr_t, uint32_t);
 
+// More specific hyper calls.
+[[gnu::used, gnu::const]] extern Memory *__remill_x86_set_segment_es(Memory *);
```
pgoodman, line 309: Set to what? Should there be a `uint16_t` param here? If changing the segment selector, maybe an output should be a `addr_t &` to overwrite the `*_BASE` reg value.

include/remill/Arch/Runtime/Intrinsics.h
```diff
@@ -303,4 +305,111 @@ __remill_write_io_port_16(Memory *, addr_t, uint16_t);
...
+[[gnu::used, gnu::const]] extern Memory *
+__remill_sparc_unimplemented_instruction(Memory *);
+
+[[gnu::used, gnu::const]] extern Memory *
+__remill_sparc_unhandled_dcti(Memory *);
+
+[[gnu::used, gnu::const]] extern Memory *
+__remill_sparc_window_underflow(Memory *);
+
+[[gnu::used, gnu::const]] extern Memory *__remill_sparc_trap_cond_a(Memory *);
```
pgoodman, line 377: I think a `bool` or something like it should probably be passed in here as another parameter. It feels like the trapping conditions themselves could be elided, or ordered before/after this intrinsic call.

pgoodman (changes requested):

tetsuo-cpp:
Something is wrong with the `llvm-link` step. The `__remill_sync_hyper_call` implementation isn't getting picked up for some reason. I'm in the middle of debugging this.

include/remill/Arch/Runtime/Intrinsics.h
```diff
@@ -137,8 +137,8 @@ __remill_flag_computation_carry(bool result, ...);
 [[gnu::used]] extern Memory *__remill_async_hyper_call(State &, addr_t ret_addr,
                                                        Memory *);
 
-[[gnu::used]] extern Memory *__remill_sync_hyper_call(State &, Memory *,
-                                                      SyncHyperCall::Name);
+[[gnu::used, gnu::always_inline]] extern Memory *
```
tetsuo-cpp, line 140: This tag is important. If we don't put it here, the runtime will implicitly have this function marked with `noinline` as a LLVM function attribute. The function will not get inlined and then when we move the functions across to the target module, we will not copy across the `__remill_sync_hyper_call` definition and it will be lost.

include/remill/Arch/Runtime/Int.h
```diff
@@ -0,0 +1,101 @@
...
+typedef unsigned uint128_t __attribute__((mode(TI)));
+typedef int int128_t __attribute__((mode(TI)));
+#elif defined(__aarch64__)
+typedef __uint128_t uint128_t;
+typedef __int128_t int128_t;
+#elif defined(__sparc__)
+typedef __uint128_t uint128_t;
+typedef __int128_t int128_t;
+#elif defined(__is_identifier) && __is_identifier(_ExtInt)
+typedef unsigned _ExtInt(128) uint128_t;
```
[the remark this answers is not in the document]
tetsuo-cpp: [e49e709](https://github.com/lifting-bits/remill/pull/611/commits/e49e709d877c98d0e092765e3959d37827c0acdb)

lib/Arch/Runtime/HyperCall.cpp
```diff
@@ -0,0 +1,310 @@
...
+
+    case SyncHyperCall::kSPARC64EmulateInstruction:
+      mem = __remill_sparc64_emulate_instruction(mem);
+      break;
+
+#  endif
+
+#endif
+
+    default: break;
```
tetsuo-cpp, line 306: Unfortunately, I can't `abort` here anymore since I don't have access to the standard library.

include/remill/Arch/Runtime/Intrinsics.h
tetsuo-cpp: I don't think there's enough information in `__remill_sync_hyper_call` to figure this out.

https://github.com/lifting-bits/remill/blob/master/lib/Arch/X86/Semantics/SYSTEM.cpp#L64-L77

include/remill/Arch/Runtime/Intrinsics.h
tetsuo-cpp: Same thing here. I don't think the information is there in the hyper call.

https://github.com/lifting-bits/remill/blob/master/lib/Arch/X86/Semantics/DATAXFER.cpp#L664-L669

I think we can refactor it but, given that this information is already missing from the hyper call, can we follow up with this later?

include/remill/Arch/Runtime/Intrinsics.h
tetsuo-cpp: Perhaps I can pass in the value of `HYPER_CALL_VECTOR`?

https://github.com/lifting-bits/remill/blob/master/lib/Arch/SPARC32/Semantics/TRAP.cpp#L32-L38

It's a `uint32_t` listed [here](https://github.com/lifting-bits/remill/blob/master/include/remill/Arch/Runtime/State.h#L34).

include/remill/Arch/Runtime/Intrinsics.h
pgoodman: Yes we can follow-this up later. I think a later follow-up should perhaps get rid of the hypercall, and instead use this directly in the semantics.

include/remill/Arch/Runtime/Intrinsics.h
pgoodman: That is.. subtle. Can you add it as a comment in the code?

pgoodman:
I agree about moving some stuff to a future PR. I think that you've rightly found that the pre-existing hypercall mechanism isn't flexible enough, and that where added flexibility is needed, new hyper call intrinsics should be introduced. You've done this, e.g. with setting segment registers, but then they are being called inside `__remill_sync_hyper_call`, and a future PR should instead just call those new intrinsics in the semantics rather than indirecting through `__remill_sync_hyper_call`. Realistically, maybe all of the enums should go, and then we'd just have a bunch of different intrinsic calls. Again, separate PR for that.

lib/Arch/Runtime/HyperCall.cpp
tetsuo-cpp: [b93978a](https://github.com/lifting-bits/remill/pull/611/commits/b93978abd9729b5402e677c0f411a459d2a6a054)

include/remill/Arch/Runtime/Intrinsics.h
tetsuo-cpp: [0574f2a](https://github.com/lifting-bits/remill/pull/611/commits/0574f2a4e1994026c8bf3ce58241ff360e19652c)

include/remill/Arch/Runtime/Intrinsics.h
pgoodman: Oh indeed! I was thinking this must be some kind of thing like "division by zero" with data-dependant conditions.

pgoodman (approved):

PR #612, opened 2022-07-28
Prevent the instruction category from being overwritten after a fused call/pop has been recognised

tetsuo-cpp:


lib/Arch/X86/Arch.cpp
```diff
@@ -1230,7 +1230,9 @@ bool X86Arch::DecodeInstruction(uint64_t address, std::string_view inst_bytes,
     }
   }
 
-  inst.category = CreateCategory(xedd);
+  if (inst.category == Instruction::kCategoryInvalid) {
```
tetsuo-cpp, line 1233: The gist is that our call/pop fusing logic above will set the instruction category. So we want to do a check here to make sure that we don't overwrite it.

pgoodman: Maybe most the original line up just before line 1206? Then we don't need the extra conditional.

pgoodman (changes requested):

lib/Arch/X86/Arch.cpp
tetsuo-cpp: [209db3b](https://github.com/lifting-bits/remill/pull/612/commits/209db3b7e3f03d134dcbe054a95a0bc6be48382f)

pgoodman (approved):

PR #614, opened 2022-08-08
Fix typo when parsing z3 version from headers

Ninja3047:
should be Z3_BUILD_NUMBER instead of Z3_BUILD_VERSION
see: https://github.com/llvm/llvm-project/blob/dc07d2c91dfd7f14991d07ef1df5f1f0b19b306d/llvm/cmake/modules/FindZ3.cmake#L103


2over12 (approved):
LGTM!

PR #617, opened 2022-08-15
Add Context Structure to Affect State Dependent Liftings

2over12:


include/remill/Arch/Arch.h
```diff
@@ -281,14 +285,16 @@ class Arch {
   //            walk up, one byte at a time, to `MaxInstructionSize(false)`
   //            bytes being passed to the decoder, until you successfully decode
   //            or ultimately fail.
-  virtual bool DecodeInstruction(uint64_t address, std::string_view instr_bytes,
-                                 Instruction &inst) const = 0;
+  virtual std::optional<DecodingContext::ContextMap>
```
lkorenc, line 288: Introduce type alias?

include/remill/Arch/Arch.h
```diff
@@ -281,14 +285,16 @@ class Arch {
...
+  DecodeInstruction(uint64_t address, std::string_view instr_bytes,
+                    Instruction &inst, DecodingContext context) const = 0;
```
lkorenc, line 290: `context` is taken by value - is that an intentional choice and we expect to always take care of ownership? Or do we expect to mutate it? (Just trying to understand if `const &` is not a better choice).

include/remill/Arch/Arch.h
```diff
@@ -281,14 +285,16 @@ class Arch {
...
 
   // Decode an instruction that is within a delay slot.
-  bool DecodeDelayedInstruction(uint64_t address, std::string_view instr_bytes,
-                                Instruction &inst) const {
+  std::optional<DecodingContext::ContextMap>
+  DecodeDelayedInstruction(uint64_t address, std::string_view instr_bytes,
+                           Instruction &inst, DecodingContext context) const {
     inst.in_delay_slot = true;
-    return this->DecodeInstruction(address, instr_bytes, inst);
+    return this->DecodeInstruction(address, instr_bytes, inst, context);
```
lkorenc, line 297: Copy on `context`, move it.

include/remill/Arch/ArchBase.h
```diff
@@ -31,13 +31,27 @@ namespace remill {
 
 struct Register;
 
-// Internal base architecture for all Remill-internal architectures.
-class ArchBase : public remill::Arch {
- protected:
+
+/// A simple mixin for a default decoder with a null context and the default insdtruction lifter;
+class DefaultContextAndLifter : virtual public remill::Arch {
```
lkorenc, line 36: `virtual` inheritance is a big hammer we may not need here - think how to make it without.

include/remill/Arch/Context.h
```diff
@@ -0,0 +1,49 @@
...
+
+  DecodingContext() = default;
+
+  DecodingContext(std::unordered_map<std::string, uint64_t> context_value);
+
+
+  uint64_t GetContextValue(const std::string &context_reg) const;
+  DecodingContext PutContextReg(std::string creg, uint64_t value) const;
+
+  static ContextMap UniformContextMapping(DecodingContext cst);
```
lkorenc, line 46: What is the advantage of using this compared to simply passing the "raw" `unordered_map` around?

lib/Arch/Context.cpp
```diff
@@ -0,0 +1,36 @@
...
+DecodingContext DecodingContext::PutContextReg(std::string creg,
+                                               uint64_t value) const {
+  std::unordered_map<std::string, uint64_t> new_value(this->context_value);
+  new_value.emplace(creg, value);
+  return DecodingContext(std::move(new_value));
+}
+
+DecodingContext::ContextMap
+DecodingContext::UniformContextMapping(DecodingContext cst) {
+  return [cst](uint64_t ctx) -> DecodingContext { return cst; };
```
lkorenc, line 32: `ctx` is simply ignored?

lib/BC/TraceLifter.cpp
```diff
@@ -345,7 +345,9 @@ bool TraceLifter::Impl::Lift(
 
       inst.Reset();
 
-      (void) arch->DecodeInstruction(inst_addr, inst_bytes, inst);
+      // TODO(Ian): not passing context around in trace lifter
+      (void) arch->DecodeInstruction(inst_addr, inst_bytes, inst,
```
lkorenc, line 349: `std::ignore = value_dont_care_about()`

lkorenc (changes requested):

lib/Arch/Context.cpp
lkorenc: you are doing two copies, move `cst` into lambda.

lib/Arch/Arch.cpp
```diff
@@ -850,16 +850,30 @@ const IntrinsicTable *ArchBase::GetInstrinsicTable(void) const {
...
+std::optional<DecodingContext::ContextMap>
+DefaultContextAndLifter::DecodeInstruction(uint64_t address,
+                                           std::string_view instr_bytes,
+                                           Instruction &inst,
+                                           DecodingContext context) const {
   inst.SetLifter(std::make_unique<remill::InstructionLifter>(
       this, this->GetInstrinsicTable()));
-  return this->ArchDecodeInstruction(address, instr_bytes, inst);
+  if (this->ArchDecodeInstruction(address, instr_bytes, inst)) {
+    return [](uint64_t x) -> DecodingContext { return DecodingContext(); };
```
lkorenc, line 866: Drop `x` - unused arguments do not need names.

lib/Arch/Sleigh/Arch.cpp
```diff
@@ -366,28 +366,35 @@ std::string CustomLoadImage::getArchType(void) const {
...
+                              Instruction &inst,
+                              DecodingContext context) const {
   inst.SetLifter(
       std::make_shared<SleighLifter>(this, *this->GetInstrinsicTable()));
   assert(inst.GetLifter() != nullptr);
 
-  return this->ArchDecodeInstruction(address, instr_bytes, inst);
+  if (const_cast<SleighArch *>(this)->DecodeInstructionImpl(
+          address, instr_bytes, inst)) {
+    return [this](uint64_t x) -> DecodingContext {
```
lkorenc, line 379: Drop name of `x`.

include/remill/Arch/Context.h
2over12: The context map? the lambda allows the decoders to encode arbitrary computations over successor addresses

lib/Arch/Arch.cpp
```diff
@@ -863,7 +863,7 @@ DefaultContextAndLifter::DecodeInstruction(uint64_t address,
   inst.SetLifter(std::make_unique<remill::InstructionLifter>(
       this, this->GetInstrinsicTable()));
   if (this->ArchDecodeInstruction(address, instr_bytes, inst)) {
-    return [](uint64_t x) -> DecodingContext { return DecodingContext(); };
+    return [](uint64_t _) -> DecodingContext { return DecodingContext(); };
```
lkorenc, line 864: Just `[](uint64_t) -> DecodingContext { ... }`.
No need for `_`. (In other changes as well)

As a sidenote, `-> DecodingContext` can also be dropped, I think compiler should be able to deduce this, but I can understand if you want to keep it as part of docs.

lkorenc (changes requested):

include/remill/Arch/ArchBase.h
```diff
@@ -122,4 +105,27 @@ class ArchBase : virtual public remill::Arch {
   mutable std::unique_ptr<IntrinsicTable> instrinsics{nullptr};
 };
 
+class DefaultContextAndLifter : virtual public remill::ArchBase {
```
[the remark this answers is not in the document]
2over12: So this change pushed down the virtual inheritance by one but there is still a diamond problem here due to shared structure in the AArch32 and x86 arches. ie. we have AArch32Base -> ArchBase, then we have Aarch32 (the non sleigh version) -> DefaultContextAndLifter and Aarch32 -> ArchBase, hence the virtual inheritance.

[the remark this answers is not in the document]
2over12: The core problem I guess with those setups is the base class needs access to ArchBase that the sleigh arches are then going to derive from directly while another child needs the DefaultContextAndLifter which has to have access to ArchBase as well

lkorenc (approved):

PR #620, opened 2022-08-20
Fix Type Info for Register Variable GEPs

### LINKED ISSUE #618: Opaque Pointers Lead to Incorrect Register Sizes by Pulling a Type from `getResultElementType` for a Register that is not in the Operand Cache but is Declared as a Variable in the Lifted Function
This issue is exacerbated by #615 since instructions get individual lifters to lift themselves, not sharing a cache (this probably should be addressed somehow, but cache consistency becomes tricky), causing register cache misses to occur more frequently. If we have a cache miss for lifting a register operand that is, however, found in the function as a GEP we pull a type from getResultElementType but this type does not match the desired type for the register. For instance, if we GEP a 32 bit register W0 then call getResultElementType on the GEP we are going to get a 64 bit type because of the size of the underlying field in the state structure. This line https://github.com/lifting-bits/remill/blob/c0f90b9e4a251615288ed4b65ead81cea617da8c/lib/BC/Util.cpp#L220 will not pull the correct type, we need to source the type info from the Arch's knowledge about the reg name being asked for and not from the state structure.

Regression test here: https://github.com/lifting-bits/remill/blob/c829715b95a69401a0f439b8004962a3dda6e803/tests/Thumb/TestLifting.cpp#L236

```
TEST(RegressionTests, AARCH64RegSize) {
  llvm::LLVMContext context;
  context.enableOpaquePointers();
  auto arch = remill::Arch::Build(&context, remill::OSName::kOSLinux,
                                  remill::ArchName::kArchAArch64LittleEndian);
  auto sems = remill::LoadArchSemantics(arch.get());
  remill::IntrinsicTable instrinsics(sems.get());
  auto op_lifter = arch->DefaultLifter(instrinsics);
  auto target_lift = arch->DefineLiftedFunction("test_lift", sems.get());
  auto st_ptr = remill::LoadStatePointer(target_lift);
  CHECK_NOTNULL(st_ptr);
  auto lifted =
      op_lifter->LoadRegValue(&target_lift->getEntryBlock(), st_ptr, "W0");

  CHECK_EQ(lifted->getType()->getIntegerBitWidth(), 32);
  op_lifter->ClearCache();
  auto lifted2 =
      op_lifter->LoadRegValue(&target_lift->getEntryBlock(), st_ptr, "W0");

  CHECK_EQ(lifted2->getType()->getIntegerBitWidth(), 32);
}
```
This test fails on the second check because the variable finding code returns a 64 bit size type for W0 on the second call.

2over12:
This PR closes #618 by returning the type from the arches reg info rather than the inferred type of the GEP.

### DIFF
--- a/lib/Arch/Arch.cpp
+++ b/lib/Arch/Arch.cpp
@@ -311,6 +311,7 @@ void ArchBase::ForEachRegister(std::function<void(const Register *)> cb) const {
 // Return information about a register, given its name.
 const Register *ArchBase::RegisterByName(std::string_view name_) const {
   std::string name(name_.data(), name_.size());
+
   auto [curr_val_it, added] = reg_by_name.emplace(std::move(name), nullptr);
   if (added) {
     return nullptr;
--- a/lib/BC/InstructionLifter.cpp
+++ b/lib/BC/InstructionLifter.cpp
@@ -87,7 +87,6 @@ LiftStatus InstructionLifter::LiftIntoBlock(Instruction &arch_inst,
                                             llvm::BasicBlock *block,
                                             llvm::Value *state_ptr,
                                             bool is_delayed) {
-  LOG(INFO) << "In instructionlifter";
   llvm::Function *const func = block->getParent();
   llvm::Module *const module = func->getParent();
   llvm::Function *isel_func = nullptr;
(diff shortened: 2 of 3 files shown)


pgoodman (approved):

2over12:
Hmm looks like this approach introduces a bug... not sure will have to look into it

PR #622, opened 2022-08-23
Mixed Mode AArch32 

2over12:
This PR migrates AArch32 to support a mixed mode between thumb and normal aarch32 by using the context object to mode switch.


2over12:
looks like the dockerfile doesnt like how sleigh is being installed, need to look into this

sschriner (commented):
This LGTM from an AARCH32 perspective

include/remill/Arch/Context.h
```diff
@@ -38,15 +39,23 @@ class DecodingContext {
  public:
   using ContextMap = std::function<DecodingContext(uint64_t)>;
```
pgoodman, line 40: Is it desirable to have some kind of enumeration tag as a second argument, e.g to advise on the meaning of the address? Is it possible to have a conditional branch to the fall-through address, where the taken path changes the decoding mode?

pgoodman: I think I generally would want to know why something more like an API isn't desirable. Something like of like: you can't even ask remill for the branch taken pc. Instead, you can ask it for the branch taken target, and it will give you back variant that is one of a (decoding context, pc), an indirect target tag, or not a target tag.

2over12: So my explanation for why this was important is anvill knows the indirect target of indirect jump and remill knows how to update the context given address

2over12: The case of conditional direct branch to the same fallthrough address with a mode change is possible but I don't know why we would handle it tbh. That would imply intentionally setting up an arm/thumb polyglot at some address for that to work. That being said given that the function didn't turn out how I hoped (ie. we cant just pass the branch target to resolve mode switches since it has to be in terms of the value prior to the mask) the lambda probably isn't strictly required. Honestly, I am not sure that we want to keep instruction categories at all. 

Instruction categories often fall over, granted we need them as hints for interprocedural flows, but I'd rather just return constant successors, handles into llvm to determine when those successors are reached, and contexts for them. Part of me wants to say that indirect contexts should be resolved by interpreting the llvm

2over12: ie. realistically what I want is remill is it returns a [(pc, context, Optional<reaching_llvm_exression>)] where either the program counter or the context can be non-constant, in which case remill gives back an expression in terms of some members of the cpu state, that if resolved could be evaluated to the next pc and next context.

2over12: The short answer is because I misevaluated and thought it would be possible to provide an expression from address->context for an arbitrary indirect target.

lib/Arch/Context.cpp
```diff
@@ -17,19 +17,39 @@ DecodingContext::GetContextValue(const std::string &context_reg) const {
...
+  this->context_value.erase(creg);
+}
+
+bool DecodingContext::HasContextValue(const std::string &creg) {
+  return this->context_value.find(creg) != this->context_value.end();
+}
+
+
+DecodingContext
+DecodingContext::MakeContextRegNonConstant(const std::string &creg) const {
```
lkorenc, line 44: This name is too long and also does not really say much about what is happening inside.
`CtxWithoutReg( ... )`? Or `CopyWithoutReg( ... )` - it is a method so reader already knows it operates on context.

lib/Arch/Context.cpp
```diff
@@ -17,19 +17,39 @@ DecodingContext::GetContextValue(const std::string &context_reg) const {
...
 
+void DecodingContext::UpdateContextReg(std::string creg, uint64_t value) {
+  this->context_value[creg] = value;
+}
+
+void DecodingContext::DropReg(const std::string &creg) {
+  this->context_value.erase(creg);
+}
+
+bool DecodingContext::HasContextValue(const std::string &creg) {
```
lkorenc, line 38: Why not just `HasValue`?

lib/Arch/Instruction.cpp
```diff
@@ -673,9 +673,10 @@ std::string Instruction::Serialize(void) const {
   };
 
   auto maybe_stream_branch_taken_arch = [this, &ss, &stream_arch]() {
-    if (branch_taken_arch_name != arch_name) {
+    if (branch_taken_arch_name.has_value() &&
```
lkorenc, line 676: Drop `has_value()`.

lib/Arch/Sleigh/Arch.cpp
```diff
@@ -366,41 +372,49 @@ std::string CustomLoadImage::getArchType(void) const {
...
+
+  return this->lifter;
 }
 
+Arch::DecodingResult
+SleighDecoder::DecodeInstruction(uint64_t address, std::string_view instr_bytes,
+                                 Instruction &inst,
+                                 DecodingContext context) const {
+  inst.SetLifter(this->GetLifter());
+  assert(inst.GetLifter() != nullptr);
```
lkorenc, line 398: `CHECK`

lib/Arch/Sleigh/Arch.cpp
```diff
@@ -519,9 +538,56 @@ std::optional<int32_t> SingleInstructionSleighContext::oneInstruction(
...
+      this->engine.getRegisterName(outvar->space, outvar->offset, outvar->size);
+  auto target_remill_cont_reg = this->register_mapping.find(outvar_name);
+  if (target_remill_cont_reg == this->register_mapping.end()) {
+    return;
+  }
+
+  if (this->already_assigned.find(target_remill_cont_reg->second) !=
+          this->already_assigned.end() ||
+      !(opc == OpCode::CPUI_COPY && isVarnodeInConstantSpace(vars[0]))) {
+    LOG(INFO) << "Dropping " << target_remill_cont_reg->second;
```
lkorenc, line 573: `DLOG` - sounds like something that should be printed when debugging rather normally.

lib/Arch/Sleigh/Arch.cpp
```diff
@@ -519,9 +538,56 @@ std::optional<int32_t> SingleInstructionSleighContext::oneInstruction(
...
+
+
+  this->curr_context.UpdateContextReg(target_remill_cont_reg->second,
+                                      vars[0].offset);
+
+  return;
+}
+
+DecodingContext ContextUpdater::GetContext() const {
+  LOG(INFO) << "Getting context updater value";
```
lkorenc, line 586: `DLOG`

lib/Arch/Sleigh/X86Arch.cpp
```diff
@@ -32,22 +32,49 @@
...
+class SleighX86Decoder final : public SleighDecoder {
+ public:
+  SleighX86Decoder() = delete;
+  SleighX86Decoder(const remill::Arch &arch)
+      : SleighDecoder(
+            arch, kArchX86_SLEIGH == arch.arch_name ? "x86.sla" : "x86-64.sla",
+            kArchX86_SLEIGH == arch.arch_name ? "x86.pspec" : "x86-64.pspec",
+            {}) {}
+
+  virtual ~SleighX86Decoder(void) = default;
```
lkorenc, line 44: Why is this needed?

lib/Arch/Sleigh/X86Arch.cpp
```diff
@@ -32,22 +32,49 @@
...
+
+  void InitializeSleighContext(
```
lkorenc, line 46: Add comment why it is not ok to just use the parent impl but instead an empty body is specifically defined (at least from the PR I did not deduce it in reasonable time).

lib/Arch/Sleigh/X86Arch.cpp
```diff
@@ -32,22 +32,49 @@
 namespace remill {
 namespace sleigh::x86 {
 
-class SleighX86Arch final : public SleighArch, public remill::X86ArchBase {
+class SleighX86Decoder final : public SleighDecoder {
+ public:
+  SleighX86Decoder() = delete;
+  SleighX86Decoder(const remill::Arch &arch)
+      : SleighDecoder(
+            arch, kArchX86_SLEIGH == arch.arch_name ? "x86.sla" : "x86-64.sla",
```
lkorenc, line 40: static methods (maybe in parent?)/standalone functions for those ternaries? Sound useful in some other places.

lib/BC/SleighLifter.cpp
```diff
@@ -714,28 +719,15 @@ class SleighLifter::PcodeToLLVMEmitIntoBlock : public PcodeEmit {
...
-      }
-      default: return std::nullopt;
-    }
-  }
   LiftStatus LiftBoolBinOp(llvm::IRBuilder<> &bldr, OpCode opc,
                            VarnodeData *outvar, VarnodeData lhs,
                            VarnodeData rhs) {
 
 
+    if (this->BOOL_BINARY_OPS.find(opc) == this->BOOL_BINARY_OPS.end()) {
```
lkorenc, line 727: `if (is_supported(opc))` ? I assume all `OPS` collections have unique values wrt to each other? Then one method would nicely provide this functionality to all of them.

lib/BC/SleighLifter.cpp
```diff
@@ -909,6 +897,7 @@ class SleighLifter::PcodeToLLVMEmitIntoBlock : public PcodeEmit {
           bldr, lhs, llvm::IntegerType::get(this->context, lhs.size * 8));
 
       if (lifted_lhs.has_value()) {
+        LOG(INFO) << "SUBPIECE: " << remill::LLVMThingToString(*lifted_lhs);
```
lkorenc, line 900: `DLOG`

lib/BC/SleighLifter.cpp
```diff
@@ -1051,6 +1040,7 @@ class SleighLifter::PcodeToLLVMEmitIntoBlock : public PcodeEmit {
   virtual void dump(const Address &addr, OpCode opc, VarnodeData *outvar,
                     VarnodeData *vars, int4 isize) final override {
     LOG(INFO) << "inner handle" << std::endl;
+
```
lkorenc, line 1043: `DLOG`

lib/BC/SleighLifter.cpp
```diff
@@ -1101,6 +1091,20 @@ std::unordered_set<OpCode>
         CPUI_INT_SLESS,   CPUI_INT_LESSEQUAL, CPUI_INT_SLESSEQUAL,
         CPUI_INT_SBORROW, CPUI_INT_SCARRY,    CPUI_INT_CARRY};
 
+std::map<OpCode, SleighLifter::PcodeToLLVMEmitIntoBlock::BinaryOperator>
+    SleighLifter::PcodeToLLVMEmitIntoBlock::BOOL_BINARY_OPS = {
+        {OpCode::CPUI_BOOL_AND,
+         [](llvm::Value *lhs, llvm::Value *rhs, llvm::IRBuilder<> &bldr) {
```
lkorenc, line 1097: What is an upside of `std::map` (or `std::unordered_map`) compared to a function with a `switch`?

tests/Thumb/TestLifting.cpp
```diff
@@ -50,7 +51,27 @@ const static std::unordered_map<std::string,
...
+  auto arch = remill::Arch::Build(&context, remill::OSName::kOSLinux,
+                                  remill::ArchName::kArchAArch32LittleEndian);
+  auto sems = remill::LoadArchSemantics(arch.get());
+
+
+  remill::DecodingContext dec_context;
+  dec_context.UpdateContextReg(remill::kThumbModeRegName, tm_val);
+  assert(dec_context.HasContextValue("TMReg"));
+  remill::Instruction insn;
+  auto res = arch->DecodeInstruction(address, bytes, insn, dec_context);
```
lkorenc, line 70: Direct return.

lkorenc (changes requested):

2over12:
This is getting reworked substantially. Most of these reviews won't be applicable

lib/BC/SleighLifter.cpp
2over12: This is fixing an ordering issue where in the bool function handling it would attempt to lift a value as a bool before determining if the bool ops apply. Instead now we check if we have a supported lifting operation, then lift the ops, then apply the function. This avoids having a value lifted as a boolean when it potentially should not be lifted as a boolean. An alternative is to have a list of BOOL_OPs that are checked to see if we need to do a bool lift, then lift the operand, then switch on the bool_ops again deciding what to do with that bool

2over12: This is all pretty subtle and we should figure out a way to rework the lifter so bugs like this dont happen

include/remill/Arch/X86/X86Base.h
```diff
@@ -18,27 +18,31 @@ class X86ArchBase : public virtual ArchBase {
...
+  uint64_t MinInstructionAlign(const DecodingContext &context) const override;
 
-  uint64_t MinInstructionSize(void) const;
 
-  uint64_t MaxInstructionSize(bool) const;
-  llvm::CallingConv::ID DefaultCallingConv(void) const;
+  uint64_t MinInstructionSize(const DecodingContext &context) const override;
 
-  llvm::DataLayout DataLayout(void) const;
+  uint64_t MaxInstructionSize(const DecodingContext &, bool) const override;
```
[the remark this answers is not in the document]
2over12: agreed

include/remill/Arch/Instruction.h
```diff
@@ -254,6 +257,148 @@ class Instruction {
     kCategoryConditionalAsyncHyperCall,
   } category;
 
+
+  class Flow {};
```
lkorenc, line 261: Make them structs - they are really POD +- some aux methdos.

lib/Arch/AArch32/ArchContext.cpp
```diff
@@ -0,0 +1,10 @@
+#include <remill/Arch/AArch32/ArchContext.h>
+
+namespace remill {
+const std::string_view kThumbModeRegName = "TMReg";
```
lkorenc, line 4: Any special reason why these are not declared in header? Because it will improve both readability and optimisations.

lib/Arch/Instruction.cpp
```diff
@@ -801,4 +803,49 @@ void Instruction::SetLifter(InstructionLifter::LifterPtr lifter_) {
   lifter.swap(lifter_);
 }
 
+Instruction::DirectFlow::DirectFlow(uint64_t known_target_,
```
lkorenc, line 806: Move all of these to header - there is no point in hiding them in source file.

lib/Arch/Sleigh/ControlFlowStructuring.cpp
```diff
@@ -0,0 +1,580 @@
...
+  for (size_t i = 0; i < op.vars.size(); ++i) {
+    auto iv = op.vars[i];
+    print_vardata(engine, ss, iv);
+  }
+
+  return ss.str();
+}*/
+}  // namespace
+
+bool Instruction::DirectJump::operator==(const DirectJump &rhs) const {
```
lkorenc, line 40: Move all of these to `Instruction.cpp`

include/remill/Arch/Instruction.h
```diff
@@ -254,6 +257,148 @@ class Instruction {
...
+    // We may have info in the decoder that tells us a context value
+    std::optional<DecodingContext> maybe_context;
+
+    IndirectFlow() = delete;
+    IndirectFlow(std::optional<DecodingContext> maybe_context);
+
+    bool operator==(const IndirectFlow &rhs) const;
+  };
+
+  class FallthroughFlow : Flow {
```
lkorenc, line 285: Inconsistent orderding in classes - choose one way - attrs, ctors, methods or something else but keep consistent.

include/remill/Arch/Instruction.h
```diff
@@ -254,6 +257,148 @@ class Instruction {
...
+  class NormalInsn {
+   public:
+    NormalInsn() = delete;
+    NormalInsn(FallthroughFlow fallthrough);
+
+    FallthroughFlow fallthrough;
+    bool operator==(const NormalInsn &rhs) const;
+  };
+
+  class NoOp {
```
lkorenc, line 304: `NoOp` is copy and paste of `NormalInsn`

include/remill/Arch/Instruction.h
```diff
@@ -254,6 +257,148 @@ class Instruction {
...
+
+    bool operator==(const IndirectFlow &rhs) const;
+  };
+
+  class FallthroughFlow : Flow {
+   public:
+    FallthroughFlow(DecodingContext fallthrough_context);
+    FallthroughFlow() = delete;
+
+    DecodingContext fallthrough_context;
```
lkorenc, line 290: We should know the next address right? Why is it not similar to `DirectFlow`?

include/remill/Arch/Instruction.h
```diff
@@ -254,6 +257,148 @@ class Instruction {
...
+    DirectFunctionCall(DirectFlow taken_flow);
+
+    bool operator==(const DirectFunctionCall &rhs) const;
+  };
+
+
+  class IndirectFunctionCall : IndirectJump {
+   public:
+    IndirectFunctionCall() = delete;
+    IndirectFunctionCall(IndirectFlow id_flow);
```
lkorenc, line 359: Copy & paste.
`using IndirectJump::IndirectJump;`

include/remill/Arch/Instruction.h
```diff
@@ -254,6 +257,148 @@ class Instruction {
...
+
+    bool operator==(const IndirectFunctionCall &rhs) const;
+  };
+
+  class FunctionReturn : IndirectJump {
+   public:
+    FunctionReturn() = delete;
+    FunctionReturn(IndirectFlow id_flow);
```
lkorenc, line 367: Copy & paste.

include/remill/Arch/Instruction.h
```diff
@@ -254,6 +257,148 @@ class Instruction {
...
+      std::variant<DirectFunctionCall, IndirectFunctionCall, FunctionReturn,
+                   AsyncHyperCall, IndirectJump, DirectJump>;
+
+  class ConditionalInstruction {
+   public:
+    AbnormalFlow taken_branch;
+    FallthroughFlow fall_through;
+
+    ConditionalInstruction() = delete;
+    ConditionalInstruction(AbnormalFlow taken_branch,
```
lkorenc, line 388: Make way to consturct this even from an element type of `AbnormalFlow`.

include/remill/Arch/Instruction.h
2over12: Fallthroughs "must" go to next_pc which is defined as the next instruction, otherwise it's not a fallthrough

lib/BC/SleighLifter.cpp
2over12: Well we need to know if we want to try to lift the in param etc as a bool no?

include/remill/Arch/Instruction.h
2over12: I dont understand what you are suggesting here?

include/remill/Arch/Instruction.h
2over12: Not sure what this means. Since abnormal flow is a variant we can leverage conversions to just pass an indirectjump like ConditionalInstruction(IndirectJump(

lib/Arch/Sleigh/Thumb2Arch.cpp
```diff
@@ -24,54 +25,78 @@
...
+
+  virtual bool DecodeInstruction(uint64_t address, std::string_view instr_bytes,
+                                 Instruction &inst,
+                                 DecodingContext context) const override {
+    return decoder.DecodeInstruction(address, instr_bytes, inst, context);
   }
 
-  uint64_t MinInstructionAlign(void) const final {
+
+  // TODO(pag): Eventually handle Thumb2 and unaligned addresses.
```
sschriner, line 73: @2over12 Do we handle unaligned addresses now?

include/remill/Arch/AArch32/ArchContext.h
```diff
@@ -0,0 +1,15 @@
+#pragma once
+
+#include <remill/Arch/Context.h>
+
+#include <string>
+
+namespace remill {
+inline const std::string_view kThumbModeRegName = "TMReg";
```
sschriner, line 8: Why did you not add this to the state structure?

lib/Arch/AArch32/Decode.cpp
```diff
@@ -2756,12 +2813,24 @@ static bool TryBranchImm(Instruction &inst, uint32_t bits) {
...
     inst.category = Instruction::kCategoryConditionalBranch;
+    inst.flows = Instruction::ConditionalInstruction(
+        Instruction::DirectJump(
+            Instruction::DirectFlow(inst.branch_taken_pc, taken_context)),
+        Instruction::FallthroughFlow(kARMContext));
   } else if (is_func) {
     inst.category = Instruction::kCategoryDirectFunctionCall;
+    inst.flows = Instruction::DirectFunctionCall(
+        Instruction::DirectFlow(inst.branch_taken_pc, taken_context));
   } else {
```
lkorenc, line 2830: These 2 lines are copied 3 times here, extract away into a lambda?

lib/Arch/Context.cpp
```diff
@@ -4,33 +4,50 @@
 
 namespace remill {
 
-DecodingContext::DecodingContext(
-    std::unordered_map<std::string, uint64_t> context_value)
+
+bool DecodingContext::operator==(remill::DecodingContext const &rhs) const {
```
lkorenc, line 8: Is this not the same as default? If not maybe comment why we can ignore the rest?

lib/Arch/Instruction.cpp
```diff
@@ -801,4 +802,120 @@ void Instruction::SetLifter(InstructionLifter::LifterPtr lifter_) {
...
+
+Instruction::FunctionReturn::FunctionReturn(IndirectFlow id_flow_)
+    : IndirectJump(std::move(id_flow_)) {}
+
+Instruction::ConditionalInstruction::ConditionalInstruction(
+    AbnormalFlow taken_branch_, FallthroughFlow fall_through_)
+    : taken_branch(std::move(taken_branch_)),
+      fall_through(std::move(fall_through_)) {}
+
+bool Instruction::DirectJump::operator==(const DirectJump &rhs) const {
```
lkorenc, line 850: Cannot all of this be just `= default`? If not probably needs a comment.

lib/Arch/Sleigh/Arch.cpp
```diff
@@ -43,11 +43,11 @@ class AssemblyLogger : public AssemblyEmit {
               << " " << body;
   }
 };
+
+
 }  // namespace
```
lkorenc, line 48: Drop?

lib/Arch/Sleigh/Arch.cpp
```diff
@@ -365,90 +165,138 @@ std::string CustomLoadImage::getArchType(void) const {
...
-                                       std::string_view instr_bytes,
-                                       Instruction &inst) {
+const std::unordered_map<std::string, std::string> &
+SleighDecoder::GetRegisterMapping() const {
+  return this->register_mapping;
+}
+
+std::optional<std::pair<Instruction::InstructionFlowCategory,
+                        std::optional<BranchTakenVar>>>
+SleighDecoder::DecodeInstructionImpl(uint64_t address,
```
lkorenc, line 231: Try to think of type alias for that return type?

lib/Arch/Sleigh/Arch.cpp
```diff
@@ -519,9 +367,94 @@ std::optional<int32_t> SingleInstructionSleighContext::oneInstruction(
...
+template <typename... Ts>  // (7)
+struct Overload : Ts... {
+  using Ts::operator()...;
+};
+template <class... Ts>
+Overload(Ts...) -> Overload<Ts...>;
+
+}  // namespace
+
+void SleighDecoder::ApplyFlowToInstruction(remill::Instruction &inst) const {
```
lkorenc, line 381: Maybe make this of type `Category ConvertFlow( Flow, next_pc )` ? That was it is a pure function in a sense, which is usually easier to compose/reason about. Also it makes explicit what parts of `Instruction` it cares about.

lib/Arch/Sleigh/Arch.h
```diff
@@ -15,122 +15,31 @@
...
 
 #pragma once
-
 #include <remill/Arch/ArchBase.h>
+#include <remill/BC/SleighLifter.h>
 
 #include <sleigh/libsleigh.hh>
+#include <unordered_set>
+
+#include "ControlFlowStructuring.h"
```
lkorenc, line 24: Absolute path?

lib/Arch/Sleigh/ControlFlowStructuring.cpp
```diff
@@ -0,0 +1,470 @@
+#include "ControlFlowStructuring.h"
```
lkorenc, line 1: Absolute path?

lib/Arch/Sleigh/ControlFlowStructuring.cpp
```diff
@@ -0,0 +1,470 @@
...
+  }
+};
+
+template <class... Args>
+auto variant_cast(const std::variant<Args...> &v)
+    -> variant_cast_proxy<Args...> {
+  return {v};
+}
+
+enum CoarseEffect { ABNORMAL, NORMAL };
```
lkorenc, line 33: `class`?

lib/Arch/Sleigh/ControlFlowStructuring.cpp
```diff
@@ -0,0 +1,470 @@
...
+  if (isVarnodeInConstantSpace(op.vars[0])) {
+    // this is an internal branch.. we cant handle that right now
+    return std::nullopt;
+  }
+
+  if (op.vars[0].offset == next_pc) {
+    return {{CoarseEffect::NORMAL, is_conditional}};
+  } else {
+    return {{CoarseEffect::ABNORMAL, is_conditional}};
+  }
```
lkorenc, line 64: I would extract that check into some helper (`flow_normality( x, y ) -> CoarseEffect`?

lib/BC/SleighLifter.cpp
lkorenc: Well, add this as a comment to code then, so the next reader does not have to ponder as did I.

include/remill/Arch/Instruction.h
```diff
@@ -254,6 +257,149 @@ class Instruction {
...
+    bool operator==(const ConditionalInstruction &rhs) const;
+  };
+
+
+  using InstructionFlowCategory =
+      std::variant<NormalInsn, NoOp, InvalidInsn, ErrorInsn, DirectJump,
+                   IndirectJump, IndirectFunctionCall, DirectFunctionCall,
+                   FunctionReturn, AsyncHyperCall, ConditionalInstruction>;
+
+  InstructionFlowCategory flows;
```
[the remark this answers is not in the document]
2over12: A conditional is made out of two flows.

include/remill/BC/SleighLifter.h
```diff
@@ -29,38 +29,86 @@ class Sleigh;
...
  private:
   static void SetISelAttributes(llvm::Function *);
 
-  std::pair<LiftStatus, llvm::Function *>
-  LiftIntoInternalBlock(Instruction &inst, llvm::Module *target_mod,
-                        bool is_delayed);
+  std::pair<LiftStatus, std::optional<llvm::Function *>>
+  LiftIntoInternalBlockWithSleighState(
+      Instruction &inst, llvm::Module *target_mod, bool is_delayed,
+      const std::optional<sleigh::BranchTakenVar> &btaken);
```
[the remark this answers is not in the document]
2over12: this was a mistake it is only supposed to be attached to the lifterwithstate

lib/Arch/Sleigh/Thumb2Arch.cpp
2over12: Not really, are instructions at misaligned addresses a thing, afaik instructions always align them by clearing the last bit regardless of if it mode switches or not

lib/Arch/Context.cpp
2over12: You mean the default implementation of equality? they all are i think... it's just this is C++17

lib/Arch/Sleigh/Arch.cpp
2over12: So we also have to setup the branch_taken not_taken pc for the flows for which this is applicable so. im not sure this type makes sense

lib/Arch/Sleigh/ControlFlowStructuring.cpp
2over12: i dont get it? is this referring to CoarseFlow maybe?

lib/Arch/Sleigh/ControlFlowStructuring.cpp
```diff
@@ -0,0 +1,470 @@
...
+      res.emplace(ind, *cc);
+    }
+
+    ind++;
+  }
+
+  auto last_index = ops.size();
+  // insert a pseudo control flow op at the end
+  // add a fallthrough insn at +1 to represent a last fallthrough if there is a chance we fallthrough at the end
+  if (ops.empty() ||
```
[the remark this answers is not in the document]
2over12: need to check empty first, cant extract into variables since we cant access ops.size()-1 if !ops.empty(), could invert the condition then nest two ifs and return out, but that's kinda ugly

lib/Arch/Sleigh/ControlFlowStructuring.h
```diff
@@ -0,0 +1,61 @@
...
+
+#include <sleigh/libsleigh.hh>
+#include <unordered_map>
+
+namespace remill::sleigh {
+
+bool isVarnodeInConstantSpace(VarnodeData vnode);
+
+
+struct RemillPcodeOp {
```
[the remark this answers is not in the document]
2over12: PcodeOp is also in scope

lib/Arch/Sleigh/Thumb2Arch.cpp
sschriner: That was my understanding as well

include/remill/BC/SleighLifter.h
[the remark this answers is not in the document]
lkorenc: Can it be fixed then?

lib/Arch/Instruction.cpp
lkorenc: Bump.

lib/Arch/Instruction.cpp
lkorenc: Add TODO that says it should be upgraded once we have `c++20`.

lib/Arch/Sleigh/ControlFlowStructuring.cpp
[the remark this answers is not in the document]
lkorenc: I meant more like
```
auto good_name = whole condition;
if ( good_name ) {
  ...
}
```
or I would just move the `{` to the new line, so it is really easy to see where the condition ends.

lib/BC/SleighLifter.cpp
lkorenc: Bump.

lkorenc (approved):

include/remill/BC/SleighLifter.h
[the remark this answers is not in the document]
2over12: It is, the member no longer exists, the parameter is the single source of truth

PR #625, opened 2022-09-02
Emit IR for `syscall` in addition to a `__remill_async_hyper_call`

tetsuo-cpp:
At the moment, we simply add a call to `__remill_async_hyper_call` to indicate a control flow change. We should also emit IR for the actual instruction.

I used these for reference:
https://www.cs.uaf.edu/2017/fall/cs301/lecture/11_17_syscall.html
https://www.felixcloutier.com/x86/syscall.html
https://www.felixcloutier.com/x86/sysenter
https://www.felixcloutier.com/x86/sysexit

### DIFF
--- a/lib/Arch/Runtime/HyperCall.cpp
+++ b/lib/Arch/Runtime/HyperCall.cpp
@@ -44,6 +44,20 @@
 
 Memory *__remill_sync_hyper_call(State &state, Memory *mem,
                                  SyncHyperCall::Name call) {
+
+#if REMILL_HYPERCALL_X86 || REMILL_HYPERCALL_AMD64
+  register unsigned int *rsp asm("rsp") = &state.gpr.rsp.dword;
+  register unsigned int *rbp asm("rbp") = &state.gpr.rbp.dword;
+  register unsigned int *r8 asm("r8") = &state.gpr.r8.dword;
+  register unsigned int *r9 asm("r9") = &state.gpr.r9.dword;
+  register unsigned int *r10 asm("r10") = &state.gpr.r10.dword;
+  register unsigned int *r11 asm("r11") = &state.gpr.r11.dword;
+  register unsigned int *r12 asm("r12") = &state.gpr.r12.dword;
+  register unsigned int *r13 asm("r13") = &state.gpr.r13.dword;
+  register unsigned int *r14 asm("r14") = &state.gpr.r14.dword;
+  register unsigned int *r15 asm("r15") = &state.gpr.r15.dword;
+#endif
+
   switch (call) {
 
 #if REMILL_HYPERCALL_X86 || REMILL_HYPERCALL_AMD64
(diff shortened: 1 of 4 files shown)


lib/Arch/X86/Runtime/CMakeLists.txt
```diff
@@ -35,14 +35,20 @@ function(add_runtime_helper target_name address_bit_size enable_avx enable_avx51
     set(required_cpp_standard "c++11")
   endif()
 
+  if (address_bit_size EQUAL 32)
```
tetsuo-cpp, line 38: This was broken as it was using `x86_64` in the target triple even for 32-bit.

tetsuo-cpp:
Whoops, I didn't get the inline asm syntax right. Fixing now...

lib/Arch/Runtime/HyperCall.cpp
```diff
@@ -129,6 +129,14 @@ Memory *__remill_sync_hyper_call(State &state, Memory *mem,
       mem = __remill_x86_set_segment_gs(mem);
       break;
 
+    case SyncHyperCall::kX86SysEnter: asm volatile("sysenter" : :); break;
```
pgoodman, line 132: I think these syscall-related inline asm blocks should take in the dwords of the main 7 or 8 gprs by value, e.g. rax, rbx, ..., even rsp, and output rax on 32-bit, and on 64 bit, they should take in even more, maybe even all general-purpose regs.

lib/Arch/X86/Runtime/CMakeLists.txt
pgoodman: I think the prior reason was two-fold:
1) on x86-64, we know that trivial structs in C++, e.g. what we use for the register, e.g. Rn<...> and RnW<...>, would turn into basic integral or pointer arguments in the bitcode. On a 32-bit target, this wasn't guaranteed.

2) We couldn't guarantee i386 headers were present. You merging and extension of msp430 stuff into the headers might have resolved that.

I can see this making a material difference now that there is inline assembly with register constraints. What I'm less certain of are the end-to-end impact on the lifters. But if the tests still pass then maybe it's all fine?

pgoodman (changes requested):

lib/Arch/Runtime/HyperCall.cpp
tetsuo-cpp: Just for my own understanding, is that because the argument registers for `syscall` are OS dependent?

I chose the registers based on the table here, but I realise it says Linux specifically: https://www.cs.uaf.edu/2017/fall/cs301/lecture/11_17_syscall.html

lib/Arch/X86/Runtime/CMakeLists.txt
tetsuo-cpp: We only cross-compile the hyper calls (everything else in the runtime just uses the host machine's target triple). Then we `llvm-link` everything together. I believe if I use `llvm-dis` on the resulting runtime, the target triple is that of the host machine, which was how it worked before my hypercall implementation. So I don't think this should have an effect, except fixing how the hyper calls are compiled.

lib/Arch/Runtime/HyperCall.cpp
pgoodman: There is an ABI, that is for sure, but the semantics don't actually know the target OS / ABI, so I was erring on the side of "copy everything in." The 32- vs. 64-bit distinction is there because in 32-bit bitcode, you can't pass a 64-bit value in. I suppose you could use something like `"a"(state.gpr.aword)`, and the `aword` macro will expand to one of `dword` or `qword`.

tetsuo-cpp: 4f001ec

tetsuo-cpp:
@pgoodman Ok, I think this should do the trick.

lib/Arch/Runtime/HyperCall.cpp
```diff
@@ -44,6 +44,20 @@
 
 Memory *__remill_sync_hyper_call(State &state, Memory *mem,
                                  SyncHyperCall::Name call) {
+
+#if REMILL_HYPERCALL_X86 || REMILL_HYPERCALL_AMD64
```
pgoodman, line 48: I think for x86, you can't use `r`-prefixes, you need the `e`-prefixes, e.g. `ebp` vs. `rbp`. I think you'll need different variants of this per architecture. So on amd64, have the 64-bit variants, that touch `.qword`, and on x86, have the 32-bit variants that touch `.dword`.

I am confused as to why these are all pointers into the state structure. As far as I can tell, you are saying: this register must hold the address of this field in the state structure, as opposed to: this register must hold the value of the field in the state structure.

I am slightly sketched out about the implicit stack switch here with `rsp`. I don't think we can properly express that. Otoh, it doesn't really matter, because we're trying to express the dataflow, not something that is practically compilable, so it's fine.

In the register constraints, e.g. `"r"(rbp)`, you might also be able to do `"%rbp"(rbp)` to really hammer it home.

Finally, I think the stack pointer should probably be an output.

pgoodman (changes requested):

lib/Arch/Runtime/HyperCall.cpp
tetsuo-cpp: [115fd65](https://github.com/lifting-bits/remill/pull/625/commits/115fd655eb62aad6f18bb34a4a981409a3cebf0c)

tetsuo-cpp: > In the register constraints, e.g. "r"(rbp), you might also be able to do "%rbp"(rbp) to really hammer it home.

I can't seem to find where that type of constraint (`%rbp`) is documented so I didn't go ahead with that. Addressed everything else though.

pgoodman (approved):

PR #630, opened 2022-09-29
Adjust LLVM version to include minor version

maxammann:
Remill expects the minor version to be included when looking for semantics: https://github.com/lifting-bits/remill/blob/eef338df00e0ee82e7648a28c7979c661a79526f/lib/BC/Util.cpp#L538


CMakeLists.txt
```diff
@@ -64,7 +64,7 @@ list(GET LLVM_VERSION_LIST 1 LLVM_MINOR_VERSION)
 set(LLVM_MAJOR_VERSION "${LLVM_MAJOR_VERSION}")
 set(LLVM_MINOR_VERSION "${LLVM_MINOR_VERSION}")
 
-set(REMILL_LLVM_VERSION "${LLVM_MAJOR_VERSION}")
+set(REMILL_LLVM_VERSION "${LLVM_MAJOR_VERSION}.${LLVM_MINOR_VERSION}")
```
pgoodman, line 67: Don't forget remillConfig.cmake.in ;-)

pgoodman (changes requested):

CMakeLists.txt
maxammann: I changed the other file. Please let me know if that is correct (if the CI is not checking it). Sadly I don't have a quick local setup to test this.

pgoodman:
@maxammann Can you update the binary name of remill lift to use the LLVM major version number, instead of the remill llvm number? This should resolve test failures.

image

Another reasonable question is: should semantics search look at `MAJOR_MINOR` or just focus on major version. I recall a while back we swapped from major.minor to major, right around when LLVM changed its versioning approach (LLVM 4 iirc). If the semantics are installed into remill/<major>/whatever or something like that, then it seems like the actual fix should be to change Util.cpp, and leave the cmake as is (though exporting whatever ends up being the right thing in remillConfig.cmake is a good idea).

maxammann:
> Another reasonable question is: should semantics search look at MAJOR_MINOR or just focus on major version. 

I guess you are right and we should use only the major version. Which means that this should be fixed in remill. At least on ubuntu binaries are called `clang-14` instead of `clang-14.0`.

maxammann:
@pgoodman I would just close this and fix the issue in irene3 if that is fine for you. I think using only the major version is the correct way of handling this.

PR #632, opened 2022-10-05
Sleigh Preliminary Floating Point Support

2over12:


lib/BC/SleighLifter.cpp
```diff
@@ -368,11 +374,14 @@ class SleighLifter::PcodeToLLVMEmitIntoBlock : public PcodeEmit {
...
+    auto reg_ptr = this->unknown_regs.GetUniquePtr(target_vnode.offset,
+                                                   target_vnode.size, bldr);
     print_vardata(this->insn_lifter_parent.GetEngine(), ss, target_vnode);
-    DLOG(ERROR) << "Creating unique for unkown register: " << ss.str();
+    DLOG(ERROR) << "Creating unique for unkown register: " << ss.str() << " "
+                << reg_ptr->getName().str();
 
-    return RegisterValue::CreatRegister(this->unknown_regs.GetUniquePtr(
-        target_vnode.offset, target_vnode.size, bldr));
+    return RegisterValue::CreatRegister(reg_ptr);
```
sschriner, line 384:
```suggestion
    return RegisterValue::CreateRegister(reg_ptr);
```

sschriner (changes requested):

lib/BC/SleighLifter.cpp
```diff
@@ -484,10 +495,97 @@ class SleighLifter::PcodeToLLVMEmitIntoBlock : public PcodeEmit {
...
+                                                   outvar, input_var);
+      }
+      case OpCode::CPUI_FLOAT_ROUND: {
+        return this->LiftUnaryOpWithFloatIntrinsic(bldr, llvm::Intrinsic::round,
+                                                   outvar, input_var);
+      }
+      case OpCode::CPUI_FLOAT_NAN: {
+        auto nan_inval = this->LiftFloatInParam(bldr, input_var);
+        if (nan_inval.has_value()) {
+          // LLVM trunk has an `isnan` intrinsic but to support older versions, I think we need to do this.
```
pgoodman, line 532: @2over12 Is this intrinsic available in llvm 15?

include/remill/Arch/AArch32/Runtime/State.h
```diff
@@ -105,10 +126,45 @@ struct alignas(8) SR final {
   uint8_t _padding[2];
 } __attribute__((packed));
 
+
+// Ghidra mantains a unit32_t representing FPSCR that gets synced to NG ZR CY and OV, so we mantain this state too
```
lkorenc, line 130: Typo - mantain.

lib/BC/SleighLifter.cpp
```diff
@@ -484,10 +495,97 @@ class SleighLifter::PcodeToLLVMEmitIntoBlock : public PcodeEmit {
     return LiftStatus::kLiftedInstruction;
   }
 
+  LiftStatus LiftFloatUnop(llvm::IRBuilder<> &bldr, OpCode opc,
+                           VarnodeData *outvar, VarnodeData input_var) {
```
lkorenc, line 499: I would consider taking `outvar` by `&` in case it cannot be `nullptr` (if it can be done, I see if passed along to some other calls).

lib/BC/SleighLifter.cpp
```diff
@@ -839,10 +863,49 @@ class SleighLifter::PcodeToLLVMEmitIntoBlock : public PcodeEmit {
...
+      default: return std::nullopt;
+    }
+  }
+
+
+  llvm::Value *CastFloatResult(llvm::IRBuilder<> &bldr,
+                               VarnodeData output_varnode,
+                               llvm::Value *maybe_float) {
+
+    auto curr_value = maybe_float;
```
lkorenc, line 880: Unless I am missing something we can just use `maybe_float` directly?

lkorenc (approved):

lib/BC/SleighLifter.cpp
2over12: I think I'm going to keep it as is since it gets passed down to LiftStoreIntoOutParam which is a generic function for requiring an outparam to store into and handles the nullcheck. Alternatively, I could take a reference and make it back into a pointer but that seems eh

lib/BC/SleighLifter.cpp
2over12: I dont think it's in llvm 14 unless I'm missing something

pgoodman: What I mean is that if it's in 15, and if remill i s now on 15, then we can assume it's available and drop support for 14.

2over12: Remill is on 14

pgoodman: Okay, file an issue for updating this when the intrinsic support becomes available.

lib/BC/SleighLifter.cpp
```diff
@@ -839,10 +863,48 @@ class SleighLifter::PcodeToLLVMEmitIntoBlock : public PcodeEmit {
     }
   }
 
+
+  std::optional<llvm::Type *> GetFloatTypeOfByteSize(size_t byte_size) {
+    switch (byte_size) {
+      case 4: return llvm::Type::getFloatTy(this->context);
```
pgoodman, line 869: Add support for HalfTy? or Float128Ty? Might not be needed now, but perhaps doesn't work?

pgoodman (approved):
TODOs:

- [ ] File issue for adding the isnan intrinsic support when it becomes available.
- [ ] Decide if you want to add HalfTy and other LLVM float types to the p-code lifter now or later.

Otherwise, this looks good to merge.

pgoodman (approved):

pgoodman (approved):

PR #638, opened 2022-11-23
Fix build with system LLVM

mrexodia:
When using a vanilla LLVM there were some issues with the build (when using system libraries):

1. Certain libraries were missing, this is fixed by simply linking to all available LLVM libs
2. Without `GFLAGS_USE_TARGET_NAMESPACE` the target `gflags::gflags` does not exist, not sure what vcpkg is doing different there, perhaps an older version?
3. The LLVM include directories are not mentioned anywhere so remill fails to compile

These issues were found when trying a [superbuild](https://github.com/lifting-bits/cxx-common/issues/933#issuecomment-1323678496) without vcpkg.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -41,18 +41,13 @@ find_package(Z3 CONFIG REQUIRED)
 
 # LLVM
 find_package(LLVM CONFIG REQUIRED)
-llvm_map_components_to_libnames(llvm_libs
-  support core irreader
-  bitreader bitwriter
-  passes asmprinter
-  aarch64codegen aarch64asmparser
-  armcodegen armasmparser
-  interpreter mcjit
-  nvptxdesc
-  x86codegen x86asmparser
-  sparccodegen sparcasmparser
-  webassemblydesc)
-message(STATUS "LLVM Libraries: ${llvm_libs}")
+
+# https://github.com/JonathanSalwan/Triton/issues/1082#issuecomment-1030826696
+if(LLVM_LINK_LLVM_DYLIB)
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -41,18 +41,13 @@ find_package(Z3 CONFIG REQUIRED)
...
-  x86codegen x86asmparser
-  sparccodegen sparcasmparser
-  webassemblydesc)
-message(STATUS "LLVM Libraries: ${llvm_libs}")
+
+# https://github.com/JonathanSalwan/Triton/issues/1082#issuecomment-1030826696
+if(LLVM_LINK_LLVM_DYLIB)
+  set(llvm_libs LLVM)
+else()
+  set(llvm_libs ${LLVM_AVAILABLE_LIBS})
```
pgoodman, line 49: Still want `llvm_map_components_to_libnames` here, otherwise it starts to pull in things like `LibEdit`. In the past we've also faced linking issues with `LLVM_AVAILABLE_LIBS` having some things out-of-order.

pgoodman (changes requested):

CMakeLists.txt
mrexodia: Yeah that's true, I had to disable `LibEdit` on the LLVM command line. I'll try to reproduce the linker error then and see if I can find the component. Generally it shouldn't be an issue to link to all LLVM libraries, unused things are ignored. I've been using this since LLVM 10 without issues related to out-of-order linking.

mrexodia:
Will open a new PR later.

PR #643, opened 2022-12-08
PPC Sleigh patches

Ninja3047:
* Added sleigh patches required for ppc support
* Add register definition for ppc
* Modified sleigh patch script to generate patches for multiple .sinc files
* Updated README examples


tetsuo-cpp:
Thanks @Ninja3047. `remill-lift` isn't in a working state yet on the PPC branch so it'll be hard to validate this. I'll let you know when I've got things working and you should be able to easily validate this on some PC relative instructions.

tetsuo-cpp:
@Ninja3047

I don't claim that this is working 100% but I've got `remill-lift` to the point where it's spitting out something sensible for most of the instructions that I'm pasting out of Ghidra. It's probably at the point where you could start testing these patches. I've merged in my latest changes so this should be good to go.

The way I've been testing this is by invoking `remill-lift`. I find an instruction with some interesting looking P-Code in Ghidra and then copy the hex representation. Then I do:
```console
$ remill-lift-15 --arch ppc --ir_out /dev/stdout --bytes 2a07 --alsologtostderr
```

tetsuo-cpp (approved):
LGTM! Nice work.

tetsuo-cpp:
@Ninja3047 Oops, I shouldn't have deleted this branch. Just restored it then. Not sure if you're still using this branch for your unit tests.

PR #644, opened 2022-12-20
PPC tests

Ninja3047:
Work in progress tests for the work in progress PPC architecture support
Tests are expected to fail at the moment

PR #645, opened 2022-12-20
PowerPC Support

tetsuo-cpp:


Ninja3047:
Found an issue that i fixed here https://github.com/lifting-bits/remill/pull/644/commits/4da65bb89ac4cf50011dc2c3430d3a6b8c471577
although it looks like it doesn't handle the case where the shift amount is negative

Also fixed another issue with `INT2FLOAT`, it's supposed to output the appropriate sized float based on the output size
https://github.com/lifting-bits/remill/pull/644/commits/5424abf19ca302a419ef8c20a21921a9dc501533

Ninja3047:
Unique ptrs in the entry block appears to work though 🎉

tetsuo-cpp:
> Found an issue that i fixed here [4da65bb](https://github.com/lifting-bits/remill/commit/4da65bb89ac4cf50011dc2c3430d3a6b8c471577)

Oops thanks for the fix.

> although it looks like it doesn't handle the case where the shift amount is negative

Have you found negative shift amounts being lifted? The PCode reference indicates that it should be unsigned.

Ninja3047:
Haven't seen any examples of this occurring but i was looking at how they implement the semantic behavior in Ghidra itself and they seem to also check if it's negative. I don't think it matters either way since we probably won't run into it in practice. https://github.com/NationalSecurityAgency/ghidra/blob/master/Ghidra/Framework/SoftwareModeling/src/main/java/ghidra/pcode/opbehavior/OpBehaviorIntLeft.java#L33

tetsuo-cpp:
@2over12 I'm passing the VLE value into Sleigh with [639bd99](https://github.com/lifting-bits/remill/pull/645/commits/639bd99952f17a22fcf5de9f66df8e8f40d1ca6a). This should be cleaner than what I was originally doing.

tetsuo-cpp:
@2over12 I'm thinking that we should rename the target branch to `patchable-ir-main`, similar to what we're doing with Anvill.

lib/Arch/Sleigh/PPCArch.cpp
```diff
@@ -0,0 +1,272 @@
...
+class SleighPPCArch : public ArchBase {
+ public:
+  SleighPPCArch(llvm::LLVMContext *context_, OSName os_name_,
+                ArchName arch_name_)
+      : ArchBase(context_, os_name_, arch_name_),
+        decoder(*this) {}
+  virtual ~SleighPPCArch() = default;
+
+  DecodingContext CreateInitialContext(void) const override {
+    return DecodingContext().PutContextReg(kPPCVLERegName, 0);
```
Ninja3047, line 64: is there a way to specify which one to use?
something isn't quite right here since the tests still pass with this (they currently only test with VLE) so we're not setting this consistently everywhere it needs to

Ninja3047 (commented):
added some initial feedback

lib/Arch/Sleigh/PPCArch.cpp
Ninja3047: oops, nevermind
had a stale branch checked out

this does make the tests fail (correctly) but now there's not external way (as far as i can tell) to set this to VLE mode

tetsuo-cpp: Ah good point. I'll look into this.

I did it like this because Ghidra specifies a register value for `vle` when VLE mode is on but assumes it's off.

Ninja3047: this might be something that has to be set in the test harness actually
this is the way it's being set in the harness
```cpp
std::optional<std::pair<llvm::Function *, remill::Instruction>>
LiftingTester::LiftInstructionFunction(std::string_view fname,
                                       std::string_view bytes,
                                       uint64_t address) {
  remill::Instruction insn;
  // This works for now since each arch has an initial context that represents the arch correctly.
  if (!this->arch->DecodeInstruction(address, bytes, insn,
                                     this->arch->CreateInitialContext())) {
    LOG(ERROR) << "Failed decode";
    return std::nullopt;
  }

```

Ninja3047: pushed a fix for the tests, i think `CreateInitialContext` defaulting to 0 is fine

tests/PPC/TestLifting.cpp
```diff
@@ -241,6 +242,10 @@ int main(int argc, char **argv) {
   return RUN_ALL_TESTS();
 }
 
+
+inline const remill::DecodingContext kVLEContext =
```
Ninja3047, line 246: this we probably want to move to a header file like we have with Thumb
also provide another constant for non VLE context

include/remill/BC/PCodeCFG.h
```diff
@@ -0,0 +1,91 @@
...
+#include <sleigh/libsleigh.hh>
+#include <unordered_map>
+#include <variant>
+#include <vector>
+
+#include "lib/Arch/Sleigh/ControlFlowStructuring.h"
+
+
+namespace remill {
+namespace sleigh {
```
lkorenc, line 32: `namespace remill::sleigh`

include/remill/BC/PCodeCFG.h
```diff
@@ -0,0 +1,91 @@
...
+
+
+struct IntrainstructionIndex {
+  size_t target_block_ind;
```
lkorenc, line 36: `IntrainstructionIndex` is kinda hard to parse at quick glance. Can we come up with something easier? Does it represent instruction index (as in 5th instruction) in the block/function? Or is it more like an address offset? Add doc.

Also, with name as long as `target_block_ind` there is no reason not to include the `ex`. If cutting on letters, I would do something like `trg_block_idx` instead.

lib/BC/PcodeCFG.cpp
```diff
@@ -0,0 +1,166 @@
...
+  return PcodeCFGBuilder(linear_ops).Build();
+}
+
+
+std::vector<size_t> PcodeCFGBuilder::GetBlockStarts() {
+  std::vector<size_t> res;
+  res.push_back(0);
+
+
+  for (size_t curr_index = 0; curr_index < this->linear_ops.size();
```
lkorenc, line 28: No need to use `this->`.

lib/BC/PcodeCFG.cpp
```diff
@@ -0,0 +1,166 @@
...
+  }
+
+  return PcodeBlock(start_ind, std::move(ops),
+                    this->GetBlockExitsForIndex(next_start - 1));
+}
+
+
+namespace {
+struct IntraProcTransferCollector {
+  std::vector<size_t> operator()(const InstrExit &ex) {
```
lkorenc, line 50: Cannot this class simply hold `std::vector< std::size_t >` as an attribute? Then there is no need to always return it (which needs to be later merged) but instead push can happen directly?

As an entry point a static method can be exposed that creates this class, runs the visit and then returns the attribute.

lib/BC/PcodeCFG.cpp
```diff
@@ -0,0 +1,166 @@
...
+    std::vector<size_t> res = std::visit(*this, ex.true_branch);
+    std::vector<size_t> other = std::visit(*this, ex.false_branch);
+    res.insert(res.end(), other.begin(), other.end());
+    return res;
+  }
+};
+}  // namespace
+
+std::vector<size_t> PcodeCFGBuilder::GetIntraProcTargets(size_t index) {
+  std::variant<Exit, ConditionalExit> ex = this->GetBlockExitsForIndex(index);
```
lkorenc, line 73:
```suggestion
  auto ex = GetBlockExitsForIndex(index);
```

lib/BC/PcodeCFG.cpp
```diff
@@ -0,0 +1,166 @@
...
+  return std::visit(IntraProcTransferCollector{}, ex);
+}
+
+std::variant<Exit, ConditionalExit>
```
lkorenc, line 77: Type alias the variant type.

lib/BC/PcodeCFG.cpp
```diff
@@ -0,0 +1,166 @@
...
+PcodeCFGBuilder::GetBlockExitsForIndex(size_t index) {
+  CHECK(index < this->linear_ops.size());
+  const auto &curr_op = this->linear_ops[index];
```
lkorenc, line 80: Extract the `CHECK` + getter as standalone method.

lib/BC/PcodeCFG.cpp
```diff
@@ -0,0 +1,166 @@
...
+
+  auto build_direct_target_exit = [](VarnodeData target,
+                                     size_t curr_ind) -> Exit {
+    if (isVarnodeInConstantSpace(target)) {
+      // need to treat as signed?
+      return IntrainstructionIndex{curr_ind + target.offset};
+    } else {
```
lkorenc, line 87: You can drop `else` as then branch returns. But since nothing else happens in either branch this is more of personal preference.

lib/BC/PcodeCFG.cpp
```diff
@@ -0,0 +1,166 @@
...
+  std::variant<Exit, ConditionalExit> ex = this->GetBlockExitsForIndex(index);
+  return std::visit(IntraProcTransferCollector{}, ex);
+}
+
+std::variant<Exit, ConditionalExit>
+PcodeCFGBuilder::GetBlockExitsForIndex(size_t index) {
+  CHECK(index < this->linear_ops.size());
+  const auto &curr_op = this->linear_ops[index];
+
+  auto build_direct_target_exit = [](VarnodeData target,
```
lkorenc, line 82: `target` gets copied (which is fine if it is small), if not either take as const ref.

lib/BC/PcodeCFG.cpp
```diff
@@ -0,0 +1,166 @@
...
+    case CPUI_CBRANCH: {
+
+
+      auto taken_exit = build_direct_target_exit(curr_op.vars[0], index);
+
+      Exit fallthrough_exit = InstrExit{};
+      // if we are not the last pcodeop then we have an intraproc fallthrough
+      if (index < this->linear_ops.size() - 1) {
+        fallthrough_exit = IntrainstructionIndex{index + 1};
+      }
```
lkorenc, line 105: I strongly prefer
```
auto fallthrough_exit = [ & ] {
  if ( ... )
    return Intrains ...
  return {};
}();
```
to avoid that conditional mutation in a branch. Here it is not that terrible, but once a few lines (maybe another branch) gets added it can become a nightmare really quickly.

lib/BC/PcodeCFG.cpp
```diff
@@ -0,0 +1,166 @@
...
+
+PcodeBlock::PcodeBlock(size_t base_index)
+    : base_index(base_index),
+      ops(),
+      block_exit(Exit{InstrExit{}}) {}
+
+PcodeCFG PcodeCFGBuilder::Build() {
+
+  auto starts = this->GetBlockStarts();
+  std::set s(starts.begin(), starts.end());
```
lkorenc, line 129: Unless I am missing something crucial:
```suggestion
  auto s = starts;
```

lib/BC/PcodeCFG.cpp
```diff
@@ -0,0 +1,166 @@
...
+  starts.assign(s.begin(), s.end());
```
lkorenc, line 130: Overall I am really confused what's happening here (and mainly why). Elaborate please.

lib/BC/PcodeCFG.cpp
```diff
@@ -0,0 +1,166 @@
...
+
+  std::unordered_map<size_t, PcodeBlock> blocks;
+
+  if (this->linear_ops.empty()) {
+    // There is no insturction at 0 to build a block at
+    // build an empty block so we transfer through to exit by terminating the block
+    blocks.insert({0, PcodeBlock(0)});
```
lkorenc, line 137:
```suggestion
    blocks.emplace(0, PcodeBlock(0));
```

lib/BC/PcodeCFG.cpp
```diff
@@ -0,0 +1,166 @@
...
+    return blocks;
+  }
+
+  for (size_t i = 0; i < starts.size(); i++) {
+    auto next_start = this->linear_ops.size();
+    if ((i + 1) < starts.size()) {
+      next_start = starts[i + 1];
+    }
```
lkorenc, line 145: I would format this as ternary operator, to simplify the flow.

include/remill/BC/PCodeCFG.h
```diff
@@ -0,0 +1,91 @@
...
+  std::variant<Exit, ConditionalExit> GetBlockExitsForIndex(size_t index);
+
+  std::vector<size_t> GetIntraProcTargets(size_t index);
+
+  std::vector<size_t> GetBlockStarts();
+
+ public:
+  static PcodeCFG CreateCFG(const std::vector<RemillPcodeOp> &linear_ops);
+
+  PcodeCFG Build();
```
lkorenc, line 86: This is really confusing. What does `Build` do and why is `CreateCFG` not a ctor? From the type of `Build` I would expect it creates a new instance from current one, but it is used in `CreateCFG` instead.

I would possibly rename `Build` and add documentation.

include/remill/BC/SleighLifter.h
```diff
@@ -74,7 +75,8 @@ class SleighLifter : public InstructionLifter {
   std::pair<LiftStatus, std::optional<llvm::Function *>>
   LiftIntoInternalBlockWithSleighState(
       Instruction &inst, llvm::Module *target_mod, bool is_delayed,
-      const std::optional<sleigh::BranchTakenVar> &btaken);
+      const std::optional<sleigh::BranchTakenVar> &btaken,
```
lkorenc, line 78: I would type alias `std::optional< sleigh::BranchTakenVar >` as it is used often and is pretty long.

lib/Arch/Sleigh/Arch.cpp
```diff
@@ -201,8 +202,8 @@ bool SleighDecoder::DecodeInstruction(uint64_t address,
         << "Should always emit branch taken var for conditional instruction";
   }
 
-  inst.SetLifter(std::make_shared<SleighLifterWithState>(res_cat->second,
-                                                         this->GetLifter()));
+  inst.SetLifter(std::make_shared<SleighLifterWithState>(
+      res_cat->second, std::move(context_values), this->GetLifter()));
```
lkorenc, line 206: Call the `context.GetContextValues()` here? Or is there a reason copy must happen earlier? If it is so, add comment.

lib/Arch/Sleigh/ControlFlowStructuring.cpp
```diff
@@ -127,6 +135,16 @@ CoarseCategoryFromFlows(const std::map<size_t, CoarseFlow> &ops) {
     return CoarseCategory::kCatNormal;
   }
 
+  auto all_normal_or_intra_effects = std::all_of(
+      ops.begin(), ops.end(), [](const std::pair<size_t, CoarseFlow> &op) {
```
lkorenc, line 139: Extract lambda as local var to improve readability.

lib/Arch/Sleigh/ControlFlowStructuring.cpp
```diff
@@ -410,31 +428,67 @@ ControlFlowStructureAnalysis::ComputeCategory(
...
+          return !op.outvar.has_value() ||
+                 !context_updater.GetRemillReg(*op.outvar).has_value();
+        });
+
+    if (no_context_updates) {
+      Instruction::NormalInsn norm_insn(
+          (Instruction::FallthroughFlow(entry_context)));
+      Instruction::InstructionFlowCategory ifc = norm_insn;
+      return std::make_pair(ifc, std::nullopt);
+    } else {
```
lkorenc, line 447: Drop `else`.

lib/Arch/Sleigh/ControlFlowStructuring.cpp
```diff
@@ -410,31 +428,67 @@ ControlFlowStructureAnalysis::ComputeCategory(
...
 }
 
+std::optional<std::string>
+ContextUpdater::GetRemillReg(const VarnodeData &outvar) {
+  auto reg_name =
+      this->engine.getRegisterName(outvar.space, outvar.offset, outvar.size);
+  auto maybe_remill_reg_name = this->context_reg_mapping.find(reg_name);
+  if (maybe_remill_reg_name != this->context_reg_mapping.end()) {
+    return maybe_remill_reg_name->second;
+  } else {
```
lkorenc, line 473: Drop `else` and all those `this->`.

lkorenc: I would also change `maybe_remill_reg_name` to simply `it` (since technically it is an iterator and function is short enough to not really suffer from loss of what is being retrieved).

lib/Arch/Sleigh/PPC.h
```diff
@@ -0,0 +1,41 @@
...
+class SleighPPCDecoder final : public remill::sleigh::SleighDecoder {
+ public:
+  SleighPPCDecoder(const remill::Arch &);
+
+  llvm::Value *LiftPcFromCurrPc(llvm::IRBuilder<> &, llvm::Value *,
+                                size_t) const override;
+
+  void InitializeSleighContext(
+      remill::sleigh::SingleInstructionSleighContext &,
+      const std::map<std::string, uint64_t> &) const override;
```
lkorenc, line 38: Add arguments, since there are no type aliases it can be kinda hard to know what argument is what.

lib/Arch/Sleigh/PPCArch.cpp
```diff
@@ -0,0 +1,272 @@
...
+#include "PPC.h"
+
+#define INCLUDED_FROM_REMILL
+#include <remill/Arch/PPC/Runtime/State.h>
+
+namespace remill {
+
+namespace sleighppc {
+
+static const char *kPPCVLERegName = "VLEReg";
```
lkorenc, line 26: Can be `constexpr`.

lib/Arch/Sleigh/PPCArch.cpp
```diff
@@ -0,0 +1,272 @@
...
+
+void SleighPPCDecoder::InitializeSleighContext(
+    remill::sleigh::SingleInstructionSleighContext &ctxt,
+    const std::map<std::string, uint64_t> &context_values) const {
+  // If the context value mappings specify a value for the VLE register, let's pass that into
+  // Sleigh.
+  //
+  // Otherwise, default to VLE off.
+  uint64_t vle_value = 0;
+  const auto vle_iter = context_values.find(kPPCVLERegName);
```
lkorenc, line 48: Extract this as standalone helper function since is sounds generically useful (both get/set with defaults).

lib/BC/PcodeCFG.cpp
2over12: I think it could probably could be a `void` operator yeah

lib/BC/PcodeCFG.cpp
2over12: The point of these two lines is to sort and de-duplicate the vector, ie. shove everything into a set and extract

bin/lift/Lift.cpp
```diff
@@ -432,6 +437,8 @@ int main(int argc, char *argv[]) {
 
     guide.slp_vectorize = true;
     guide.loop_vectorize = true;
+
+    CHECK(!llvm::verifyModule(dest_module, &llvm::errs()));
```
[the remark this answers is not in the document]
lkorenc: I am not sure from the top of my head but it may be possible there is some version of that function that did `CHECK` inside itself (but maybe that was fixed some time ago) - definitely better to use existing utility though.

lib/Arch/Sleigh/PPCArch.cpp
```diff
@@ -0,0 +1,272 @@
...
+
+    auto &entry_block = bb_func->getEntryBlock();
+    llvm::IRBuilder<> ir(&entry_block);
+
+    const auto pc_arg = NthArgument(bb_func, kPCArgNum);
+    const auto state_ptr_arg = NthArgument(bb_func, kStatePointerArgNum);
+    ir.CreateStore(pc_arg,
+                   ir.CreateAlloca(addr, nullptr, kNextPCVariableName.data()));
+    ir.CreateStore(pc_arg, ir.CreateAlloca(addr, nullptr,
+                                           kIgnoreNextPCVariableName.data()));
```
lkorenc, line 254: Can be polished as
```
auto mk_alloca = [ & ]( auto &from ) {
  return ir.CreateAlloca( addr, nullptr, from.data() );
};
ir.CreateStore( pc_arg, mk_alloca( kNextPCVariableName ) );
...
```

lib/BC/PcodeCFG.cpp
```diff
@@ -0,0 +1,166 @@
...
+
+PcodeCFG
+PcodeCFGBuilder::CreateCFG(const std::vector<RemillPcodeOp> &linear_ops) {
+  return PcodeCFGBuilder(linear_ops).Build();
+}
+
+
+std::vector<size_t> PcodeCFGBuilder::GetBlockStarts() {
+  std::vector<size_t> res;
+  res.push_back(0);
```
lkorenc, line 25: `std::vector< size_t > res = { 0 };`

lib/BC/PcodeCFG.cpp
```diff
@@ -0,0 +1,166 @@
...
+
+
+namespace {
+struct IntraProcTransferCollector {
+  std::vector<size_t> operator()(const InstrExit &ex) {
+    return std::vector<size_t>();
+  }
+
+  std::vector<size_t> operator()(const IntrainstructionIndex &ex) {
+    return std::vector<size_t>(1, ex.target_block_ind);
```
lkorenc, line 55:
```suggestion
    return { ex.target_block_ind };
```

lib/BC/PcodeCFG.cpp
```diff
@@ -0,0 +1,166 @@
...
+  return res;
+}
+
+PcodeBlock PcodeCFGBuilder::BuildBlock(size_t start_ind, size_t next_start) {
+  std::vector<RemillPcodeOp> ops;
+
+  for (size_t i = start_ind; i < next_start; i++) {
+    ops.push_back(this->linear_ops[i]);
+  }
+
```
lkorenc, line 42:
```
using pcode_ops_t = ...;

auto begin = std::next( linear_ops.begin() + start );
auto end = std::next( linear_ops.begin() + next_start );
return ...( start, pcode_ops_t( begin, end ), ... );
```

lib/BC/PcodeCFG.cpp
```diff
@@ -0,0 +1,166 @@
...
+  };
+  switch (curr_op.op) {
+    case CPUI_BRANCH:
+    case CPUI_CALL: {
+      return Exit{build_direct_target_exit(curr_op.vars[0], index)};
+    }
+    case CPUI_CBRANCH: {
+
+
+      auto taken_exit = build_direct_target_exit(curr_op.vars[0], index);
```
lkorenc, line 99: I would move this down before the return - while I need both `taken` and `fallthrough` in it (so it cannot be said just define stuff at the closest place it is used). Since `fallthrough` needs extra computation (if branch) I would introduce `taken` later to simply say it plays no role in `fallthrough` computation to be kind to reader.

lib/BC/PcodeCFG.cpp
lkorenc: Yeah, add a comment, ideally also with reasoning why the order is not damaged (or why we don't care).

lib/BC/SleighLifter.cpp
```diff
@@ -297,12 +304,17 @@ class SleighLifter::PcodeToLLVMEmitIntoBlock : public PcodeEmit {
...
 
+  llvm::BasicBlock *entry_block;
   llvm::BasicBlock *exit_block;
 
   size_t curr_id;
 
   const std::optional<remill::sleigh::BranchTakenVar> &to_lift_btaken;
 
+  std::unordered_map<size_t, llvm::BasicBlock *>
+      block_start_index_to_basic_block;
```
lkorenc, line 315: Shorter name?

lib/BC/SleighLifter.cpp
```diff
@@ -297,12 +304,17 @@ class SleighLifter::PcodeToLLVMEmitIntoBlock : public PcodeEmit {
   // Generic sleigh arch
   std::vector<std::string> user_op_names;
 
+  llvm::BasicBlock *entry_block;
```
lkorenc, line 307: I would defensively both initialise to `nullptr`.

lib/BC/SleighLifter.cpp
```diff
@@ -313,6 +325,19 @@ class SleighLifter::PcodeToLLVMEmitIntoBlock : public PcodeEmit {
   }
 
  public:
+  llvm::BasicBlock *GetOrCreateBlock(size_t target) {
+    auto blk = block_start_index_to_basic_block.find(target);
+    if (blk != block_start_index_to_basic_block.end()) {
+      return blk->second;
+    }
```
lkorenc, line 332: ^ is a separate function. Then here
```
if ( auto blk = fetch( target ) )
  return blk;
... create it ...
```

lib/BC/SleighLifter.cpp
```diff
@@ -620,15 +656,15 @@ class SleighLifter::PcodeToLLVMEmitIntoBlock : public PcodeEmit {
         // directs dont read the address of the variable, the offset is the jump
         // TODO(Ian): handle other address spaces
         if (isVarnodeInConstantSpace(input_var)) {
-          DLOG(ERROR) << "Internal control flow not supported";
-          return LiftStatus::kLiftedUnsupportedInstruction;
+          this->TerminateBlock();
+          return LiftStatus::kLiftedInstruction;
+        } else {
```
lkorenc, line 661: Drop `else`.

lib/BC/SleighLifter.cpp
```diff
@@ -700,24 +736,65 @@ class SleighLifter::PcodeToLLVMEmitIntoBlock : public PcodeEmit {
   static std::unordered_set<OpCode> INTEGER_COMP_OPS;
 
 
+  struct VisitExit {
+    PcodeToLLVMEmitIntoBlock &th;
```
lkorenc, line 740: I am ok with short names in use-cases like this, but as a random reader I cannot deduce what `th` is, so maybe chose something else?

lib/BC/SleighLifter.cpp
```diff
@@ -700,24 +736,65 @@ class SleighLifter::PcodeToLLVMEmitIntoBlock : public PcodeEmit {
...
+
+  struct VisitBlockExitFalse {
+    PcodeToLLVMEmitIntoBlock &th;
+    llvm::BasicBlock *operator()(const sleigh::Exit &exit) {
+      return std::visit(VisitExit{th}, exit);
+    }
+    llvm::BasicBlock *operator()(const sleigh::ConditionalExit &exit) {
+      return std::visit(VisitExit{th}, exit.false_branch);
+    }
+  };
```
lkorenc, line 767: Copy & pasted 2 times (almost 3). Extract all shared parts ( attribute + `Exit` visit) into base class from which then inherit. Figure out, whether the other two should also inherit from the first one to inherit those visits.

lib/BC/SleighLifter.cpp
```diff
@@ -726,47 +803,45 @@ class SleighLifter::PcodeToLLVMEmitIntoBlock : public PcodeEmit {
...
+          bldr, lhs, this->insn_lifter_parent.GetWordType());
 
 
-    auto pc_reg_param = this->LiftNormalRegister(bldr, "PC");
-    assert(pc_reg_param.has_value());
-    auto pc_reg_ptr = *pc_reg_param;
-    auto orig_pc_value =
-        pc_reg_ptr->LiftAsInParam(bldr, this->insn_lifter_parent.GetWordType());
+      auto pc_reg_param = this->LiftNormalRegister(bldr, "PC");
+      assert(pc_reg_param.has_value());
```
lkorenc, line 818:
```suggestion
      CHECK(pc_reg_param.has_value());
```

lib/BC/SleighLifter.cpp
```diff
@@ -1279,14 +1364,26 @@ std::map<OpCode, SleighLifter::PcodeToLLVMEmitIntoBlock::BinaryOperator>
...
              rhs = bldr.CreateZExtOrTrunc(rhs, lhs->getType());
            }
-           return bldr.CreateLShr(lhs, rhs);
+           // If the number of bits we're shifting exceeds the bit width of the
+           // other operand, the result should be zero.
+           auto max_shift = llvm::ConstantInt::get(
+               lhs->getType(), lhs->getType()->getIntegerBitWidth());
+           return bldr.CreateSelect(bldr.CreateICmpSGE(rhs, max_shift),
+                                    llvm::ConstantInt::get(lhs->getType(), 0),
+                                    bldr.CreateLShr(lhs, rhs));
```
lkorenc, line 1386: Copy & paste of the change above. Extract.

lib/BC/SleighLifter.cpp
```diff
@@ -1461,25 +1556,31 @@ SleighLifter::LiftIntoInternalBlockWithSleighState(
...
 
+  ir.CreateBr(lifter.GetOrCreateBlock(0));
 
-  lifter.TerminateBlock();
 
   // Setup like an ISEL
   SleighLifter::SetISelAttributes(target_func);
   remill::InitFunctionAttributes(target_func);
 
+  CHECK(!llvm::verifyFunction(*target_func, &llvm::errs()));
```
lkorenc, line 1576: Use utility function.

test_runner_lib/include/test_runner/TestOutputSpec.h
```diff
@@ -0,0 +1,167 @@
...
+  RegisterConditionList register_preconditions;
+  RegisterConditionList register_postconditions;
+  MemoryConditionList initial_memory_conditions;
+  MemoryConditionList expected_memory_conditions;
+  RegisterAccessorMap reg_to_accessor;
+
+  template <typename T>
+  void ApplyCondition(S &state, const std::string &reg, T value) const {
+    auto accessor = reg_to_accessor.find(reg);
+    if (accessor != reg_to_accessor.end()) {
```
lkorenc, line 66: Early return.

test_runner_lib/include/test_runner/TestOutputSpec.h
```diff
@@ -0,0 +1,167 @@
...
+      auto wrapper = accessor->second(state);
+      if (auto underlying = std::get_if<std::reference_wrapper<T>>(&wrapper)) {
```
lkorenc, line 68: Wrap `std::get_if< std::reference_wrapper< T > >( arg )` into a helper fn, it is used below as well.

lkorenc: Actually, since it always throws on fail, the throw may be hidden in the helper and it can return `underlying->get()`.

test_runner_lib/include/test_runner/TestOutputSpec.h
```diff
@@ -0,0 +1,167 @@
...
+        throw std::runtime_error(
+            std::string("Reg value " + reg + " has incorrect type"));
+      }
+    } else {
+      throw std::runtime_error(std::string("Unknown reg: ") + reg);
+    }
+  }
+
+  template <typename T>
+  void CheckCondition(S &state, const std::string &reg, T value) const {
```
lkorenc, line 80: And since we need the whole mechanism here again:
```
T  &get_reg( ... )
{
  if ( !accesor_present )
    throw;
  return get_if_helper();
}
 
void CheckCond( ... ) { CHECK_EQ( actual, helper( ... ) ) }
void Apply( ... ) { get_reg( ... ) = value; } 
```

lkorenc (changes requested):

tetsuo-cpp:
@Ninja3047 What happened with [e4a4b74](https://github.com/lifting-bits/remill/pull/645/commits/e4a4b7429edb7250c892eb055d8c67499631329f)? Can we no longer support LLVM 14?

include/remill/Arch/PPC/Runtime/State.h
```diff
@@ -0,0 +1,375 @@
...
+
+static_assert(512 == sizeof(FPR), "Invalid structure packing of `FPR`.");
+
+// These are actually bitflags.
+//
+// Remill's notion of registers operates at a byte granularity so these need to
+// take a full byte.
+struct alignas(8) CRFlags final {
+
+  uint8_t cr0;
```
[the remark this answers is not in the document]
tetsuo-cpp: [97b32f8](https://github.com/lifting-bits/remill/pull/645/commits/97b32f8a1e9d5ab32a06956cec6fa3aa29bcadc1)

include/remill/Arch/Name.h
```diff
@@ -64,6 +69,15 @@
...
 #    endif
+#  elif defined(__PPC__)
+#    define REMILL_ARCH "ppc"
+#    define REMILL_ON_AMD64 0
+#    define REMILL_ON_X86 0
+#    define REMILL_ON_AARCH64 0
+#    define REMILL_ON_AARCH32 0
+#    define REMILL_ON_SPARC64 0
+#    define REMILL_ON_SPARC32 0
+#    define REMILL_ON_PPC 1
```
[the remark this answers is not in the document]
tetsuo-cpp: I don't think we're using this. We do cross-compile the hyper call implementations to give us access to inline asm but I don't believe they're using these.

lib/Arch/Sleigh/ControlFlowStructuring.cpp
```diff
@@ -29,14 +32,19 @@ auto variant_cast(const std::variant<Args...> &v)
...
-enum CoarseEffect { kAbnormal, kNormal };
+enum CoarseEffect { kAbnormal, kNormal, kIntraInstruction };
 
 struct CoarseFlow {
   CoarseEffect eff;
   bool is_conditional;
 };
 
-enum CoarseCategory { kCatNormal, kCatAbnormal, kCatConditionalAbnormal };
+enum CoarseCategory {
```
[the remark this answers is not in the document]
tetsuo-cpp: [e1f9720](https://github.com/lifting-bits/remill/pull/645/commits/e1f9720b411ccb75ca78a629919d395bf7f04875)

lib/Arch/Sleigh/ControlFlowStructuring.cpp
```diff
@@ -29,14 +32,19 @@ auto variant_cast(const std::variant<Args...> &v)
   return {v};
 }
 
-enum CoarseEffect { kAbnormal, kNormal };
+enum CoarseEffect { kAbnormal, kNormal, kIntraInstruction };
```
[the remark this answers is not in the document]
tetsuo-cpp: [e1f9720](https://github.com/lifting-bits/remill/pull/645/commits/e1f9720b411ccb75ca78a629919d395bf7f04875)

lib/Arch/Sleigh/PPCArch.cpp
```diff
@@ -0,0 +1,272 @@
...
+
+    (void) this->RegisterByName(kPCVariableName)->AddressOf(state_ptr_arg, ir);
```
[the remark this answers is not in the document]
tetsuo-cpp: [f2618e2](https://github.com/lifting-bits/remill/pull/645/commits/f2618e2b12f21890f1462e20104ec847ccf66f83)

lib/BC/PcodeCFG.cpp
tetsuo-cpp: [d1be39d](https://github.com/lifting-bits/remill/pull/645/commits/d1be39d63b0e2f42de0bad4a739c7405c360b391)

lib/BC/PcodeCFG.cpp
tetsuo-cpp: [d1be39d](https://github.com/lifting-bits/remill/pull/645/commits/d1be39d63b0e2f42de0bad4a739c7405c360b391)

lib/Arch/Sleigh/PPCArch.cpp
```diff
@@ -0,0 +1,272 @@
...
+
+    REG(TEA, signals.tea.qword, u64);
+  }
+
+  void
+  FinishLiftedFunctionInitialization(llvm::Module *module,
+                                     llvm::Function *bb_func) const override {
+    auto &context = module->getContext();
+    const auto u64 = llvm::Type::getInt64Ty(context);
+    const auto addr = u64;
```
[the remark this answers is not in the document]
tetsuo-cpp: [d1be39d](https://github.com/lifting-bits/remill/pull/645/commits/d1be39d63b0e2f42de0bad4a739c7405c360b391)

lib/BC/SleighLifter.cpp
tetsuo-cpp: It's assigned in the constructor's initialiser list already. Is this in case it gets removed from the list?

lib/BC/SleighLifter.cpp
tetsuo-cpp: [a932c11](https://github.com/lifting-bits/remill/pull/645/commits/a932c11a5d8df0c7916959980345d080b2d7c019)

lib/BC/SleighLifter.cpp
tetsuo-cpp: [4fd557e](https://github.com/lifting-bits/remill/pull/645/commits/4fd557ecdb39a18733876be6d28ea71f63b10e7e)

lib/Arch/Sleigh/PPCArch.cpp
tetsuo-cpp: [b50f203](https://github.com/lifting-bits/remill/pull/645/commits/b50f203b348ecb3efc044d836ee8634d04904ea5)

include/remill/BC/PCodeCFG.h
tetsuo-cpp: [0a88d79](https://github.com/lifting-bits/remill/pull/645/commits/0a88d79823a0e4e7ff144c175c0f8cbe4f8c2e6e)

lib/BC/PcodeCFG.cpp
tetsuo-cpp: [ed9069f](https://github.com/lifting-bits/remill/pull/645/commits/ed9069fac284272cf1778130ec1d4659dc80960c)

lib/BC/PcodeCFG.cpp
tetsuo-cpp: [ed9069f](https://github.com/lifting-bits/remill/pull/645/commits/ed9069fac284272cf1778130ec1d4659dc80960c) and [39ffa7b](https://github.com/lifting-bits/remill/pull/645/commits/39ffa7b8f169bbc3b9b8d577cc8921855404249f)

cmake/settings.cmake
```diff
@@ -29,7 +29,7 @@ macro(main)
   #
 
   # Globally set the required C++ standard
-  set(CMAKE_CXX_STANDARD 17)
+  set(CMAKE_CXX_STANDARD 20)
```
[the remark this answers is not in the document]
tetsuo-cpp: Which ones are you referring to?

lib/Arch/Sleigh/ControlFlowStructuring.cpp
tetsuo-cpp: [ddcbe39](https://github.com/lifting-bits/remill/pull/645/commits/ddcbe397f88ca11fbc6d2e93acf240f4e5611a2d)

lib/Arch/Sleigh/ControlFlowStructuring.cpp
tetsuo-cpp: [ddcbe39](https://github.com/lifting-bits/remill/pull/645/commits/ddcbe397f88ca11fbc6d2e93acf240f4e5611a2d)

lib/Arch/Sleigh/PPCArch.cpp
tetsuo-cpp: [0f8ec93](https://github.com/lifting-bits/remill/pull/645/commits/0f8ec9393382a33c6af520851652da7ec90b1af6)

tests/PPC/TestLifting.cpp
tetsuo-cpp: It looks like the only reason that Thumb does that is because it gets used in the decoding implementation and not just the tests. Since we're only using them in the tests, this seems fine right?

lib/BC/SleighLifter.cpp
tetsuo-cpp: [9c640d6](https://github.com/lifting-bits/remill/pull/645/commits/9c640d61c3ae41e1724a63b67df6dea265bffa7f)

There wasn't a function verification util, but I added one.

lib/Arch/Sleigh/ControlFlowStructuring.cpp
```diff
@@ -410,31 +428,67 @@ ControlFlowStructureAnalysis::ComputeCategory(
...
+      DLOG(ERROR)
+          << "Had an instructon with intrainstruction flow, but also decoding context updates";
+      return std::nullopt;
+    }
+  }
+
   auto flows = GetBoundContextsForFlows(ops, cc, context_updater);
 
   switch (*maybe_ccategory) {
+    case CoarseCategory::kCatNormalWithIntraInstructionFlow:
```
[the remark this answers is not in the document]
tetsuo-cpp: It just fails to decode that particular instruction but continues with the rest. I don't think it'd be good to kill here.

lib/Arch/Sleigh/ControlFlowStructuring.cpp
tetsuo-cpp: [7245171](https://github.com/lifting-bits/remill/pull/645/commits/7245171900f55e1665acbcbed385aa6b02875c06)

lib/BC/PcodeCFG.cpp
tetsuo-cpp: [ddcbe39](https://github.com/lifting-bits/remill/pull/645/commits/ddcbe397f88ca11fbc6d2e93acf240f4e5611a2d)

lib/Arch/Sleigh/PPC.h
tetsuo-cpp: [111f0b3](https://github.com/lifting-bits/remill/pull/645/commits/111f0b309f4a152c00cebdcc879130858fb1887e)

Using a type alias now so it should be clear.

include/remill/Arch/Context.h
```diff
@@ -55,6 +55,8 @@ class DecodingContext {
   uint64_t GetContextValue(const std::string &context_reg) const;
   DecodingContext PutContextReg(std::string creg, uint64_t value) const;
   DecodingContext ContextWithoutRegister(const std::string &creg) const;
+
+  const std::map<std::string, uint64_t> &GetContextValues() const;
```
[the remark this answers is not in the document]
tetsuo-cpp: [111f0b3](https://github.com/lifting-bits/remill/pull/645/commits/111f0b309f4a152c00cebdcc879130858fb1887e)

bin/lift/Lift.cpp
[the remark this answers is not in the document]
tetsuo-cpp: [96f5b36](https://github.com/lifting-bits/remill/pull/645/commits/96f5b360926d0cbc5a803b30c16f4f62b763a148)

lib/BC/PcodeCFG.cpp
tetsuo-cpp: It's small so let's just keep it as is.

lib/BC/PcodeCFG.cpp
tetsuo-cpp: What getter? I'm not following what you mean. This function doesn't look super complex either.

lib/BC/PcodeCFG.cpp
tetsuo-cpp: [16522e5](https://github.com/lifting-bits/remill/pull/645/commits/16522e558a9d425589c5145be4faf80550b6e7e3)

include/remill/BC/SleighLifter.h
tetsuo-cpp: [501636a](https://github.com/lifting-bits/remill/pull/645/commits/501636ae96dc4961ee46e8e1114954a18bc9f6a5)

include/remill/BC/PCodeCFG.h
tetsuo-cpp: [062a7e2](https://github.com/lifting-bits/remill/pull/645/commits/062a7e2ecc9d56301524f3639f0c6bdb9f8e4c46)

I think the intension should be clearer now. This class just exposes a constructor and a `Build` function now that returns a CFG (and since it's called a CFG builder, that makes sense).

We then have a `CreateCFG` helper which is short hand for constructing a `PcodeCFGBuilder` and immediately calling `Build` on it. I've moved this out of the class since it's just using the public API and there's no need for it to have access to private members, etc.

lib/BC/PcodeCFG.cpp
tetsuo-cpp: 143d5e6

lib/BC/SleighLifter.cpp
tetsuo-cpp: [9cec23d](https://github.com/lifting-bits/remill/pull/645/commits/9cec23d73bc4f9a228be3f92e3ee275141cd2987)

lib/BC/PcodeCFG.cpp
tetsuo-cpp: [806568c](https://github.com/lifting-bits/remill/pull/645/commits/806568ccd203366a2b7d8e9c23ce06ebeda62467)

lib/BC/SleighLifter.cpp
tetsuo-cpp: [afa64c3](https://github.com/lifting-bits/remill/pull/645/commits/afa64c318388068d6459e908650c178bf5dfd2e1)

lib/BC/SleighLifter.cpp
tetsuo-cpp: [71a4ddf](https://github.com/lifting-bits/remill/pull/645/commits/71a4ddf6d96b25b0b09ba908ad4b91fb82624e34)

lib/BC/SleighLifter.cpp
tetsuo-cpp: [71a4ddf](https://github.com/lifting-bits/remill/pull/645/commits/71a4ddf6d96b25b0b09ba908ad4b91fb82624e34)

include/remill/BC/PCodeCFG.h
tetsuo-cpp: [71a4ddf](https://github.com/lifting-bits/remill/pull/645/commits/71a4ddf6d96b25b0b09ba908ad4b91fb82624e34)

Renamed the variable but still needs some docs. Could you please add that @2over12?

lib/BC/PcodeCFG.cpp
tetsuo-cpp: [44d87c5](https://github.com/lifting-bits/remill/pull/645/commits/44d87c5628964c437947b4ccf0be1668db605480)

lib/Arch/Sleigh/PPCArch.cpp
tetsuo-cpp: [a689327](https://github.com/lifting-bits/remill/pull/645/commits/a6893274aad322ae67bac72af8c28cfc52a966d0)

lib/BC/PcodeCFG.cpp
tetsuo-cpp: [fa650f5](https://github.com/lifting-bits/remill/pull/645/commits/fa650f549045057c1767379fd45c5351fe6a9220)

I decided to just use `std::copy` here.

lib/BC/PcodeCFG.cpp
tetsuo-cpp: [c095e26](https://github.com/lifting-bits/remill/pull/645/commits/c095e26f02a32457b6c65574c4f01e049b5fae6a)

lib/BC/PcodeCFG.cpp
tetsuo-cpp: [b70c59d](https://github.com/lifting-bits/remill/pull/645/commits/b70c59da0de7aeca218c80dd979df05e0e33f21e) and [34f4113](https://github.com/lifting-bits/remill/pull/645/commits/34f41131c31fcaf8b5e3bcce22d7e2baacebdbef)

lib/BC/SleighLifter.cpp
tetsuo-cpp: I don't really understand the machinery of `std::visit` but it doesn't seem to accept the `operator` overloads if they're in a base class. As soon as I move the shared `Exit` visit in a base class, the compiler complains about the visitor not being exhaustive.

tetsuo-cpp:
I believe I've addressed everything so far. Any comment that I wasn't sure about or requires a second opinion has been left unresolved.

cmake/settings.cmake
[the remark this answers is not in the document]
2over12: More of a note to myself referring to this todo now that we are on CXX20 https://github.com/lifting-bits/remill/blob/65817f05b86ed7c215ea5871325ff0461a467dea/lib/Arch/Instruction.cpp#L834

lib/Arch/Sleigh/ControlFlowStructuring.cpp
```diff
@@ -60,7 +68,7 @@ CoarseFlowFromControlFlowOp(const RemillPcodeOp &op, uint64_t next_pc) {
   auto is_conditional = op.op == CPUI_CBRANCH;
   if (isVarnodeInConstantSpace(op.vars[0])) {
     // this is an internal branch.. we cant handle that right now
```
Ninja3047, line 70: comment needs update?

include/remill/BC/PCodeCFG.h
tetsuo-cpp: 3724c268b07f1f9e69645c44801a423c8a1eaec2

Thanks @2over12.

Ninja3047 (approved):
lgtm

lib/BC/SleighLifter.cpp
tetsuo-cpp: [892cada](https://github.com/lifting-bits/remill/pull/645/commits/892cadada6d35d71e58253fed4f02cf28fc497cf)

2over12 (approved):
LGTM, let's get this merged

lib/BC/SleighLifter.cpp
lkorenc: More in case someone introduces a new ctor and forgets about this attribute.

lkorenc (approved):

PR #648, opened 2023-01-12
Add Optional Param to Ignore LoadStatePointer Check

2over12:
This is pretty ugly but I don't really have a way around it without just reimplementing these utilities in anvill. In anvill, basic block functions lift instructions into a quad argument function with an additional out parameter indicating the next program counter that gets passed back up to the parent.


tetsuo-cpp:
@2over12 Sorry, that happened automatically when I renamed the branch.

PR #650, opened 2023-01-30
Weaken remill check in utils to allow additional parameters

2over12:
This PR per discussion on https://github.com/lifting-bits/remill/commit/51ed49f8cf4cbdef220ab473c653c2e33622d760, instead allows users to add additional parameters to state functions by weakening the check by default.

### DIFF
--- a/lib/BC/Util.cpp
+++ b/lib/BC/Util.cpp
@@ -239,9 +239,14 @@ FindVarInFunction(llvm::Function *function, std::string_view name_,
   return {nullptr, nullptr};
 }
 
+bool IsCompatibleStateFunction(llvm::Function *function) {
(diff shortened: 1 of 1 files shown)


lib/BC/Util.cpp
```diff
@@ -239,9 +239,14 @@ FindVarInFunction(llvm::Function *function, std::string_view name_,
   return {nullptr, nullptr};
 }
 
+bool IsCompatibleStateFunction(llvm::Function *function) {
```
lkorenc, line 242: `State` is already pretty overloaded word. Maybe `HasRemillArgPrefix`/`LiftableInto`?

lkorenc (approved):

PR #652, opened 2023-02-08
Adds Missing Case in IntraProcTransferCollector to Avoid Looping on InstrExit

2over12:
There was a missing case in the visitor causing `void operator()(const Exit &ex)` to call itself


tetsuo-cpp (approved):
LGTM.

PR #653, opened 2023-02-09
Support PowerPC syscall

tetsuo-cpp:


tetsuo-cpp:
The unit test still seems to be failing. It doesn't seem to have access to the runtime as it complains about missing `__remill_sync_hyper_call` when trying to execute it.

lib/BC/SleighLifter.cpp
```diff
@@ -1263,6 +1266,22 @@ class SleighLifter::PcodeToLLVMEmitIntoBlock {
...
         return kLiftedInstruction;
       }
+      if (other_func_name == kSysCallName &&
+          insn.arch_name == ArchName::kArchPPC) {
+        DLOG(INFO) << "Invoking syscall";
+        const auto [mem_ptr_ref, mem_ptr_ref_type] =
+            insn_lifter_parent.LoadRegAddress(
+                bldr.GetInsertBlock(), state_pointer, kMemoryVariableName);
+
+        llvm::Value *args[] = {
```
lkorenc, line 1276: `std::array` - there is no good reason to use C-style arrays in modern C++, use either `vector` or `array`.

lib/BC/SleighLifter.cpp
```diff
@@ -1263,6 +1266,22 @@ class SleighLifter::PcodeToLLVMEmitIntoBlock {
...
+            state_pointer, mem_ptr_ref,
+            llvm::ConstantInt::get(
```
lkorenc, line 1278: Extract out - look at the indents, hard to follow.

Ninja3047:
> The unit test still seems to be failing. It doesn't seem to have access to the runtime as it complains about missing `__remill_sync_hyper_call` when trying to execute it.

i think a stub implementation needs to be added here similar to how we have implementations for `__remill_read_memory_8` and others?
https://github.com/lifting-bits/remill/blob/master/test_runner_lib/TestRunner.cpp

lib/BC/SleighLifter.cpp
tetsuo-cpp: [51a495e](https://github.com/lifting-bits/remill/pull/653/commits/51a495ec193a2724467694a3b0d58d1274a50afa)

lib/BC/SleighLifter.cpp
tetsuo-cpp: [51a495e](https://github.com/lifting-bits/remill/pull/653/commits/51a495ec193a2724467694a3b0d58d1274a50afa)

lib/Arch/Runtime/HyperCall.cpp
```diff
@@ -386,6 +386,10 @@ Memory *__remill_sync_hyper_call(State &state, Memory *mem,
       mem = __remill_ppc_emulate_instruction(mem);
       break;
 
+  case SyncHyperCall::kPPCSysCall:
+      mem = __remill_ppc_syscall(mem);
```
tetsuo-cpp, line 390: Unfortunately, I don't think we can implement this with inline asm like we've been doing for the x86 and ARM syscalls. The GCC/Clang inline assembly machine constraints for PowerPC don't seem to allow you to specify a particular register.

Ninja3047 (approved):
lgtm

tetsuo-cpp:
Ping @2over12 @lkorenc. This is good to go.

lib/BC/SleighLifter.cpp
```diff
@@ -1263,6 +1267,27 @@ class SleighLifter::PcodeToLLVMEmitIntoBlock {
                                                   vars[2]);
         return kLiftedInstruction;
       }
+      if (other_func_name == kSysCallName &&
+          insn.arch_name == ArchName::kArchPPC) {
+        DLOG(INFO) << "Invoking syscall";
+
+        const auto [mem_ptr_ref, mem_ptr_ref_type] =
+            insn_lifter_parent.LoadRegAddress(
```
2over12, line 1275: Is there a reason `remill::LoadMemoryPointerRef` doesnt work here?

lib/BC/SleighLifter.cpp
```diff
@@ -1263,6 +1267,27 @@ class SleighLifter::PcodeToLLVMEmitIntoBlock {
...
+
+        // Get a LLVM value for the sync hyper call enumeration.
+        auto hyper_call_int =
+            static_cast<uint32_t>(SyncHyperCall::Name::kPPCSysCall);
+        auto hyper_call = llvm::ConstantInt::get(
+            llvm::IntegerType::get(this->context, 32), hyper_call_int);
+        std::array<llvm::Value *, 3> args = {state_pointer, mem_ptr_ref,
+                                             hyper_call};
+
+        bldr.CreateCall(insn_lifter_parent.GetIntrinsicTable()->sync_hyper_call,
```
2over12, line 1286: So is the reasoning here for why we need a custom signature for this sync call because we want to pass the type sync call explicitly? Just wondering why we cant take the strategy of other intrinsic hyper calls where it's a typical (state,PC,memory) lifted function and we can use AddCall transparently.

2over12 (dismissed):
LGTM if we can use the LoadMemoryPointerRef utility Id like to do that

lib/BC/SleighLifter.cpp
tetsuo-cpp: Yep that's correct. With `__remill_sync_hyper_call`, we need to provide a `Name` enumeration to signal what kind of sync call we're making.

lib/BC/SleighLifter.cpp
tetsuo-cpp: [5de355d](https://github.com/lifting-bits/remill/pull/653/commits/5de355d1253bd84460bad7cbb60e178c21c65421)

PR #657, opened 2023-02-24
Fixes next pc ARM/Arch in Sleigh Cbranch implementation

2over12:
The cbranch implementation loaded the original pc from the PC reg which on ARM points to the next next instruction, causing incorrect fallthrough. Changes CBRANCH to load from next-pc which points past the current instruction

### DIFF
--- a/lib/BC/SleighLifter.cpp
+++ b/lib/BC/SleighLifter.cpp
@@ -20,6 +20,7 @@
 #include <llvm/ADT/StringExtras.h>
 #include <llvm/IR/BasicBlock.h>
 #include <llvm/IR/DerivedTypes.h>
+#include <llvm/IR/IRBuilder.h>
 #include <llvm/IR/Value.h>
 #include <llvm/IR/Verifier.h>
 #include <remill/Arch/Name.h>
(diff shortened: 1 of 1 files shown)


lib/BC/SleighLifter.cpp
```diff
@@ -857,20 +858,15 @@ class SleighLifter::PcodeToLLVMEmitIntoBlock {
       auto jump_addr = this->replacement_cont.LiftOffsetOrReplace(
           bldr, lhs, this->insn_lifter_parent.GetWordType());
 
-      auto pc_reg_param = this->LiftNormalRegister(bldr, "PC");
-      CHECK(pc_reg_param.has_value());
-      auto pc_reg_ptr = *pc_reg_param;
-      auto orig_pc_value = pc_reg_ptr->LiftAsInParam(
-          bldr, this->insn_lifter_parent.GetWordType());
+      auto orig_pc_value = this->GetNextPc(bldr);
+      //CHECK(pc_reg_param.has_value());
```
Ninja3047, line 862: remove uneeded comment but otherwise looks good to me

Ninja3047 (approved):

PR #659, opened 2023-03-14
Replace the Manual AArch32 Semantics and Decoder with Sleigh

2over12:


pgoodman:
Can you delete any dead code?

2over12:
Are we sure we want to get rid of the manually implemented aarch32 semantics, this is just replacing those with a default mode of using the sleigh lifter instead. Im also still working on some issues with context registers at the moment

lkorenc:
I would remove any dead code - we use `git` if someone needs them later can just pull it out from history.

lkorenc (approved):

PR #661, opened 2023-04-04
Fix bug in rendering actual PC from NEXT_PC with thumb preffix/suffix instructions

2over12:
Since we group preffix suffix instructions using the current instruction size to work out the mode was buggy 
https://github.com/lifting-bits/remill/blob/b677387f6f7507193ddd4e3c9a2ec8bc2da94fc0/lib/Arch/Sleigh/Thumb2Arch.cpp#L48

instead this passes down the decoding context to figure out where PC should be

### DIFF
--- a/lib/Arch/Sleigh/Arch.h
+++ b/lib/Arch/Sleigh/Arch.h
@@ -99,6 +99,9 @@ struct ContextRegMappings {
 
  private:
   std::unordered_map<std::string, std::string> context_reg_mapping;
+  // Stores the size of the context register in bytes.
+  // We need to allocate space for an instruction to manipulate a
+  // Context reg as needed. This space is also populated with the incoming value.
   std::unordered_map<std::string, size_t> vnode_size_mapping;
 
  public:
--- a/lib/Arch/Sleigh/Arch.h
+++ b/lib/Arch/Sleigh/Arch.h
@@ -112,9 +112,9 @@ class SleighDecoder {
                                        const ContextValues &) const = 0;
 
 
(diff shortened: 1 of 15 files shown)


2over12:
There's an additional issue here in how we compute the patch expressions for sleigh files. We assume we can compute the address of `inst_next` given the value in PC, but we do it using inst_last, in reality regardless of the instruction size this should be -4 on thumb and -8 on arm which requires determining what mode, this will require another macro

2over12:
replacing all context reg references with a known constant is too naive, we need to just assign the entry value then lift the register like normal. A better way to do this is to allocate the local context reg as a pointer, store the assumed constant, then lift

2over12:
Should really integrate TestLifting.cpp from ARM and PPC but Im pressed for time at the moment.

lib/Arch/Sleigh/Arch.h
```diff
@@ -95,13 +95,31 @@ class SingleInstructionSleighContext {
...
+struct ContextRegMappings {
+
+ private:
+  std::unordered_map<std::string, std::string> context_reg_mapping;
+  std::unordered_map<std::string, size_t> vnode_size_mapping;
+
+ public:
+  ContextRegMappings(
+      std::unordered_map<std::string, std::string> context_reg_mapping,
+      std::unordered_map<std::string, size_t> vnode_size_mapping)
```
Ninja3047, line 107: can you put a comment somewhere that explains what vnode_size_mapping is for? looks like it's set maps to 1 for both ppc and thumb?

PR #665, opened 2023-04-11
Get Remill building with LLVM 16

tetsuo-cpp:


lib/BC/Util.cpp
```diff
@@ -1684,6 +1698,12 @@ void CloneFunctionInto(llvm::Function *source_func, llvm::Function *dest_func,
 
 
       new_insts.push_back(new_inst);
+#else
+      // TODO(alex): Figure out how we can reproduce the above hack in LLVM 16...
```
tetsuo-cpp, line 1702: I still need to figure this out.

lib/BC/Util.cpp
```diff
@@ -1339,6 +1340,17 @@ MoveConstantIntoModule(llvm::Constant *c, llvm::Module *dest_module,
             moved_c = ret;
             return ret;
           }
+#else
+          // In LLVM 16, cast is the only unary constexpr.
```
lkorenc, line 1344: This is copy & paste, de-duplicate (as far as I can tell the first branch is here twice?).

lib/BC/Util.cpp
lkorenc: Also, I would prefer to move as much of these changes as possible to `Compat/`.

lkorenc (changes requested):

lib/BC/Util.cpp
tetsuo-cpp: The difference is that there's a conditional wrapping this logic above. I could de-duplicate but it would involve having multiple `#if` checks since I'd have to wrap the condition and the `else` block. I thought it'd be better cleaner and easier to just duplicate it.

Either that or I could put this in a lambda and make the logic look like:
```c++
#if LLVM_VERSION_NUMBER < LLVM_VERSION(16, 0)
if (uop->isCast()) {
  return move_cast();
} else {
  auto ret = llvm::ConstantExpr::get(
    ce->getOpcode(),
    MoveConstantIntoModule(ce->getOperand(0), dest_module,
      value_map, type_map));
  moved_c = ret;
  return ret;
}
#else
return move_cast();
#endif
```

lib/BC/Util.cpp
tetsuo-cpp: Oh actually, I just realised something. The whole point of that hack above is to get access to a private member. We could probably just do that one more time.
- Once to get a handle on the instruction list.
- Once to mutate the instruction like we were doing before.
Not sure how I feel about that though... 🤮

pgoodman: Overall, this was terrible / evil stuff. The fundamental problem was that I was too lazy to implement cloning instructions, so I wanted to reuse `->clone` and that had issues.

pgoodman: We could just not port the hack.

pgoodman (approved):

lib/BC/Util.cpp
tetsuo-cpp: @pgoodman Hmm, what would that look like? I'm imagining some giant switch that covers every instruction type in LLVM and then recreates it in the target module. Is that what you had in mind?

pgoodman: I think that's what LLVM does. The better question is: is it worth it to support cloning from one llvm context into another llvm context. If it's not worth it, then we can just error on the code. **My opinion is that supporting this use case is not worth the effort.**

It used to be in IRENE2 that I was trying to support different context cloning because I was sharing results with a different thread, and llvm contexts and other methods are not thread-safe. I ended up running up against a wall with the cloning approach, and abandoned it in favour of a serializing to IR to a string, then deserializing in the new LLVMContext, which was much safer and worked correctly, and so this stuff in Remill largely became dead code without a usecase.

tetsuo-cpp: Ok cool, that seems like a good idea. At the moment, the Remill test suite uses this clone function helper and Anvill uses it directly. I'll try to see if I can write a helper that serializes to a string (shouldn't be hard considering that we've got one that writes it to a file) and then try to replace those usages.

pgoodman: Does the test suite actually alter the LLVM context?

tetsuo-cpp: Yes. It moves functions into their own module so they can be executed: https://github.com/lifting-bits/remill/blob/master/tests/PPC/TestLifting.cpp#L203

pgoodman: That looks like it would reuse the same context? That is, there is no explicit passing of a new context, so presumably it shares the same LLVM context with the input module.

tetsuo-cpp: Oh sorry, I thought you meant module not context. What's the consequence of that though? Seems like I'd still have trouble moving the function across modules.

pgoodman: I don't think there are any issues with cross-module cloning, as long as the llvm context is shared between them. The instruction cloning is find cross-module as long as it's not cross-context.

tetsuo-cpp: [bf59658](https://github.com/lifting-bits/remill/pull/665/commits/bf59658d796a8e6870adc7c6cefb6c61a37c3384)

Gotcha, thanks for explaining. I think we can just remove the part where we recontextualize the type and leave everything as is.

lib/BC/Util.cpp
```diff
@@ -1670,20 +1673,13 @@ void CloneFunctionInto(llvm::Function *source_func, llvm::Function *dest_func,
...
       }
 
-      // NOTE(pag): This is pretty evil, there's no reliable way to move the
-      //            type to the destination context, and there are assertions,
-      //            e.g. in `llvm::CallBase::setCalledFunction`, that
-      //            (correctly) detect that the types don't match.
-      new_inst->*REMILL_ACCESS_MEMBER(llvm, Value, VTy) =
-          RecontextualizeType(new_inst->getType(), dest_context, type_map);
-
+      new_block_builder.Insert(new_inst);
```
tetsuo-cpp, line 1681: I should be able to do this without constructing an `IRBuilder`, lemme just find the right API.

tetsuo-cpp:
Oops, still need to fix up the scripts to work with the new `cxx-common`.

lkorenc (approved):

lib/BC/IntrinsicTable.cpp
```diff
@@ -57,7 +58,9 @@ static llvm::Function *FindPureIntrinsic(llvm::Module *module,
 
   // We want memory intrinsics to be marked as not accessing memory so that
   // they don't interfere with dead store elimination.
+#if LLVM_VERSION_NUMBER < LLVM_VERSION(16, 0)
```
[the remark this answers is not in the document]
pgoodman: https://releases.llvm.org/16.0.0/docs/ReleaseNotes.html#changes-to-the-llvm-ir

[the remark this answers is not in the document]
tetsuo-cpp: Thanks, `memory(none)` it is.

PR #671, opened 2023-05-09
Fix compilation with GCC

ekilmer:
Fixes a few errors reported by GCC and not Clang

1. "non-trivial designated initializers not supported"

    - Fixed by adding array entries that were missing initializers and
      reordering

2. "Cannot bind packed field"

    - This fixes arrays missing initializers for some entries

TODO:

- [ ] Test in CI with GCC to prevent regressions


ekilmer:
Superseded by #672

PR #678, opened 2023-07-19
Fix clang 16.0.4 Compilation on Windows

Pigrecos:


pgoodman (approved):

lib/BC/Util.cpp
```diff
@@ -1229,23 +1218,17 @@ MoveConstantIntoModule(llvm::Constant *c, llvm::Module *dest_module,
       }
       case llvm::Instruction::LShr: {
```
pgoodman, line 1219: Thinking about it, probably a bunch of cases could be merged as follows:
```c++
case llvm::Instruction::Foo:
...
case llvm::Instruction::Bar: {
  if (auto op = llvm::dyn_cast<llvm::BinaryOperator>(...)) {
    ... = ConstantFoldBinaryOpOperands(
    ...
  }
}
```

scripts/build.sh
```diff
@@ -300,6 +300,10 @@ function GetLLVMVersion
       LLVM_VERSION=llvm-16
       return 0
     ;;
+    17)
```
pgoodman, line 303: Probably should drop support for 15 and 16, especially if 17-related changes were made (e.g. dropping constant expression stuff).

test_runner_lib/CMakeLists.txt
```diff
@@ -14,21 +14,21 @@
...
 set(TEST_RUNNER_INCLUDE_DIR "${CMAKE_CURRENT_SOURCE_DIR}/include")
 
-add_library(
-  test-runner
-  STATIC
-  TestRunner.cpp
-  "${TEST_RUNNER_INCLUDE_DIR}/test_runner/TestRunner.h"
-  "${TEST_RUNNER_INCLUDE_DIR}/test_runner/TestOutputSpec.h"
-)
+#add_library(
```
pgoodman, line 17: Why?

pgoodman (changes requested):

CMakeLists.txt
```diff
@@ -131,7 +165,9 @@ add_library(remill_settings INTERFACE)
 
 target_include_directories(remill_settings INTERFACE
   $<BUILD_INTERFACE:${REMILL_INCLUDE_DIR}>
-  $<INSTALL_INTERFACE:include>)
+  $<INSTALL_INTERFACE:include>
```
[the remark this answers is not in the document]
mrexodia: The `$<INSTALL_INTERFACE>` with the GNUInstallDirs is already implicitly adding when exporting the target, so you should be able to actually remove it from the target.

lib/BC/Util.cpp
```diff
@@ -1205,17 +1205,6 @@ MoveConstantIntoModule(llvm::Constant *c, llvm::Module *dest_module,
         moved_c = ret;
         return ret;
       }
-      case llvm::Instruction::Select: {
```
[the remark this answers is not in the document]
mrexodia: Yeah, they are in the process of gutting the `ConstantExpr` functionality, this is what's left in LLVM 19: https://github.com/mrexodia/llvm-headers/blob/17bb678153ba1c5c87bb17f6cd36894cb88d5c0d/19.x/llvm/IR/Constants.h#L1101-L1159

PR #682, opened 2023-08-09
Don't mark instruction as "invalid" when we can't determine a control flow category

tetsuo-cpp:
Do not merge! This is just a starting point for discussion.

At the moment, when we can't determine a control flow category (happens with instructions like `call`, `ret`, etc which have branching in the middle of their PCode on `DECOMPILE_MODE` or `didrestore`), we just skip lifting this instruction. This happens [here](https://github.com/lifting-bits/remill/blob/master/lib/BC/SleighLifter.cpp#L1729). I think that check needs to remain since we don't want to attempt to lift for a truly invalid instruction (i.e. we failed to even decode it). Therefore, the change needs to happen here so we categorise this situation differently.

This doesn't solve the problem since we now generate IR with conditional branching right in the middle of a basic block, meaning we start running into this:
```
E20230809 18:37:37.050169 2359682 Util.cpp:460] Error verifying function: Terminator found in the middle of a basic block!
```

@2over12 @Ninja3047 Did you have ideas on how to best handle this? This is a bit of an odd situation since the lifted PCode doesn't conform to a valid basic block anymore.

### DIFF
--- a/lib/Arch/Sleigh/Arch.cpp
+++ b/lib/Arch/Sleigh/Arch.cpp
@@ -299,7 +299,9 @@ SleighDecoder::DecodeInstructionImpl(uint64_t address,
     LOG(ERROR) << "Failed to compute category for inst at " << std::hex
                << inst.pc;
     inst.flows = Instruction::InvalidInsn();
-    inst.category = Instruction::Category::kCategoryInvalid;
+    // Do not mark the instruction category as "invalid". Even if we can't determine a control flow
+    // category, we still want to attempt to lift this instruction.
+    inst.category = Instruction::Category::kCategoryNormal;
     return std::make_pair(inst.flows, std::nullopt);
   }
 
--- a/lib/BC/PcodeCFG.cpp
+++ b/lib/BC/PcodeCFG.cpp
@@ -81,11 +81,15 @@ BlockExit PcodeCFGBuilder::GetBlockExitsForIndex(size_t index) const {
   CHECK(index < linear_ops.size());
   const auto &curr_op = linear_ops[index];
 
-  auto build_direct_target_exit = [](VarnodeData target,
+  auto build_direct_target_exit = [&](VarnodeData target,
                                      size_t curr_ind) -> Exit {
     if (isVarnodeInConstantSpace(target)) {
       // need to treat as signed?
       return IntrainstructionIndex{curr_ind + target.offset};
+    }
+    // we cannot assume a call pcodeop is at the end of a basic block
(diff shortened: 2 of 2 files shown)


Ninja3047:
```
; Function Attrs: alwaysinline inlinehint nounwind
define internal ptr @sleigh_remill_instruction_function_40048b80(ptr %state, ptr %memory, ptr %btaken, ptr %npc) #10 {
entry_block:
  %G1 = getelementptr inbounds %struct.State, ptr %state, i32 0, i32 0, i32 4, i32 51, i32 0, !remill_register !1352
  %DIDRESTORE = getelementptr inbounds %struct.State, ptr %state, i32 0, i32 0, i32 19, !remill_register !1394
  %O7 = getelementptr inbounds %struct.State, ptr %state, i32 0, i32 0, i32 4, i32 47, i32 0, !remill_register !1350
  %MEMORY = alloca ptr, align 4
  store ptr %memory, ptr %MEMORY, align 4
  %"unique_ff00:1" = alloca i8, align 1
  br label %1

exit_block:                                       ; preds = %11, %1, %1
  %0 = load ptr, ptr %MEMORY, align 4
  ret ptr %0

1:                                                ; preds = %entry_block
  store i32 1074039680, ptr %O7, align 4
  store i8 0, ptr %DIDRESTORE, align 1
  %2 = load i32, ptr %G1, align 4
  %3 = or i32 %2, 0
  store i32 %3, ptr %O7, align 4
  store i32 1074039860, ptr %npc, align 4
  br label %exit_block
  %4 = load i8, ptr %DIDRESTORE, align 1
  %5 = icmp eq i8 %4, 0
  %6 = zext i1 %5 to i8
  store i8 %6, ptr %"unique_ff00:1", align 1
  %7 = load i8, ptr %"unique_ff00:1", align 1
  %8 = trunc i8 %7 to i1
  %9 = load i32, ptr %npc, align 4
  %10 = select i1 %8, i32 1074039688, i32 %9
  store i32 %10, ptr %npc, align 4
  br i1 %8, label %exit_block, label %11 // terminator 1

11:                                               ; preds = %1
  %12 = load i32, ptr %O7, align 4
  store i32 %12, ptr %npc, align 4
  br label %exit_block  // terminator 2
}
```
fails verification on this block

tetsuo-cpp:
In the example I'm looking at, things are still looking broken. I'm investigating this atm but most of the code seems to be optimized out due to a poison value somewhere.

However, that's a separate issue so I think this is good to go. Marking as ready.

Ninja3047 (approved):

PR #688, opened 2023-09-26
llvm 17 fixes

2over12:


lib/BC/Optimizer.cpp
```diff
@@ -46,81 +53,39 @@ namespace remill {
...
-  builder.populateFunctionPassManager(func_manager);
-  builder.populateModulePassManager(module_manager);
-  func_manager.doInitialization();
-  llvm::Function *func = nullptr;
-  while (nullptr != (func = generator())) {
-    func_manager.run(*func);
-  }
-  func_manager.doFinalization();
-  module_manager.run(*module);
+  OptimizeBareModule(module, guide);
```
2over12, line 56: Not sure what we want to do about this i dont see an easy way to do single function inlining anymore i guess we could probably manually call llvm::InlineFunction etc but that seems excessive for something only called in the trace lifter idk

scripts/build.sh
```diff
@@ -23,11 +23,11 @@ DOWNLOAD_DIR="$( cd "$( dirname "${SRC_DIR}" )" && pwd )/lifting-bits-downloads"
 CURR_DIR=$( pwd )
 BUILD_DIR="${CURR_DIR}/remill-build"
 INSTALL_DIR=/usr/local
-LLVM_VERSION=llvm-16
+LLVM_VERSION=llvm-17
 OS_VERSION=
 ARCH_VERSION=
 BUILD_FLAGS=
-CXX_COMMON_VERSION="0.3.1"
+CXX_COMMON_VERSION="0.6.0"
```
Ninja3047, line 30:
```suggestion
CXX_COMMON_VERSION="0.6.1"
```

lib/BC/Util.cpp
```diff
@@ -1205,17 +1205,6 @@ MoveConstantIntoModule(llvm::Constant *c, llvm::Module *dest_module,
         moved_c = ret;
         return ret;
       }
-      case llvm::Instruction::Select: {
```
Ninja3047, line 1208: why are we removing this case?

Ninja3047 (approved):
build stuff looks good
just a question about why the select case is removed

lib/BC/Util.cpp
2over12: llvm 17 weakened constant expressions significantly, you cant create a select anymore

PR #689, opened 2023-10-12
Clear space and don't use cxx-common docker image

ekilmer:


.github/workflows/ci.yml
```diff
@@ -13,8 +13,6 @@ on:
       - "*"
 
   pull_request:
-    branches:
-      - "*"
```
Ninja3047, line 17: i think we want this so we have CI run when making PR's to other branches?

cmake/utils.cmake
```diff
@@ -161,7 +161,7 @@ function(InstallExternalTarget target_name target_path install_type installed_fi
 
   install(FILES "${output_file_path}"
     TYPE ${install_type}
-    PERMISSIONS OWNER_READ OWNER_EXECUTE
+    PERMISSIONS OWNER_READ OWNER_EXECUTE OWNER_WRITE
```
Ninja3047, line 164: why do we want this?

2over12 (approved):
lgtm

cmake/utils.cmake
ekilmer: There was an issue with packaging where `strip` could not rewrite the installed file which failed here https://github.com/lifting-bits/remill/actions/runs/6498034215/job/17648485113#step:6:6691
and was resolved by this change here https://github.com/lifting-bits/remill/actions/runs/6498574821/job/17650196369#step:6:6651

.github/workflows/ci.yml
ekilmer: I had to remove this to get CI to run on my PR

PR #692, opened 2024-01-04
Fix for successful build of container image.

yomaytk:
This pull request enables us to successfully build container images from the Dockerfile on both x86_64 and aarch64.
This fixed 3 files as follows.
- **Dockerfile**: Fist, the Dockerfile is dependent on `trailofbits/cxx-common-vcpkg-builder-ubuntu:${UBUNTU_VERSION}`, but the  name of deps image for aarch64 ends with `${UBUNTU_VERSION}_arm64`. Thus, we had to specify the argument when building on aarch64 (e.g. `--build-arg UBUNTU_VERSION=22.04_arm64`). To avoid this, I fixed to install all dependencies on the Dockerfile. Second, I commented out the line `CTEST_OUTPUT_ON_FAILURE=1 cmake --build . --verbose --target test -- -j $(nproc) && \` because the test has failed.
- **scritps/docker-lifter-entrypoint.sh**: remill should support the version of LLVM 16, but only 17 had been supported and therefore I added the support of LLVM 16.
- **lib/Arch/Arch.cpp**: When I build with LLVM 16, `#include <llvm/IR/AttributeMask.h>` caused and error and this is not used and therefore I removed this.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,27 +1,45 @@
-# Choose your LLVM version
-ARG LLVM_VERSION=17
-ARG ARCH=amd64
+# Choose your LLVM version (16+)
+ARG LLVM_VERSION=16
+ARG ARCH=aarch64
 ARG UBUNTU_VERSION=22.04
-ARG DISTRO_BASE=ubuntu${UBUNTU_VERSION}
-ARG BUILD_BASE=ubuntu:${UBUNTU_VERSION}
-ARG LIBRARIES=/opt/trailofbits
 
+# base ubuntu stage
+FROM ubuntu:${UBUNTU_VERSION} as base
+ARG LLVM_VERSION
+ARG ARCH
+ARG UBUNTU_VERSION
 
 # Run-time dependencies go here
-FROM ${BUILD_BASE} as base
-
-# Build-time dependencies go here
-# See here for full list of those dependencies
-# https://github.com/lifting-bits/cxx-common/blob/master/docker/Dockerfile.ubuntu.vcpkg
-FROM trailofbits/cxx-common-vcpkg-builder-ubuntu:${UBUNTU_VERSION} as deps
-ARG UBUNTU_VERSION
-ARG ARCH
-ARG LLVM_VERSION
-ARG LIBRARIES
+FROM base as deps
+RUN apt-get update && apt-get install -qqy --no-install-recommends apt-transport-https software-properties-common gnupg ca-certificates wget && \
+apt-add-repository ppa:git-core/ppa --yes
 
 RUN apt-get update && \
-    apt-get install -qqy python3 python3-pip libc6-dev wget liblzma-dev zlib1g-dev curl git build-essential ninja-build libselinux1-dev libbsd-dev ccache pixz xz-utils make rpm && \
-    if [ "$(uname -m)" = "x86_64" ]; then dpkg --add-architecture i386 && apt-get update && apt-get install -qqy gcc-multilib g++-multilib zip zlib1g-dev:i386; fi && \
+  if [ "$( uname -m )" = "x86_64" ]; then \
+    dpkg --add-architecture i386 && apt-get update && apt-get install -qqy gcc-multilib g++-multilib zlib1g-dev:i386; \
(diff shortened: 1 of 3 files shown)


ekilmer:
Hello! Thank you for pointing out the issue with the ARM images being named differently and working on a fix.

I need to think about this a bit because the initial reason we used the cxx-common base image was to keep the build environment as close as possible to how the dependencies were built, with the same compiler (exact version of LLVM), runtime libraries outside of vcpkg (like glibc), etc.

I'd like to revisit whether we can build multi-platform Docker images in cxx-common before I consider these changes.

mrexodia:
I have successfully built remill in a multi-arch image without customizing it at all: https://github.com/mrexodia/cxx-common-cmake/

You build it with this command (from an M1 mac in my case):

```
git submodule update --init
docker buildx build --platform linux/arm64,linux/amd64 -t ghcr.io/mrexodia/cxx-common-cmake:latest .
```

That being said, I do not use the `cxx-common` repository because vcpkg is a trash fire and it unnecessarily complicates things. Additionally my pure CMake approach (a superbuild to handle the dependencies) allows for easy customization where vcpkg doesn't.

yomaytk:
@ekilmer
I see. As you say, we should keep the environment as close as possible to how the dependencies were built, and we don't have to expose all dependencies on the Dockerfile if we prepare the image with another name (e.g. `trailofbits/cxx-common-vcpkg-builder-ubuntu:22.04_amd64` and `trailofbits/cxx-common-vcpkg-builder-ubuntu:22.04_arm64`). Do you have a plan to upload to dockerhub a new container image?

@mrexodia 
Thank you for the interesting proposal. Could you tell me not only the command to build `cxx-common-cmake` but also the command to build remill with it?

mrexodia:
Remill is also included in the cxx-common-cmake repository (sorry for the name confusion)

yomaytk:
@mrexodia 
Thanks. However, I am concerned that it will be tough to maintain all repositories if the dependencies on third-party projects increase.
@ekilmer 
May I have your thoughts on this?🙏

Dockerfile
```diff
@@ -43,7 +61,7 @@ RUN pip3 install ./scripts/diff_tester_export_insns
 
 RUN cd remill-build && \
     cmake --build . --target test_dependencies -- -j $(nproc) && \
-    CTEST_OUTPUT_ON_FAILURE=1 cmake --build . --verbose --target test -- -j $(nproc) && \
+    # CTEST_OUTPUT_ON_FAILURE=1 cmake --build . --verbose --target test -- -j $(nproc) && \
```
ekilmer, line 64: This should be guarded by a check on the architecture. We use this in CI to test that the Docker build continues to work, at least for x86_64.

ekilmer:
@yomaytk I just opened #695 to test the new way (https://github.com/lifting-bits/cxx-common/pull/1045) we build Docker images in cxx-common.

Can you try my PR and let me know if it works for you or not?

As for your other fixes, I think we can include them, but it would be great if you could update this PR (or make a new one) to be a bit more organized.

Thank you!

----------

Also, @mrexodia, I agree that vcpkg isn't perfect, and we try to work around or document the issues and confusing aspects.

I am curious about what specific aspects of vcpkg cause you the most headaches for the workflows or issues you have when developing these tools (or others). Maybe we can document those better or potentially smooth out the experience?

For our situation, we use vcpkg primarily for binary caching during our CI runs. Local developers almost always build the vcpkg dependencies from source so that they have the Debug versions as well. We've also had decent success using the ASAN triplets to catch more issues.

Also, your cxx-common-cmake repo is a great reference for managing a superbuild with CMake. Thank you! CMake superbuilds are the best solution for having absolute control over the dependencies and are certainly better if/when you need to fix/add/debug something in your dependencies.

Dockerfile
ekilmer: Made this change in #695 👍

tamaroning:
I got an error for this. after removing this error, build succeeded. thank you.

> lib/Arch/Arch.cpp: When I build with LLVM 16, #include <llvm/IR/AttributeMask.h> caused and error and this is not used and therefore I removed this.

PR #697, opened 2024-01-12
Lift the declared branch taken store for a given flow

### LINKED ISSUE #694: LiftCBranch does not set BRANCH_TAKEN
Hello,
I've been implementing a MIPS backend via Sleigh and noticed the conditional blocks are not evaluating branch_taken correctly because it's never stored.
It looks to me that SPARC/AArch etc do have some own handling so this might've caused a oversight?
I've seen the same happen with the PPC backend, which my MIPS backend is based on.

I have this as a fix right now in LiftCBranch, but given that LiftBranchTaken sets it as well I am not sure if this is the appropriate way, it might just need some minor refactoring (given LiftBranchTaken is currently a stub since its only called via LiftBtakenIfReached which is never invoked)?

image
```cpp
      auto orig_pc_value = this->GetNextPc(bldr);
      //CHECK(pc_reg_param.has_value());
      auto next_pc_value =
          bldr.CreateSelect(trunc_should_branch, jump_addr, orig_pc_value);
      auto should_branch_value =
          bldr.CreateSelect(trunc_should_branch, 
          llvm::ConstantInt::get(llvm::Type::getInt8Ty(context), 1),
          llvm::ConstantInt::get(llvm::Type::getInt8Ty(context), 0));

      bldr.CreateStore(should_branch_value, this->GetBranchTakenRef());
      bldr.CreateStore(next_pc_value, this->GetNextPcRef());
  ```

Here how it looks with my MIPS backend with the change:
image

Before the change (Pulling uninitialized memory from stack):
image

### PR DESCRIPTION
**Pull request by 2over12**
Anvill does not use branch taken for intraprocedural flows and instead [just switches on PC ](https://github.com/lifting-bits/anvill/blob/70209a8c3311cc97875605a137da210233fe9cd6/lib/Lifters/BasicBlockLifter.cpp#L405). Unfortunately we do rely on the btaken variable for a flow being lifted in the case of a conditional call ie. https://github.com/lifting-bits/anvill/blob/70209a8c3311cc97875605a137da210233fe9cd6/lib/Lifters/BasicBlockLifter.cpp#L232 

Regardless since we compute these flow hints we should keep them consistent in pcode. We should test this against conditional control flow in anvill. Closes #694

### DIFF
--- a/lib/BC/SleighLifter.cpp
+++ b/lib/BC/SleighLifter.cpp
@@ -410,8 +410,6 @@ class SleighLifter::PcodeToLLVMEmitIntoBlock {
   llvm::BasicBlock *entry_block;
   llvm::BasicBlock *exit_block;
 
-  size_t curr_id;
-
   const sleigh::MaybeBranchTakenVar &to_lift_btaken;
 
   std::unordered_map<size_t, llvm::BasicBlock *> start_index_to_block;
@@ -466,7 +464,6 @@ class SleighLifter::PcodeToLLVMEmitIntoBlock {
         user_op_names(user_op_names_),
         entry_block(target_block),
         exit_block(exit_block_),
-        curr_id(0),
         to_lift_btaken(to_lift_btaken_),
         context_reg_lifter(std::move(context_reg_lifter)) {}
 
@@ -1393,9 +1390,9 @@ class SleighLifter::PcodeToLLVMEmitIntoBlock {
   }
 
 
-  void LiftBtakenIfReached(llvm::IRBuilder<> &bldr, OpCode opc) {
+  void LiftBtakenIfReached(llvm::IRBuilder<> &bldr, OpCode opc, size_t index) {
 
-    if (this->to_lift_btaken && curr_id == this->to_lift_btaken->index) {
+    if (this->to_lift_btaken && index == this->to_lift_btaken->index) {
       this->UpdateStatus(this->LiftBranchTaken(bldr, *this->to_lift_btaken),
                          opc);
     }
@@ -1448,8 +1445,11 @@ class SleighLifter::PcodeToLLVMEmitIntoBlock {
     // we have a problem with block terminators where a cbranch <relative> -> fallthrough, need to either exit to the exit block
     // or transfer to a block. So really our cfg needs to tell us how to terminate a block
     // either exit (means real control flow), to block (fake control flow)
+    size_t index = 0;
     for (auto pc : blk.ops) {
+      this->LiftBtakenIfReached(bldr, pc.op, index);
       this->LiftPcodeOp(bldr, pc.op, pc.outvar, pc.vars.data(), pc.vars.size());
+      index += 1;
     }
 
     this->TerminateBlock();

### CODE REVIEW

**Review by tetsuo-cpp — APPROVED** [review]
LGTM!

**Comment by tetsuo-cpp** [comment]
> Regardless since we compute these flow hints we should keep them consistent in pcode. We should test this against conditional control flow in anvill. Closes https://github.com/lifting-bits/remill/issues/694

I won't merge this yet since it sounds like it still needs testing.

**Comment by m4xw** [comment]
Hi, just saw this PR and gave it a quick spin:
image
Looks good so far, the double conditions I had are gone now too, codegen is surprisingly quite a bit different in how it merges the tails now, bit weird but still looks correct on a quick look.

Also I have been meaning to take a look at anvill, quick question here: does it work with the Sleigh based backends in remill? What are the benefits of using anvill?

My use case will be for emulator AOT recompilation for games, so I am not sure how much i will benefit from it

**Comment by 2over12** [comment]
Conditional call in libc seems to work anvill side:
```
define i8 @func949680basic_block949684_22(ptr %stack, i32 %program_counter, ptr noalias nocapture %memory, ptr noalias nocapture %D12, ptr noalias nocapture %D8, ptr noalias nocapture %R1, ptr noalias nocapture %D13, ptr noalias nocapture %R9, ptr noalias nocapture %D14, ptr noalias nocapture %D15, ptr noalias nocapture %D11, ptr noalias nocapture %LR, ptr noalias nocapture %D9, ptr noalias nocapture %R10, ptr noalias nocapture %R0, ptr noalias nocapture %R8, ptr noalias nocapture %R6, ptr noalias nocapture %R5, ptr noalias nocapture %R3, ptr noalias nocapture %R4, ptr noalias nocapture %R2, ptr noalias nocapture %D10, ptr noalias nocapture %R11, ptr noalias nocapture %R7) local_unnamed_addr #7 !__anvill_basic_block_uid_md !2 {
sleigh_remill_instruction_function_e7db8.exit:
  %0 = load i32, ptr %LR, align 4
  %1 = load i32, ptr %R11, align 4
  %2 = icmp ne i32 %1, 0
  %3 = icmp ne i32 %0, 0
  %narrow.not = select i1 %2, i1 %3, i1 false
  br i1 %narrow.not, label %func949680basic_block949684_22lowlift.exit, label %.critedge

.critedge:                                        ; preds = %sleigh_remill_instruction_function_e7db8.exit
  %4 = call i8 @sub_e732c__AvB_B_0()
  br label %func949680basic_block949684_22lowlift.exit

func949680basic_block949684_22lowlift.exit:       ; preds = %sleigh_remill_instruction_function_e7db8.exit, %.critedge
  %5 = tail call i8 @func949680basic_block949700_23(ptr %stack, i32 %program_counter, ptr nonnull %memory, ptr nonnull %D12, ptr nonnull %D8, ptr %R1, ptr nonnull %D13, ptr nonnull %R9, ptr nonnull %D14, ptr nonnull %D15, ptr nonnull %D11, ptr nonnull %LR, ptr nonnull %D9, ptr nonnull %R10, ptr %R0, ptr nonnull %R8, ptr nonnull %R6, ptr nonnull %R5, ptr %R3, ptr nonnull %R4, ptr %R2, ptr nonnull %D10, ptr nonnull %R11, ptr nonnull %R7)
  ret i8 %5
}
```

2over12:
> Hi, just saw this PR and gave it a quick spin: image Looks good so far, the double conditions I had are gone now too, codegen is surprisingly quite a bit different in how it merges the tails now, bit weird but still looks correct on a quick look.
> 
> Also I have been meaning to take a look at anvill, quick question here: does it work with the Sleigh based backends in remill? What are the benefits of using anvill?
> 
> My use case will be for emulator AOT recompilation for games, so I am not sure how much i will benefit from it

Anvill does use the sleigh remill backends. I dont think anvill is likely to be a good fit for your usecase because we cannot guarantee consistently correct recompilation inside of anvill. Anvill is trying to use brightening to produce simplified bitcode, similar to what you would see coming out of clang or the C written by a human. Since we are doing decompilation in anvill there are fundamental limitations that mean it cannot always succeed or recompile.

m4xw:
> Anvill does use the sleigh remill backends. I dont think anvill is likely to be a good fit for your usecase because we cannot guarantee consistently correct recompilation inside of anvill. Anvill is trying to use brightening to produce simplified bitcode, similar to what you would see coming out of clang or the C written by a human. Since we are doing decompilation in anvill there are fundamental limitations that mean it cannot always succeed or recompile.

I see, thanks. Further down the line i plan to use this in some decompilation projects too but so far its on the backburner, might be interesting if it manages to yeet the state structs.

Btw minor thing, for PPC I use custom sleigh definitions and the current ghidra-fork and src handling (with the generated patches) chokes if new definitions (files) are introduced that dont exist in the original ghidra sourcetree

Any recommendation you have to work around this / want me to create a separate issue?

I'd like to avoid having to keep syncing changes across 2 ghidra codebases

PR #699, opened 2024-02-13
Fix big endian type conversions for registers

2over12:
In the sleigh lifter parameters are sometimes lifted as word type, regardless of their varnode size in order to produce an address (ie. at CPUI_LOAD/STORE). Previously this was transparently handled by the Register param loading. This doesnt really work on big endian systems since we will load the high bits. We now insert explicit integer<->integer conversion

### DIFF
--- a/lib/BC/SleighLifter.cpp
+++ b/lib/BC/SleighLifter.cpp
@@ -141,7 +141,7 @@ class SleighLifter::PcodeToLLVMEmitIntoBlock {
                                                        llvm::Type *ty) {
       auto mayberes = this->LiftAsInParamNoConvert(bldr, vnode_type);
       if (!mayberes) {
-        return mayberes;
+        return std::nullopt;
       }
 
       auto res = *mayberes;


lib/BC/SleighLifter.cpp
```diff
@@ -125,11 +126,40 @@ llvm::Value *CreatePcodeBitShift(llvm::Value *lhs, llvm::Value *rhs,
...
+          allows_conversions(allows_conversions) {}
+
     virtual ~Parameter(void) = default;
 
     virtual std::optional<llvm::Value *> LiftAsInParam(llvm::IRBuilder<> &bldr,
-                                                       llvm::Type *ty) = 0;
+                                                       llvm::Type *ty) {
+      auto mayberes = this->LiftAsInParamNoConvert(bldr, vnode_type);
+      if (!mayberes) {
+        return mayberes;
```
Ninja3047, line 144:
```suggestion
        return std::nullopt;
```

Ninja3047 (approved):

PR #702, opened 2024-02-28
Fix placement of 32 bit regs on e500 and populate the low bits of PC on BE

2over12:
This fixes a type error related to storing the next pc in the pc reg after the original BE fix. Also fixes the location of the _r[N] regs on the e500


m4xw:
Isnt that just basically reintroducing DoubleReg https://github.com/lifting-bits/remill/blob/608acf8edde0cbb50f7b0a036c3b2bea5442514f/include/remill/Arch/SPARC32/Runtime/State.h#L32 ?
Should maybe also streamline the names into hi/lo ?

2over12:
@Ninja3047 i think the testing framework reads/writes registers in the wrong endianness as well.. this is going to be a pain to fix

Ninja3047 (approved):

PR #703, opened 2024-03-03
Fixed windows build errors

cctv130:


include/remill/Arch/Arch.h
```diff
@@ -59,6 +60,7 @@ class IntegerType;
 class LLVMContext;
 class Module;
 class PointerType;
+class AttributeMask;
```
pgoodman, line 63: Why do you forward declare it if you include it above? If a forward declaration is sufficient then that is the preferred solution for this header, as `Arch.h` is included in a lot of places.

lib/Arch/Arch.cpp
```diff
@@ -30,6 +30,7 @@
 #include <llvm/IR/Metadata.h>
 #include <llvm/IR/Module.h>
 #include <remill/Arch/ArchBase.h>  // For `Arch` and `ArchBase`.
+#include <llvm/IR/AttributeMask.h>
```
pgoodman, line 33: Re-order to keep includes in alphabetic order.

lib/BC/InstructionLifter.h
```diff
@@ -51,6 +51,10 @@
 #include "remill/BC/Util.h"
 #include "remill/OS/OS.h"
 
+namespace llvm{
+  class Type;
```
pgoodman, line 55: The stylistically conformant formatting would be as follows:
```c++
namespace llvm {
class Type;
}  // namespace llvm
```

In general though, this shouldn't be needed because `Util.h` is included, and there is an identical forward declaration there.

pgoodman (changes requested):

pgoodman:
Can you turn this into a different issue or PR? Perhaps the `pack` pragma is needed?

PR #714, opened 2025-05-12
Fix breakpoint instruction on AArch64.

yomaytk:
## Description

Hello! `Arm® Architecture Reference Manual for A-profile architecture` explains about the breakpoint instructions as follows.
> The breakpoint instruction in the A64 instruction set is BRK #<immediate>. It is unconditional.
The breakpoint instruction in the A32 and T32 instruction sets is BKPT #<immediate>.

When I built Remill on an arm64 host, the following error occurred.

```bash
<inline asm>:1:2: error: unrecognized instruction mnemonic
        bkpt
        ^
error: cannot compile inline asm
```
Then, this PR changes the arm64 breakpoint instruction from `bkpt` to `brk #0`.

### DIFF
--- a/lib/Arch/Runtime/HyperCall.cpp
+++ b/lib/Arch/Runtime/HyperCall.cpp
@@ -282,7 +282,18 @@ Memory *__remill_sync_hyper_call(State &state, Memory *mem,
       mem = __remill_aarch64_emulate_instruction(mem);
       break;
 
-    case SyncHyperCall::kAArch64Breakpoint: asm volatile("bkpt" :); break;
+    case SyncHyperCall::kAArch64Breakpoint:
+#  ifdef __has_builtin
+#    if __has_builtin(__builtin_debugtrap)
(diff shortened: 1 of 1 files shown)


pgoodman:
I think a better choice than the inline asm would be as follows:
```c++
#ifdef __has_builtin
# if __has_builtin(__builtin_debugtrap)
__builtin_debugtrap();
#   define REMILL_USED_DEBUG_TRAP
# endif
#endif
#ifndef REMILL_USED_DEBUG_TRAP
asm(...);
#endif
#undef REMILL_USED_DEBUG_TRAP
```

yomaytk:
Thanks. Ok, I have made the changes you suggested.

pgoodman:
CC @Ninja3047. I don't have reviewer privileges here anymore.

Ninja3047 (approved):

Ninja3047:
Thanks

PR #715, opened 2025-05-19
Update the some artifact versions of workflows/ci.yml.

yomaytk:
## Description

Hi. This PR fixed the github actions workflows running error: https://github.com/lifting-bits/remill/pull/714/checks

1. Update some artifact versions.
2. `actions/checkout@v4` occurred the following error when we use the same artifact name for the same workflows. To avoid the error, I added `matrix.compiler.CC` to the artifact name of the `build_linux` job.
```
Error: Failed to CreateArtifact: Received non-retryable error: Failed request: (409) Conflict: an artifact with this name already exists on the workflow run
```

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -40,7 +40,7 @@ jobs:
         password: ${{ secrets.GITHUB_TOKEN }}
 
     steps:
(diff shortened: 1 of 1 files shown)


yomaytk:
@Ninja3047 Could you review?

PR #716, opened 2025-07-03
Pass undef if semantic function argument is unused

momo5502:
Using `IntendedArgumentType` to resolve the intended argument type fails if the argument has no uses.
This is for example the case for the destination register of the `FLD` instruction semantics: https://github.com/lifting-bits/remill/blob/d13d43f81a6fc927c34cea3d2d71737e65e3dcfc/lib/Arch/X86/Semantics/X87.cpp#L119

Not being able to resolve the intended type will then lead to the destination register being treated as source.
This in turn causes an assertion to be thrown.

As the argument is unused though, it doesn't matter what's passed. Therefore passing `undef` will work just fine.


Ninja3047 (approved):
makes sense, thanks!

PR #717, opened 2025-07-03
Update settings.cmake

cctv130:
Fix the issue where debugging on Windows in VSCode doesn’t work.


mrexodia:
Ah my bad, I only saw the change in the prefix path 🤦🏻

Ninja3047 (approved):

PR #720, opened 2025-08-19
Remove duplicate flag -fPIC

cctv130:


mrexodia:
You can use [`CMAKE_POSITION_INDEPENDENT_CODE`](https://cmake.org/cmake/help/latest/variable/CMAKE_POSITION_INDEPENDENT_CODE.html) instead.

PR #721, opened 2025-08-22
CMake: update to C++20, clean up properties, optimize fetch content

cctv130:
- Enable C++20 on both Windows and Linux
   - Remove unnecessary set_property calls to silence LOCATION warnings
   - Replace duplicate -fPIC flags with INTERFACE_POSITION_INDEPENDENT_CODE
   - Set FETCHCONTENT_BASE_DIR to speed up sleigh and ghidra-fork downloads/builds
   - fixed:Does not match the generator used previously: Unix Makefiles,use ninja instead


CMakeLists.txt
```diff
@@ -82,7 +82,7 @@ set(REMILL_BUILD_SEMANTICS_DIR_SPARC64 "${CMAKE_CURRENT_BINARY_DIR}/lib/Arch/SPA
 set(REMILL_BUILD_SEMANTICS_DIR_PPC64_32ADDR "${CMAKE_CURRENT_BINARY_DIR}/lib/Arch/PPC/Runtime")
 set(REMILL_INCLUDE_DIR "${CMAKE_CURRENT_SOURCE_DIR}/include")
 set(REMILL_LIB_DIR "${CMAKE_CURRENT_SOURCE_DIR}/lib")
-
+set(FETCHCONTENT_BASE_DIR ${CMAKE_CURRENT_SOURCE_DIR}/deps)
```
mrexodia, line 85: image

If you set it, use a CACHE variable (with the original default value) so it becomes visible in the CMake GUI and users can edit it. If you want to put it in a `deps` folder here, at the very least it should be added to `.gitignore` as well.

xray服务器端
```diff
@@ -0,0 +1,95 @@
+{
```
mrexodia, line 1: ?

mrexodia: Everything else looks good!

ekilmer (approved):
LGTM also! Thank you for the changes @cctv130, and thank you for the review @mrexodia

PR #723, opened 2025-10-26
Major refactor

### LINKED ISSUE #573: windows下构建的问题
因为我在构建中出现了很大的问题
clang-cl : error : no such file or directory: '/external:W0' [G:\mcsema\remill_build\CMakeFiles\3.22.1\CompilerIdC\CompilerIdC.vcxproj]
我在互联网上找不到解决方法

### LINKED ISSUE #663: C++ header file not found
I try to build the source code on mac, and output a lot of warning about c++ header not found. How can I fix it, thx !

### LINKED ISSUE #718: Readme suggest LLVM 15+ but it seems only to work till LLVM 17
Hello,

when building the docker container it fails after using something higher than llvm 17.
It is unclear to me if it is supposed to work with something like llvm 20.
Could you either give clear instructions in the README if its only working till llvm 17 or provide some guidance on how to use a higher version?

I am also curios since my setup doesn't really allow me to use Ubuntu if you see any problems when using static compiled libraries on another distro?

Thank you

mrexodia:
This is a huge refactor doing the following things:
- [x] Add an alternative `cxx-common` implementation as a superbuild project. You can find more information here: https://github.com/LLVMParty/packages/blob/main/dependencies.md. This is of course optional, but makes it _much_ easier to users to build the dependencies necessary for remill.
- [x] Cross-compile the semantics instead of the current hack with `g++-multilib` and `-m32`. The semantics are now compiled as `-ffreestanding` and just enough of the C++ standard library was implemented in a fake `--sysroot` to make it work.
- [x] Change how FPU exceptions are handled. Previously they would call `std::fetestexcept` and directly use `FE_XXX` flags. This has now been decoupled properly so you can implement FPU exceptions however you want. It looks like #636 was discussing similar bugs, all of which have been fixed (I think).
- [x] Properly handle patches for `sleigh` without a separate Ghidra fork for PPC
- [x] Fix a few bugs in semantics uncovered by porting to `-ffreestanding` and test failures with later LLVM versions
- [x] The project now builds and runs on LLVM 15-21. There are some `#ifdef`s for LLVM versions, but I think it would be nice to widen the compatibility instead of only supporting LLVM 17. Due to bugs in LLVM JSON this required switching back to C++17, but the project uses no C++20 features, so I don't think this is an issue.
- [x] Build and run on Windows (clang and clang-cl)
- [x] Fix `-fsanitize` (@fvrmatteo's work)
- [x] Properly document the superbuild dependency management solution in the README. This pattern is poorly documented, but it solves essentially all problems with CMake dependency management and keeps things flexible for integration into package managers etc. 
- [x] New GitHub Actions CI that builds dependencies from scratch (except LLVM, which is installed by the package manager)

Closes #678
Closes #718
Closes #663
Closes #573

Unfortunately splitting this up is going to be extremely tough. I can split the changes for LLVM15-21 compatibility, but the `-ffreestanding` changes are dependent on the FPU semantic changes and a bunch of CMake stuff. The sleigh "changes" (nothing actually changed) could be split out too, the diff just looks massive.

**Note**: The CI failure is something I cannot reproduce locally with any LLVM version. I suspect it is specific to the LLVM fork used in `cxx-common`. The docker images I use are official LLVM versions without changes, compiled with assertions enabled.


kyle-elliott-tob (approved):
Thanks for all the hard work over the past few days, LGTM!

Waiting on @wizardengineer for review

wizardengineer (approved):
LGTM, thanks again!

mrexodia:
> I compiled using the main branch

This is not a supported configuration, so there is no reason to try this. There also is **no** `-m32` flag anymore, so you did something wrong. Newer XED versions are not supported and cause lifting problems, so I use the same one as the original cxx-common.

PR #729, opened 2025-11-24
Fix signature of memory_f80 intrinsics

mrexodia:
This fixes a longstanding issue where the 80-bit floats would not be passed correctly to the `memory_f80` functions.

After #723 the semantics are compiled as `x86_64-none-elf` with `-ffreestanding` (fully OS independent) and `-mlong-double-80` to support 80-bit floats in the generated IR. The idea is that the instruction semantics themselves are 'pure' LLVM IR that we can freely retarget for our purposes without having to worry about ABI changes. This works because the `State` structure is packed and we use `uintXX_t` in the helper signatures.

The `native_float80_t` appears to have been designed to convert between `struct float80_t { uint8_t data[10]; }` and whatever is the widest `long double` available on the target platform. Unfortunately this type was used in the intrinsics:

```cpp
[[gnu::used]] extern Memory *__remill_read_memory_f80(Memory *, addr_t,
                                                      native_float80_t &);
[[gnu::used]] extern Memory *
__remill_write_memory_f80(Memory *, addr_t, const native_float80_t &);
```

This causes an ABI break of sorts, because the intrinsics might be targeting Windows (MSVC) which does not support `long double`. We should instead pass `float80_t`, which has a known layout across all target triples and the intrinsic implementation for that platform is responsible for handling conversions (using `native_float80_t` if they desire).

In practice this likely would not matter much, since the higher 2 bytes are for additional precision and they do not represent any 'meaningful' bits of the float. If your platform does not support 80-bit floats it's just not the same as native x87 operations, but with this change you can (in theory) use a softfloat implementation to get perfect semantics.


mrexodia:
It would apply to MSVC/clang-cl/clang on Windows, since long double isn't supported when targeting Windows at all.

PR #732, opened 2025-11-25
Included doxygen for better code documentation

wizardengineer:


kyle-elliott-tob (changes requested):
We should also actually attempt to document some of the various classes and methods with a PR like this

PR #749, opened 2026-02-06
TestRunner: respect PC bitwidth

redthing1:
don't hardcode the pc bitwidth, instead check the type


mrexodia:
Same question, when is this a problem?

redthing1:
I made this series of patches while working on a RISCV arch support via sleigh

PR #750, opened 2026-02-06
Int: provide int128 typedef fallback

redthing1:
tiny fix: provide a `__int128` fallback


mrexodia:
Under what circumstances does this path trigger exactly?

PR #770, opened 2026-03-26
Change XGETBV architecture modeling to hypercall

mrexodia:
See #766

### DIFF
--- a/include/remill/Arch/Runtime/HyperCall.h
+++ b/include/remill/Arch/Runtime/HyperCall.h
@@ -30,6 +30,7 @@ class SyncHyperCall {
     kX86CPUID,
     kX86ReadTSC,
     kX86ReadTSCP,
+    kX86XGetBV,
     kX86LoadGlobalDescriptorTable,
     kX86LoadInterruptDescriptorTable,
     kX86LoadAccessRights,
--- a/include/remill/Arch/X86/Runtime/State.h
+++ b/include/remill/Arch/X86/Runtime/State.h
@@ -445,34 +445,6 @@ struct alignas(8) ArithFlags final {
 
 static_assert(16 == sizeof(ArithFlags), "Invalid packing of `ArithFlags`.");
 
-union XCR0 {
-  uint64_t flat;
-
-  struct {
-    uint32_t eax;
-    uint32_t edx;
-  } __attribute__((packed));
-
-  // Bits specify what process states should be saved.
-  struct {
-    uint64_t x87_fpu_mmx : 1;  // Must be 1; bit 0.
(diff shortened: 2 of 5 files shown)


include/remill/Arch/Runtime/HyperCall.h
```diff
@@ -62,6 +62,7 @@ class SyncHyperCall {
     kX86SysCall,
     kX86SysEnter,
     kX86SysExit,
+    kX86XGetBV,
```
kyle-elliott-tob: Should probably place this near ReadTSC/ReadTSCP

include/remill/Arch/X86/Runtime/State.h
```diff
@@ -794,10 +766,10 @@ struct alignas(16) X86State : public ArchState {
   X87Stack st;  // 128 bytes.
   MMX mmx;  // 128 bytes.
   FPUStatusFlags sw;  // 24 bytes
-  XCR0 xcr0;  // 8 bytes.
   FPU x87;  // 512 bytes
   SegmentCaches seg_caches;  // 96 bytes
   K_REG k_reg; // 128 bytes.
+  uint8_t _padding[8]; // 8 bytes
```
kyle-elliott-tob: Test case failures are due to the reordering of fields with this, turns out it does actually matter :P

kyle-elliott-tob: Also we should have a small comment documenting the drop of support for CR0. Should likely revisit this later, as if we're to add back support for control registers, should probably support them properly this time

kyle-elliott-tob (changes requested):

include/remill/Arch/Runtime/HyperCall.h
mrexodia: I put it at the end to not break all the indices, but if you don't care about that I can move it down np.

include/remill/Arch/X86/Runtime/State.h
mrexodia: Why does it matter exactly? Shouldn't that be fixed instead?

kyle-elliott-tob: x87 FPU reads are misaligned otherwise. Starts reading into FPUStatusFlags for whatever reason (haven't found exactly where yet)

include/remill/Arch/Runtime/HyperCall.h
kyle-elliott-tob: Yeah, I think it's fine. Rather have some consistency and breaking changes if we're all going to continue actively working on it

kyle-elliott-tob (approved):

PR #774, opened 2026-04-20
Add REMILL_FETCH_SLEIGH option to allow find_package(sleigh)

mrexodia:
The goal is to allow the user to consume remill with `add_subdirectory(remill)` without having to clone+patch+compile sleigh constantly. This is especially useful when using an agent, because it can see/modify remill directly as it works on your lifter.


kyle-elliott-tob (approved):
LGTM

PR #777, opened 2026-05-22
Minor integration fixes

mrexodia:
Ran into these issues when trying to use remill as a submodule in my template with `REMILL_ENABLE_TESTING=OFF` and a newer clang that spams warnings about this deprecated literal operator.


kyle-elliott-tob (approved):
New behavior from P0588R1, LGTM

PR #779, opened 2026-05-24
fix(deps): patch LLVM build for GCC 15 compatibility

moleium:
Fixes LLVM 17 compilation failures on GCC 15+ toolchains due to a missing `<cstdint>` include in `SmallVector.h` (`error: use of undeclared identifier 'uint64_t'`). Caused by [GCC 15 header dependency changes](https://gcc.gnu.org/gcc-15/porting_to.html#header-dep-changes) in `libstdc++`

Adds a CMake script and updates `dependencies/llvm.cmake` with a `PATCH_COMMAND` to inject `#include <cstdint>` into `SmallVector.h` during extraction.

### DIFF
--- a/dependencies/llvm.cmake
+++ b/dependencies/llvm.cmake
@@ -53,6 +53,8 @@ ExternalProject_Add(llvm
         ${LLVM_URL}
     URL_HASH
         "SHA256=${LLVM_SHA256}"
+    PATCH_COMMAND
+        ${CMAKE_COMMAND} -P "${CMAKE_CURRENT_LIST_DIR}/patches/patch_llvm_cstdint.cmake"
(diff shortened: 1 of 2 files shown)


kyle-elliott-tob:
Thanks but this is an extremely fragile patch we won't be accepting. Use you distro's LLVM package with DUSE_EXTERNAL_LLVM that likely already contains a patch, or try and land something upstream.

moleium:
> Thanks but this is an extremely fragile patch.

Since this turned out to be broader than `SmallVector.h`, so I agree this patch is too brittle to carry.
 I’ll see whether this can be addressed upstream in LLVM instead, thanks.

PR #783, opened 2026-08-21
fix TraceLifter not terminating on self-recursive calls

### LINKED ISSUE #782: TraceLifter never terminates on a direct call to the trace's entry address
`remill-lift` hangs and allocates until it's OOM-killed whenever a trace contains a direct call back to its entry address (i.e. any self-recursive function). You can reproduce the issue by following these steps:

1. Create `main.c`:

```c
int f(int n) {
    return !n ? 1 : n * f(n - 1);
}

int main(void) {
    return f(5);
}
```

2. Build `main.c`:

```bash
docker run --rm --entrypoint 'env' -w '/mnt' -v "${PWD}:/mnt" 'ghcr.io/lifting-bits/remill/llvm22-ubuntu24.04:v6.0.1' \
    clang-22 -O'1' -f'no-optimize-sibling-calls' ./main.c
```

3. Disassemble `f` from `a.out`:

```bash
docker run --rm --entrypoint 'env' -w '/mnt' -v "${PWD}:/mnt" 'ghcr.io/lifting-bits/remill/llvm22-ubuntu24.04:v6.0.1' \
    llvm-objdump-22 --disassemble-symbols='f' ./a.out
```

```console
./a.out:	file format elf64-x86-64

Disassembly of section .text:

0000000000001130 <f>:
    1130: 85 ff                        	testl	%edi, %edi
    1132: 74 13                        	je	0x1147 <f+0x17>
    1134: 53                           	pushq	%rbx
    1135: 8d 47 ff                     	leal	-0x1(%rdi), %eax
    1138: 48 89 fb                     	movq	%rdi, %rbx
    113b: 89 c7                        	movl	%eax, %edi
    113d: e8 ee ff ff ff               	callq	0x1130 <f>
    1142: 0f af c3                     	imull	%ebx, %eax
    1145: 5b                           	popq	%rbx
    1146: c3                           	retq
    1147: b8 01 00 00 00               	movl	$0x1, %eax
    114c: c3                           	retq
    114d: 0f 1f 00                     	nopl	(%rax)
```

4. Lift `f`:

```bash
docker run --rm --entrypoint 'env' -w '/mnt' -v "${PWD}:/mnt" -m '1g' 'ghcr.io/lifting-bits/remill/llvm22-ubuntu24.04:v6.0.1' \
    remill-lift-22 -os 'linux' -arch 'amd64' -entry_address '0x1130' -address '0x1130' -bytes '85ff7413538d47ff4889fb89c7e8eeffffff0fafc35bc3b801000000c30f1f00' -bc_out /dev/null
```

5. Check process exit code:

```bash
echo "${?}"
```

```console
137
```

t0rr3sp3dr0:
`kCategoryDirectFunctionCall` and `kCategoryConditionalDirectFunctionCall` push their target onto the trace work list unconditionally. When that target is the trace's own entry, the trace gets lifted again and again until the process is OOM-killed. This patch skips the requeue when the target is the trace being lifted to address the issue.

Fixes #782.

### DIFF
--- a/lib/BC/TraceLifter.cpp
+++ b/lib/BC/TraceLifter.cpp
@@ -485,7 +485,9 @@ bool TraceLifter::Impl::Lift(
         direct_func_call:
           try_add_delay_slot(true, block);
           if (inst.branch_not_taken_pc != inst.branch_taken_pc) {
-            trace_work_list.insert(inst.branch_taken_pc);
+            if (inst.branch_taken_pc != trace_addr) {
+              trace_work_list.insert(inst.branch_taken_pc);
+            }
             auto target_trace = get_trace_decl(inst.branch_taken_pc);
             AddCall(block, target_trace, *intrinsics);
           }
@@ -524,7 +526,9 @@ bool TraceLifter::Impl::Lift(
           llvm::BranchInst::Create(taken_block, not_taken_block,
                                    LoadBranchTaken(block), block);
 
-          trace_work_list.insert(inst.branch_taken_pc);
+          if (inst.branch_taken_pc != trace_addr) {
+            trace_work_list.insert(inst.branch_taken_pc);
+          }
           auto target_trace = get_trace_decl(inst.branch_taken_pc);
 
           AddCall(taken_block, intrinsics->function_call, *intrinsics);


pgoodman:
Practically speaking, the trace lifter should really be overhauled and whatever is in Anvill's implementation should ideally be brought over for completeness, etc.

t0rr3sp3dr0:
Well, that requires a bit more effort than a simple one-line bug fix 🙃

pgoodman:
Yeah, just suggesting future directions ;-)

kyle-elliott-tob (approved):

kyle-elliott-tob:
Don't worry about the docker failures, CI needs to be re-mangled anyways. ty

